PR #10, opened 2022-07-27
Add Speedometer Governance Policy

rniwa:


addyosmani (approved):
LGTM as discussed.

bgrins:
Text looks good to me. Could we put this in a separate file (something like policy.md or governance.md)? That way we could keep the readme more focused on an introduction (what the project is, goals, etc) and link to this doc from there.

rniwa:
> Text looks good to me. Could we put this in a separate file (something like policy.md or governance.md)? That way we could keep the readme more focused on an introduction (what the project is, goals, etc) and link to this doc from there.

Since the first sentence reads "[i]n order to achieve the goals above," it seems that we want the goal & governance in th same document?

bgrins (approved):
Thanks!

rniwa:
@addyosmani : does this still look good to you after moving it to governance.md?

addyosmani:
> @addyosmani : does this still look good to you after moving it to governance.md?

It does indeed. LGTM.

PR #11, opened 2022-08-08
Add react-stockcharts benchmarks

squarewave:
This is a work-in-progress patch that adds two react-stockcharts benchmark suites.

Something we may want to consider adding to the runner is the ability to wait for certain async conditions within the body of the test. For example, these tests pan and zoom the stockchart view repeatedly, but we do not wait for animation frames or anything inside the benchmark function body. Accordingly, there are some pieces of the browser's execution of these tasks in the real world that goes unmeasured. However, if we were to wait for animation frames, it would set a floor on the execution time for the test dependent on the refresh rate. It seems like a general policy should be established here regarding what kinds of things we want to measure and what compromises we're willing to make.


bgrins:
> However, if we were to wait for animation frames, it would set a floor on the execution time for the test dependent on the refresh rate.

@rniwa do you think this would be a problem in practice? Do we expect any tests would execute end to end faster than the refresh rate. Because being able to await inside of a test and track that time does seem like it'd allow us to capture more diverse types of input and scenarios

rniwa:
> @rniwa do you think this would be a problem in practice? Do we expect any tests would execute end to end faster than the refresh rate. Because being able to await inside of a test and track that time does seem like it'd allow us to capture more diverse types of input and scenarios

Yes, many sub tests finish much quicker than 16ms.

bgrins:
> the ability to wait for certain async conditions within the body of the test

Even if we don't wait for animation frames, it does seem like having the ability to await for some page-specific event is going to be necessary as we start to pull in more variety in user journeys, libraries, etc which may signal completion via events/promises/etc. Is that something already available in the runner, and if not what do you think about adding it?

resources/tests.js
```diff
@@ -578,3 +578,55 @@ Suites.push({
         }),
     ]
 });
+
+var stockchartsTests = [
```
rniwa, line 582: This needs to be duplicated for each suite, or else the test runner is gonna get confused.

rniwa:
Hm... I'm observing that sometimes async time for "PanTheChart" step disappears / not getting measured properly. It's usually ~50ms on my machine but sometimes shows up as ~0.5ms. I suspect React is waiting for rAF and depending on the order by which rAF and our 0s timer fires, we're sometimes failing to account for the rendering time by React.

squarewave:
> Hm... I'm observing that sometimes async time for "PanTheChart" step disappears / not getting measured properly. It's usually ~50ms on my machine but sometimes shows up as ~0.5ms. I suspect React is waiting for rAF and depending on the order by which rAF and our 0s timer fires, we're sometimes failing to account for the rendering time by React.

I tested locally and it looks like the proposed system of initiating exectuion of the benchmark steps from within an rAF should resolve this issue.

More generally, this should be more or less ready for review. The original code in the GrandPrix version looks to have been taken from an example for the react-stockcharts library, which was MIT licensed. So I've now just pulled the example from the same place. However given that it was just an example intended to document the library's APIs, I cut out a bunch of code to eliminate distracting things and I made all the graphs fit in the iframe.

Also I realized while poking at this that the interactive runner is broken, so I adjusted it to use modules.

squarewave:
All right - this is ready to be looked at, at least. I added it under the `tentative` directory to indicate that we're still not sure about it. It's fairly unrealistic that we're sending 100 pans and 100 zooms without any paints in between. I don't think any browsers are currently set up to game this and avoid updating the canvas, as that would likely have little benefit to end users, but that is entirely possible. Admittedly this is a similarly unrealistic scenario as our benchmarks of TodoMVC, but it still makes me a little nervous. I think we should talk about this test a bit more in our next meeting.

housseindjirdeh:
New `tentative` directory LGTM 👍

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+        }),
+        new BenchmarkTestStep('PanTheChart', page => {
+            const cursor = page.querySelector('.react-stockcharts-crosshair-cursor');
+            let x = 150;
+            let y = 200;
+            const coords = (i) => ({ clientX: x + i * 10, clientY: y + i * 2, bubbles: true, cancelable: true });
+            for (let i = 0; i < 100; ) {
+                cursor.dispatchEvent('mousedown', coords(i), MouseEvent);
+                for (let j = 10; j--; ) {
+                    cursor.dispatchEvent('mousemove', coords(++i), MouseEvent);
```
rniwa, line 489: Nit: No curly braces around a single line statement.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+                }
+                cursor.dispatchEvent('mouseup', coords(i), MouseEvent);
+            }
+        }),
+        new BenchmarkTestStep('ZoomTheChart', page => {
+            const cursor = page.querySelector('.react-stockcharts-crosshair-cursor');
+            let ev = { clientX: 200, clientY: 200, deltaMode: 0, delta: -10, deltaY: -10, bubbles: true, cancelable: true  };
+            for (let delta = 0; delta < 30; delta++) {
+                cursor.dispatchEvent('wheel', ev, WheelEvent);
```
rniwa, line 498: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+            }
+            ev = { clientX: 650, clientY: 200, deltaMode: 0, delta: 10, deltaY: 10, bubbles: true, cancelable: true  };
+            for (let delta = 0; delta < 50; delta += 5) {
+                cursor.dispatchEvent('wheel', ev, WheelEvent);
```
rniwa, line 502: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+            }
+            ev = { clientX: 200, clientY: 200, deltaMode: 0, delta: -10, deltaY: -10, bubbles: true, cancelable: true  };
+            for (let delta = 0; delta < 20; delta += 2) {
+                cursor.dispatchEvent('wheel', ev, WheelEvent);
```
rniwa, line 506: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+        }),
+        new BenchmarkTestStep('PanTheChart', page => {
+            const cursor = page.querySelector('.react-stockcharts-crosshair-cursor');
+            let x = 150;
+            let y = 200;
+            const coords = (i) => ({ clientX: x + i * 10, clientY: y + i * 2, bubbles: true, cancelable: true });
+            for (let i = 0; i < 100; ) {
+                cursor.dispatchEvent('mousedown', coords(i), MouseEvent);
+                for (let j = 10; j--; ) {
+                    cursor.dispatchEvent('mousemove', coords(++i), MouseEvent);
```
rniwa, line 530: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+                }
+                cursor.dispatchEvent('mouseup', coords(i), MouseEvent);
+            }
+        }),
+        new BenchmarkTestStep('ZoomTheChart', page => {
+            const cursor = page.querySelector('.react-stockcharts-crosshair-cursor');
+            let ev = { clientX: 200, clientY: 200, deltaMode: 0, delta: -10, deltaY: -10, bubbles: true, cancelable: true  };
+            for (let delta = 0; delta < 30; delta++) {
+                cursor.dispatchEvent('wheel', ev, WheelEvent);
```
rniwa, line 539: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+            }
+            ev = { clientX: 650, clientY: 200, deltaMode: 0, delta: 10, deltaY: 10, bubbles: true, cancelable: true  };
+            for (let delta = 0; delta < 50; delta += 5) {
+                cursor.dispatchEvent('wheel', ev, WheelEvent);
```
rniwa, line 543: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+            }
+            ev = { clientX: 200, clientY: 200, deltaMode: 0, delta: -10, deltaY: -10, bubbles: true, cancelable: true  };
+            for (let delta = 0; delta < 20; delta += 2) {
+                cursor.dispatchEvent('wheel', ev, WheelEvent);
```
rniwa, line 547: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,85 @@ Suites.push({
...
+                cursor.dispatchEvent('mousedown', coords(i), MouseEvent);
+                for (let j = 10; j--; ) {
+                    cursor.dispatchEvent('mousemove', coords(++i), MouseEvent);
+                }
+                cursor.dispatchEvent('mouseup', coords(i), MouseEvent);
+            }
+        }),
+        new BenchmarkTestStep('ZoomTheChart', page => {
+            const cursor = page.querySelector('.react-stockcharts-crosshair-cursor');
+            let ev = { clientX: 200, clientY: 200, deltaMode: 0, delta: -10, deltaY: -10, bubbles: true, cancelable: true  };
```
rniwa, line 537: Please spell out event.

squarewave:
@rniwa your comments should be addressed if you can take a look again. Thanks!

resources/tests.mjs
```diff
@@ -467,3 +467,127 @@ Suites.push({
...
+            event = {
+                clientX: 650,
+                clientY: 200,
+                deltaMode: 0,
+                delta: 10,
+                deltaY: 10,
+                bubbles: true,
+                cancelable: true
+            };
+            for (let delta = 0; delta < 50; delta += 5)
```
rniwa, line 515: I'm confused. Were this delta being used anywhere? Not not just delta++ and delta < 10?

resources/tests.mjs
```diff
@@ -467,3 +467,127 @@ Suites.push({
...
+            event = {
+                clientX: 200,
+                clientY: 200,
+                deltaMode: 0,
+                delta: -10,
+                deltaY: -10,
+                bubbles: true,
+                cancelable: true
+            };
+            for (let delta = 0; delta < 20; delta += 2)
```
rniwa, line 527: Ditto.

resources/tests.mjs
```diff
@@ -467,3 +467,127 @@ Suites.push({
...
+                cancelable: true
+            };
+            for (let delta = 0; delta < 30; delta++)
+                cursor.dispatchEvent('wheel', event, WheelEvent);
+
+            event = {
+                clientX: 650,
+                clientY: 200,
+                deltaMode: 0,
+                delta: 10,
```
rniwa, line 510: delta here is 10. Perhaps you meant to increment by 10 below as well?

resources/tests.mjs
squarewave: Sorry! Yeah those are just leftovers.

rniwa (approved):

PR #14, opened 2022-08-25
Delete SimplePromise and use regular native Promise.

rniwa:
Also deleted the unused _waitAndWarmUp.

### DIFF
--- a/resources/benchmark-runner.js
+++ b/resources/benchmark-runner.js
@@ -1,37 +1,3 @@
-// FIXME: Use the real promise if available.
-// FIXME: Make sure this interface is compatible with the real Promise.
-function SimplePromise() {
-    this._chainedPromise = null;
-    this._callback = null;
-}
-
-SimplePromise.prototype.then = function (callback) {
-    if (this._callback)
-        throw "SimplePromise doesn't support multiple calls to then";
-    this._callback = callback;
-    this._chainedPromise = new SimplePromise;
-    
-    if (this._resolved)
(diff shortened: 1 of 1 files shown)


shvaikalesh (approved):

resources/benchmark-runner.js
```diff
@@ -95,33 +61,12 @@ BenchmarkRunner.prototype._appendFrame = function (src) {
...
-    return promise;
-}
-
 BenchmarkRunner.prototype._writeMark = function(name) {
     if (window.performance && window.performance.mark)
         window.performance.mark(name);
 }
 
-// This function ought be as simple as possible. Don't even use SimplePromise.
+// This function ought be as simple as possible. Don't even use Promise.
```
smaug----, line 69: (random comment, the setTimeout call there isn't "simple". An implementation may or may not call rAF before the timeout runs. And rAF can be rather heavy operation. So the results can be a bit surprising.  But not about this patch :) )

rniwa: Yeah, I think we want to refine this stuff in v3. I'm observing odd behavior for https://github.com/WebKit/Speedometer/pull/11 where async time sometimes disappears.

smaug---- (approved):
This of course does change a bit how tests are run, since real promises use microtask scheduling, but anyhow, we should do this.

kyliau (approved):

rniwa:
Thanks for the review, everyone.

PR #17, opened 2022-09-16
Use ES6 Classes

camillobruni:


camillobruni:
Given that this might be hard to diff, I tried only restructure code at the function-level and leave the rest untouched.

camillobruni:
Note 2: I will format the code in follow-up CL to make this review easier.

resources/benchmark-runner.js
```diff
@@ -1,15 +1,65 @@
-function BenchmarkTestStep(testName, testFunction) {
+class BenchmarkTestStep {
+constructor(testName, testFunction) {
     this.name = testName;
     this.run = testFunction;
 }
+}
+
+class BenchmarkState {
```
rniwa, line 8: Can we not change the order in which BenchmarkRunner and BenchmarkState are declared / defined?

resources/benchmark-runner.js
```diff
@@ -1,15 +1,65 @@
...
+    this._testIndex = 0;
+    this.next();
+}
+
+currentSuite() {
+    return this._suites[this._suiteIndex];
+}
+
+currentTest() {
+    let suite = this.currentSuite();
```
rniwa, line 21: We can use const here.

resources/main.js
```diff
@@ -1,33 +1,41 @@
-window.benchmarkClient = {
-    displayUnit: 'runs/min',
-    iterationCount: 10,
-    stepCount: null,
-    suitesCount: null,
-    _measuredValuesList: [],
-    _finishedTestCount: 0,
-    _progressCompleted: null,
+// TODO(camillobruni): Add base class
```
rniwa, line 1: Please use FIXMEs instead of TODOs.

rniwa (approved):

bgrins (approved):

PR #19, opened 2022-09-19
Convert to ES6 Modules

camillobruni:
- Rename files from .js to .mjs (not strictly necessary but it makes clear that we use modules)
- Pull in missing statistics.js file from https://github.com/WebKit/WebKit/blob/main/PerformanceTests/resources/statistics.js
- Restructure UI init code to be all in main.mjs
- Moved free-floating UI code to MainBenchmarkClient
- Added `_isRunning` instance variable to MainBenchmarkClient
- Fix CSS to address removed links
- Use `<button>` instead of links for better keyboard navigation


camillobruni:
@rniwa I missed your formatting CL. I'll wait for that one to land first.

bgrins:
> Pull in missing statistics.js file

Just for reference, can you link to the source you pulled in for this file?

resources/main.mjs
```diff
@@ -234,28 +238,36 @@ function showSection(sectionIdentifier, pushState) {
...
-}
+function prepareUI() {
+    if (benchmarkClient.prepareUI)
+        benchmarkClient.prepareUI();
 
-function showAbout() {
-    showSection('about', true);
+    document.getElementById("logo").onclick = () => {
+        showSection('home', true);
+    };
```
juliandescottes, line 255: I can't comment on the proper lines, but see L42-46 of this file, there is still logic bound to the "logo-link" id, which was changed to "logo" here. If I understand correctly, this was preventing clicks on the logo while the tests are running.

We need to update those references, and maybe preserve the behavior with the new "logo" element.

juliandescottes (changes requested):
Thanks, overall this looks good to me. There are remaining references to the "logo-link" id to cleanup (I assume this breaks displaying the results?)

bgrins:
I trust Julian's reviews, so removing myself

resources/main.mjs
camillobruni: You're right.
I've now moved more code to address this.

index.html
```diff
@@ -4,15 +4,11 @@
...
-    <script src="resources/benchmark-runner.js" defer></script>
-    <script src="resources/benchmark-report.js" defer></script>
-    <script src="../resources/statistics.js" defer></script>
-    <script src="resources/tests.js" defer></script>
+    <script src="resources/main.mjs" type="module"></script>
 </head>
 <body>
 <main>
-    <a id="logo-link" href="javascript:showHome()"><img id="logo" src="resources/logo.png"></a>
+    <img id="logo" src="resources/logo.png">
```
rniwa, line 11: This will make it less keyboard accessible. We need to either make the logo itself focusable or add back anchor element.

index.html
```diff
@@ -24,9 +20,9 @@
             It's currently <span id="screen-size"></span>.
         </strong></p>
         <div class="buttons">
-            <button onclick="startTest()">Start Test</button>
+            <button class="start-tests-button">Start Test</button>
         </div>
-        <p class="show-about"><a href="javascript:showAbout()">About Speedometer</a></p>
+        <p class="show-about">About Speedometer</p>
```
rniwa, line 25: Ditto.

index.html
```diff
@@ -53,10 +49,10 @@ <h1>Detailed Results</h1>
         <table class="results-table"></table>
         <div class="arithmetic-mean"><label>Arithmetic Mean:</label><span id="results-with-statistics"></span></div>
         <div class="buttons">
-            <button onclick="startTest()">Test Again</button>
-            <button id="show-summary" onclick="showResultsSummary()">Summary</button>
+            <button class="start-tests-button">Test Again</button>
+            <button id="show-summary"> Summary</button>
         </div>
-        <p class="show-about"><a href="javascript:showAbout()">About Speedometer</a></p>
+        <p class="show-about">About Speedometer</p>
```
rniwa, line 55: Ditto.

resources/tests.mjs
```diff
@@ -1,5 +1,7 @@
+import{ BenchmarkTestStep} from "./benchmark-runner.mjs";
```
rniwa, line 1: Nit: There should be a space between `BenchmarkTestStep` and `}`.

index.html
camillobruni: changed to <button>.

resources/tests.mjs
camillobruni: DONE.

index.html
```diff
@@ -4,15 +4,11 @@
...
-    <script src="resources/benchmark-runner.js" defer></script>
-    <script src="resources/benchmark-report.js" defer></script>
-    <script src="../resources/statistics.js" defer></script>
-    <script src="resources/tests.js" defer></script>
+    <script src="resources/main.mjs" type="module"></script>
 </head>
 <body>
 <main>
-    <a id="logo-link" href="javascript:showHome()"><img id="logo" src="resources/logo.png"></a>
+    <a id="logo"><img src="resources/logo.png" alt="Speedometer"></a>
```
rniwa, line 11: a without href is still not focusable.

resources/main.mjs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    _startBenchmarkHandler() {
+        if (this.startBenchmark())
+            this._showSection('running');
+    }
+
+    _logoClickHandler(event) {
+        // Prevent any accidental UI changes during benchmark runs.
+        if (!this._isRunning) this._showSection('home', true);
```
rniwa, line 245: Please insert a line break between `)` and `this`.

index.html
rniwa, line 11: a without href is still not focusable.

resources/main.mjs
camillobruni: DONE.

index.html
camillobruni: A thanks for the hint. Seems like I didn't check in Safari (Chrome just accepts all links).
And I realized that this didn't work well in FireFox in the old version.
- changed this to `<button>` as well and cleaned up the CSS a bit to style the button better

index.html
```diff
@@ -24,9 +20,10 @@
             It's currently <span id="screen-size"></span>.
         </strong></p>
         <div class="buttons">
-            <button onclick="startTest()">Start Test</button>
+            <button class="start-tests-button">Start Test</button>
+            <br/>
```
rniwa, line 24: I don't think we want to have / in <br> since we're using HTML parser, not XML parser.

index.html
```diff
@@ -53,10 +50,11 @@ <h1>Detailed Results</h1>
         <table class="results-table"></table>
         <div class="arithmetic-mean"><label>Arithmetic Mean:</label><span id="results-with-statistics"></span></div>
         <div class="buttons">
-            <button onclick="startTest()">Test Again</button>
-            <button id="show-summary" onclick="showResultsSummary()">Summary</button>
+            <button class="start-tests-button">Test Again</button>
+            <button id="show-summary"> Summary</button>
+            <br/>
```
rniwa, line 55: Ditto.

resources/main.css
```diff
@@ -79,6 +77,15 @@ a {
 }
 
 button {
+    -webkit-appearance: none;
```
rniwa, line 80: Why -webkit- prefix?

index.html
camillobruni: remove.

resources/main.css
camillobruni: This was just copied over from a few lines above.
- Updated all outdate `-webkit-` prefixed CSS properties for consistency

index.html
camillobruni: Switched to `<button>`.

resources/main.css
```diff
@@ -19,17 +19,14 @@ code {
...
-
 hr {
     border: 1px solid rgb(235, 235, 235);
     width: 50%;
     margin: 40px auto;
 }
 
 img {
+    user-select: none;
     -webkit-user-select: none;
```
camillobruni, line 30: `user-select` seems to not work as expected in Safari 16.0, thus I left this in.

camillobruni:
PTAL again

resources/main.mjs
```diff
@@ -0,0 +1,274 @@
...
+            this._addDetailedResultsRow(resultsTables[1], i, formattedValues[i]);
+    }
+
+    prepareUI() {
+        window.addEventListener('popstate', this._popStateHandler.bind(this), false);
+        window.addEventListener('resize', this._resizeScreeHandler.bind(this));
+        this._resizeScreeHandler();
+
+        document.getElementById("logo").onclick = this._logoClickHandler.bind(this);
+        document.getElementById("show-summary").onclick = (e) => this.showResultsSummary();
```
rniwa, line 207: Please don't use one letter variable name like "e". Please spell out event.

resources/main.mjs
```diff
@@ -0,0 +1,274 @@
...
+        document.getElementById("show-details").onclick = (e) => this.showResultsDetails();
```
rniwa, line 208: Ditto

resources/main.mjs
```diff
@@ -0,0 +1,274 @@
...
+        document.querySelectorAll(".show-about").forEach(
+            each => { each.onclick = () => this._showSection('about', true); }
```
rniwa, line 210: Can we add this as a method like showResultsSummary, etc...

resources/main.mjs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    _resizeScreeHandler() {
+        // FIXME: Detect when the window size changes during the test.
+        const screenIsTooSmall = window.innerWidth < 850 || window.innerHeight < 650;
+        document.getElementById('screen-size').textContent = window.innerWidth + 'px by ' + window.innerHeight + 'px';
+        document.getElementById('screen-size-warning').style.display = screenIsTooSmall ? null : 'none';
+    }
+
+    _startBenchmarkHandler() {
```
rniwa, line 238: I think we should just call this startBenchmarkHandler. _ prefix doesn't really buy us anything.

camillobruni: Agreed there is no real "protected". Since we do have other `_`-prefixed methods in the `MainBenchmarkClient`, how about following one of the following strategies for more consistency:
a) Consistently use `_`-prefix for "protected" methods that are not referenced from outside the class
b) Drop `_` prefix allt ogether

resources/main.mjs
camillobruni: Used a "cleaner" `.bind(this)` approach.

rniwa (approved):
r=me

resources/main.css
```diff
@@ -54,6 +51,8 @@ main {
     top: 115px;
     width: 75px;
     height: 406px;
+    cursor: pointer;
+
```
juliandescottes, line 55: nit: remove extra line

resources/main.css
```diff
@@ -123,7 +131,7 @@ section.selected {
     height: 600px;
 }
 
-section#home > p {
+section#home > p{
```
juliandescottes, line 134:
```suggestion
section#home > p {
```

juliandescottes (approved):
LGTM, sorry about the delay

PR #22, opened 2022-10-13
Remove flight-js mailclient

camillobruni:
Remove the flightjs-mailclient app that was not included in the main Suites list.

PR #23, opened 2022-10-13
Add Page object

camillobruni:
Goal: Add a fixed interface to interact with test frames and their DOM nodes.
This will allow us to run more complex scenarios in the future without sacrificing flexibility or performance for the existing tests.

- Add `Page` object
- Refactor tests to interact with the frame only through the Page object


camillobruni:
Depending on the resolution of PR #22 I might have to fix the mail client implementation to work with the `Page` API.

resources/benchmark-runner.mjs
```diff
@@ -41,43 +41,59 @@ class BenchmarkState {
         return !this._testIndex;
     }
 
-    prepareCurrentSuite(runner, frame) {
+    async prepareCurrentSuite(frame, page) {
```
camillobruni, line 44: The diff is still a bit messy due to depending on PR #19. 
For easier reviewing, I'll leave this PR as WIP until we land #19.

resources/benchmark-runner.mjs
```diff
@@ -41,43 +41,71 @@ class BenchmarkState {
...
-            function resolveIfReady() {
-                const element = contentDocument.querySelector(selector);
-                if (element) {
-                    window.requestAnimationFrame(function () {
-                        return resolve(element);
-                    });
-                    return;
-                }
+          const resolveIfReady = () => {
+            const element = this.querySelector(selector);
```
rniwa, line 66: Nit: What is up with the indentation here? Seems to be 2 space instead of 4?

resources/benchmark-runner.mjs
```diff
@@ -41,43 +41,71 @@ class BenchmarkState {
...
                 setTimeout(resolveIfReady, 50);
             }
-
-            resolveIfReady();
+          };
+          resolveIfReady();
         });
     }
 
+    querySelector(selector) {
```
rniwa, line 79: WebKit style is to put { in a new line.
I know we haven't been consistent but we might as well as in new code.

resources/benchmark-runner.mjs
```diff
@@ -41,43 +41,71 @@ class BenchmarkState {
...
         });
     }
 }
 
-export class BenchmarkRunner {
-    constructor(suites, client) {
-        this._suites = suites;
-        this._prepareReturnValue = null;
-        this._client = client;
+const ENTER_KEY_CODE = 13;
```
rniwa, line 56: It doesn't seem like this needs to be global?
We can just declare it within `triggerEnter` instead.

resources/benchmark-runner.mjs
```diff
@@ -147,57 +172,54 @@ export class BenchmarkRunner {
         }, 0);
     }
 
-    step(state) {
+    async step(state) {
```
rniwa, line 175: Can we do this `async` syntax adoption in a separate patch?
It's making the code review harder.

resources/benchmark-runner.mjs
```diff
@@ -147,57 +172,54 @@ export class BenchmarkRunner {
...
         if (state.isFirstTest()) {
             this._removeFrame();
-            let self = this;
-            return state.prepareCurrentSuite(this, this._appendFrame()).then(prepareReturnValue => {
-                self._prepareReturnValue = prepareReturnValue;
-                return self._runTestAndRecordResults(state);
-            });
+            this._appendFrame();
+            this._page = new Page(this._frame);
+            await state.prepareCurrentSuite(this._frame, this._page);
```
rniwa, line 191: Why are we passing both this._frame and this._page?
Can't we just use pass page here?

resources/benchmark-runner.mjs
```diff
@@ -147,57 +172,54 @@ export class BenchmarkRunner {
...
         }
 
-        return this._runTestAndRecordResults(state);
+        return await this._runTestAndRecordResults(state);
     }
 
-    runAllSteps(startingState) {
+    async runAllSteps(startingState) {
```
rniwa, line 197: Same comment about doing this `async` syntax adoption in a separate PR.

resources/tests.mjs
```diff
@@ -4,39 +4,29 @@ const numberOfItemsToAdd = 100;
...
 Suites.push({
     name: 'VanillaJS-TodoMVC',
     url: 'todomvc/vanilla-examples/vanillajs/index.html',
-    prepare(runner, contentWindow, contentDocument) {
-        return runner.waitForElement('.new-todo').then(element => {
-            element.focus();
-            return element;
-        });
+    async prepare(page) {
+        const element = await page.waitForElement('.new-todo');
```
rniwa, line 11: This local variable seems unnecessary.
We can just do: `(await ~).focus();`

resources/tests.mjs
```diff
@@ -106,31 +94,31 @@ Suites.push({
...
     name: 'React-TodoMVC',
     url: 'todomvc/architecture-examples/react/index.html',
-    prepare(runner, contentWindow, contentDocument) {
-        contentWindow.app.Utils.store = () => {};
-        return runner.waitForElement('.new-todo').then(element => {
-            element.focus();
-            return element;
-        });
+    async prepare(page) {
+        // FIXME: use app code
```
rniwa, line 98: What does this FIXME mean?

resources/tests.mjs
```diff
@@ -227,28 +212,28 @@ Suites.push({
...
     url: 'todomvc/architecture-examples/backbone/index.html',
-    prepare(runner, contentWindow, contentDocument) {
-        contentWindow.Backbone.sync = () => {};
-        return runner.waitForElement('#appIsReady').then(element => {
-            const newTodo = contentDocument.querySelector('.new-todo');
-            newTodo.focus();
-            return newTodo;
-        });
+    async prepare(page) {
+        // FIXME: use app code
```
rniwa, line 216: Ditto.

resources/tests.mjs
```diff
@@ -357,55 +339,53 @@ Suites.push({
...
+        new BenchmarkTestStep('CompletingAllItems', page => {
+            const checkboxes = page.querySelectorAll('.toggle');
             for (let i = 0; i < checkboxes.length; i++)
                 checkboxes[i].click();
         }),
-        new BenchmarkTestStep('DeletingAllItems', (newTodo, contentWindow, contentDocument) => {
+        new BenchmarkTestStep('DeletingAllItems', page => {
             for (let i = 0; i < numberOfItemsToAdd; i++)
-                contentDocument.querySelector('.destroy').click();
+            page.querySelector('.destroy').click();
```
rniwa, line 362: This seems like wrong indentation.

resources/tests.mjs
```diff
@@ -415,37 +395,38 @@ Suites.push({
...
             for (let i = 0; i < numberOfItemsToAdd; i++)
-                contentDocument.querySelector('.destroy').click();
+                page.querySelector('.destroy').click();
         }),
     ]
 });
 
-function processElmWorkQueue(contentWindow) {
+function processElmWorkQueue(page) {
+    // TODO move workload to test
```
rniwa, line 428: What is this TODO about? Also, please use FIXME instead of TODO.

squarewave (commented):
This all looks good to me, as far as the substance of the change goes. Thanks for taking this on.

resources/tests.mjs
camillobruni: DONE.

camillobruni:
I've addressed your comments.
Will upload the async-conversion PR soon.

resources/benchmark-runner.mjs
```diff
@@ -5,24 +5,29 @@ export class BenchmarkTestStep {
     }
 }
 
-class BenchmarkState {
-    constructor(suites) {
+class BenchmarkState
```
rniwa, line 8: Oh, class declaration puts { on the same line. Just function declaration that puts { on a new line.
Sorry, should have been more clear.

resources/tests.mjs
```diff
@@ -415,37 +394,38 @@ Suites.push({
...
             for (let i = 0; i < checkboxes.length; i++)
                 checkboxes[i].click();
         }),
-        new BenchmarkTestStep('DeletingItems', (params, contentWindow, contentDocument) => {
-            const deleteButtons = contentDocument.querySelectorAll('.destroy');
+        new BenchmarkTestStep('DeletingItems', page => {
+            const deleteButtons = page.querySelectorAll('.destroy');
             for (let i = 0; i < numberOfItemsToAdd; i++)
-                contentDocument.querySelector('.destroy').click();
+                page.querySelector('.destroy').click();
```
rniwa, line 421: Huh, it seems unnecessary to keep calling querySelector if we to all the buttons to delete in the first place.
This is unrelated to your change so we can fix it up in a separate PR.

resources/tests.mjs
```diff
@@ -415,37 +394,38 @@ Suites.push({
...
         }),
     ]
 });
 
-function processElmWorkQueue(contentWindow) {
+function processElmWorkQueue(page) {
```
rniwa, line 426: Nit: { should be on new line.

rniwa (approved):
r=me

bgrins (approved):
Approving as per Doug's comment

bgrins:
@camillobruni is this good to land?

camillobruni:
Sorry, long weekend. double checking now.

PR #26, opened 2022-11-08
Add more Page methods and PageElement

camillobruni:
Migrate all test code to go through Page or PageElement.

- Page defines a fixed interface to interact with a test frame
- PageElement defines a fixed interface to interact with Elements from an iframe

Cleanups:
- Move app initialization and custom code to the specific index.html files
- Update `even.initEvent(...)` code to use the recommended `Event` constructor

Depending on #29

PR #28, opened 2022-11-22
BenchmarkRunner test measurement refactoring

camillobruni:
- Added separate MeasureTask for measuring 
- Added params.asyncMetric to experiment with the existing "timeout"-based and new "raf"-based approach


camillobruni:
Depending on https://github.com/WebKit/Speedometer/pull/26

resources/benchmark-runner.mjs
```diff
@@ -160,112 +176,107 @@ export class BenchmarkRunner {
...
-        this.step(startingState).then(nextState => {
-            if (nextState)
-                nextCallee(nextState);
+    async _runTestAndRecordResults(suite, test)
+    {
+        return new Promise((resolve) => {
+            if (this._client && this._client.willRunTest)
+                this._client.willRunTest(suite, test);
+
+             window.requestAnimationFrame(
```
camillobruni, line 233: Change of this PR

resources/benchmark-runner.mjs
```diff
@@ -160,112 +176,107 @@ export class BenchmarkRunner {
...
+            window.requestAnimationFrame(
+                this._recordTestResults.bind(this, suite, test, syncTime, asyncTime, test_done_callback));
+        }, 0);
     }
 
-    _runTestAndRecordResults(state) {
-        return new Promise((resolve) => {
-            const suite = state.currentSuite();
-            const test = state.currentTest();
+    _recordTestResults(suite, test, syncTime, asyncTime, test_done_callback)
```
camillobruni, line 264: moved the callback code to here.

resources/benchmark-runner.mjs
```diff
@@ -160,112 +176,107 @@ export class BenchmarkRunner {
...
+        this._writeMark(suite.name + '.' + test.name + '-sync-end');
 
-        if (this._client && this._client.willStartFirstIteration)
-            this._client.willStartFirstIteration(iterationCount);
+        const syncTime = endTime - startTime;
+        this._measureTestAsyncTime(suite, test, syncTime, test_done_callback)
+    }
 
-        self.runAllSteps();
+    _measureTestAsyncTime(suite, test, syncTime, test_done_callback)
```
camillobruni, line 251: New separate function (probably not very critical, but could be inlined again)

resources/benchmark-runner.mjs
```diff
@@ -160,112 +176,107 @@ export class BenchmarkRunner {
...
+    {
+        const startTime = window.performance.now();
+        setTimeout(() => {
+            const endTime = window.performance.now();
+            this._writeMark(suite.name + '.' + test.name + '-async-end');
+
+            const asyncTime = endTime - startTime;
+            window.requestAnimationFrame(
+                this._recordTestResults.bind(this, suite, test, syncTime, asyncTime, test_done_callback));
```
camillobruni, line 260: I got into the habit of using bind instead of arrow-function to prevent potential context-variable leaks. (can be changed back of course)

resources/tests.mjs
```diff
@@ -13,8 +13,8 @@ Suites.push({
         new BenchmarkTestStep('Adding' + numberOfItemsToAdd + 'Items', page => {
             const newTodo = page.querySelector(".new-todo");
             for (let i = 0; i < numberOfItemsToAdd; i++) {
-                newTodo.value = 'Something to do ' + i;
-                newTodo.change();
+                newTodo.setValue('Something to do ' + i);
+                newTodo.dispatchEvent("change");
```
rniwa, line 17: Hm... maybe we should stick with ' instead of " for self consistency.

resources/benchmark-runner.mjs
```diff
@@ -90,19 +90,18 @@ class PageElement {
...
+        if (eventName == 'submit') {
+            // FIXME FireFox doesn't like `new Event('submit')
+            this._dispatchSubmitEvent();
+        }  else {
+            this._node.dispatchEvent(new Event(eventName, options))
+        }
     }
 
-    submit(options = NATIVE_OPTIONS)
+    _dispatchSubmitEvent()
```
rniwa, line 103: L

resources/benchmark-runner.mjs
```diff
@@ -90,19 +90,18 @@ class PageElement {
...
     {
-        // FIXME FireFox doesn't like `new Event('submit')
```
rniwa: Looks like this comment about Firefox is no longer applicable.
We should probably just use dispatchEvent instead.

resources/benchmark-runner.mjs
```diff
@@ -90,19 +90,18 @@ class PageElement {
...
-
-    input(options = NATIVE_OPTIONS)
-    {
-        this._node.dispatchEvent(new Event('input', options))
+        if (eventName == 'submit') {
+            // FIXME FireFox doesn't like `new Event('submit')
+            this._dispatchSubmitEvent();
+        }  else {
+            this._node.dispatchEvent(new Event(eventName, options))
+        }
```
rniwa, line 100: No curly braces around a single line statement.

resources/benchmark-runner.mjs
```diff
@@ -160,112 +175,107 @@ export class BenchmarkRunner {
...
-        self.runAllSteps();
+    _measureTestAsyncTime(suite, test, syncTime, test_done_callback)
+    {
+        const startTime = window.performance.now();
+        setTimeout(() => {
+            const endTime = window.performance.now();
+            this._writeMark(suite.name + '.' + test.name + '-async-end');
+
+            const asyncTime = endTime - startTime;
+            window.requestAnimationFrame(
```
rniwa, line 258: It seems this rAF is unnecessary now?

resources/benchmark-runner.mjs
```diff
@@ -5,108 +5,123 @@ export class BenchmarkTestStep {
...
-        element.dispatchEvent(event);
+        let eventOptions = {
+            bubbles: true, 
+            cancelable: true,
+            keyCode: ENTER_KEY_CODE,
+            which: ENTER_KEY_CODE,
+            key: 'ENTER'
+        };
+        if (options !== undefined) {
+            eventOptions = Object.assign(eventOptions, options)
```
rniwa, line 121: Nit: No curly braces around single line statements.

resources/benchmark-runner.mjs
```diff
@@ -160,112 +175,107 @@ export class BenchmarkRunner {
...
-
-        self._writeMark(suite.name + '.' + test.name + '-start');
-        let startTime = now();
-        test.run(page);
-        let endTime = now();
-        self._writeMark(suite.name + '.' + test.name + '-sync-end');
+        if (this._client && this._client.willStartFirstIteration)
+            this._client.willStartFirstIteration(iterationCount);
+        for (let i= 0; i< iterationCount; i++) {
+            await self._runAllSuites();
```
rniwa, line 184: Nit: No curly braces around single line statements.

resources/benchmark-runner.mjs
```diff
@@ -160,112 +175,107 @@ export class BenchmarkRunner {
...
-        const now = window.performance && window.performance.now ? () => window.performance.now() : Date.now;
-
-        self._writeMark(suite.name + '.' + test.name + '-start');
-        let startTime = now();
-        test.run(page);
-        let endTime = now();
-        self._writeMark(suite.name + '.' + test.name + '-sync-end');
+        if (this._client && this._client.willStartFirstIteration)
+            this._client.willStartFirstIteration(iterationCount);
+        for (let i= 0; i< iterationCount; i++) {
```
rniwa, line 183: Nit: Need a space before = and <.

resources/tests.mjs
camillobruni: Done in PR 26

resources/benchmark-runner.mjs
camillobruni: Which version are you using? FF 107.0 (64-bit) still get's stuck on AngularJS-TodoMVC if I don't use the custom event.

resources/benchmark-runner.mjs
camillobruni: Ah true, we have the one RAF when we start the test step.

resources/benchmark-runner.mjs
```diff
@@ -31,28 +31,94 @@ class Page {
 
     querySelector(selector)
     {
-        return this._frame.contentDocument.querySelector(selector);
+        const element = this._frame.contentDocument.querySelector(selector);
+        if (element === null) return null
```
rniwa, line 35: return needs to be on its own line.

resources/benchmark-runner.mjs
```diff
@@ -31,28 +31,94 @@ class Page {
...
+        for (let i = 0; i < elements.length; i++)
+            elements[i] = this._wrapElement(elements[i]);
+        return elements
     }
 
     getElementById(id)
     {
-        return this._frame.contentDocument.getElementById(id);
+        const element = this._frame.contentDocument.getElementById(id);
+        if (element === null) return null
```
rniwa, line 50: Ditto.

resources/benchmark-runner.mjs
smaug----: That is surprising when the c++ code for new Event literally calls initEvent
https://searchfox.org/mozilla-central/rev/670e2e0999f04dc7734c8c12b2c3d420a1e31f12/dom/events/Event.cpp#365-367

squarewave: > That is surprising when the c++ code for new Event literally calls initEvent https://searchfox.org/mozilla-central/rev/670e2e0999f04dc7734c8c12b2c3d420a1e31f12/dom/events/Event.cpp#365-367

I agree it's surprising, though I observe the same behavior on Nightly.

rniwa:
Could you rebase this against the current main?

camillobruni:
Ups, I thought I did that already yesterday.

resources/benchmark-runner.mjs
```diff
@@ -227,50 +227,50 @@ export class BenchmarkRunner {
...
 
         const syncTime = endTime - startTime;
+        this._measureTestAsyncTime(suite, test, syncTime, test_done_callback);
+    }
 
-        startTime = now();
+    _measureTestAsyncTime(suite, test, syncTime, test_done_callback)
+    {
+        const startTime = window.performance.now();
         setTimeout(() => {
```
smaug----, line 251: Isn't it possible that test.run() triggered some promises, which have their callbacks run at the microtask checkpoint, which happens _after_ this setTimeout has been called. And those callbacks might trigger some setTimeout(0) stuff. All that, maybe non-insignificant work, would happen after the timeout above runs.

We could have two rAF callbacks. First one runs sync tests, and since microtasks run right after it, all the promises are handled there too. Then the second rAF callback would be capture the end time of the sync testing and call setTimeout(, 0). That would push the setTimeout of the testing framework to happen after the ones triggered by run.test()

resources/benchmark-runner.mjs
```diff
@@ -227,50 +227,50 @@ export class BenchmarkRunner {
...
+            const endTime = window.performance.now();
             this._writeMark(suite.name + '.' + test.name + '-async-end');
-            window.requestAnimationFrame(() => {
-                callback(syncTime, endTime - startTime, height);
-            });
+
+            const asyncTime = endTime - startTime;
+            this._recordTestResults(suite, test, syncTime, asyncTime, test_done_callback);
         }, 0);
     }
```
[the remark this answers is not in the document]
rniwa: Have we gotten around to do any analysis on this?

camillobruni:
Created PR #116 which adds custom testInitiator modes.

resources/benchmark-runner.mjs
```diff
@@ -116,6 +116,106 @@ class PageElement {
...
+        this._measureForceLayout();
+        const endTime = performance.now();
+        this.asyncTime = endTime - this._asyncStartTime;
+        performance.mark(this._asyncEndLabel);
+        this._asyncDoneCallback();
+    }
+
+    _measureAsyncRaf() {
+        this._measureForceLayout();
+        const endTime = performance.now();
```
smaug----, line 203: But this called as the second rAF callback. This wouldn't capture setTimeout(, 0).
Don't we want to do start a setTimeout(function_which_captures_the_end_of_async ,0).
That way the setTimeout(,0)s triggered by the sync part would be captured as part of the async time.

resources/benchmark-runner.mjs
```diff
@@ -116,6 +116,106 @@ class PageElement {
...
+
+    async run() {
+        if (this._wasRun)
+            throw Error("MeasureTask can only run once.");
+        if (params.asyncMetric === "raf") {
+            requestAnimationFrame(this._measureSyncCallback);
+            requestAnimationFrame(this._measureAsyncRafCallback);
+        } else {
+            setTimeout(this._measureSyncCallback, 0);
+        }
```
rniwa, line 161: Nit: No curly braces around a single line statement.

resources/benchmark-runner.mjs
```diff
@@ -116,6 +116,106 @@ class PageElement {
...
+
+        this._unusedHeight = undefined;
+
+        this._wasRun = false;
+
+        this._asyncMeasurePromise = new Promise((resolve) => {
+            this._asyncDoneCallback = resolve;
+        });
+        this._measureSyncCallback = this._measureSync.bind(this);
+        this._measureAsyncTimeoutCallback = this._measureAsyncTimeout.bind(this);
```
rniwa, line 135: Since we're no longer using setTimeout, maybe we should call this._measureAsyncCallback?

resources/benchmark-runner.mjs
```diff
@@ -116,6 +116,106 @@ class PageElement {
...
+        const startTime = performance.now();
+        this._unusedHeight = this._frame.contentDocument.body.getBoundingClientRect().height;
+        const endTime = performance.now();
+        performance.mark(this._forcedLayoutEndLabel);
+        this.layoutTime = endTime - startTime;
+        // Prevent dead code elimination.
+        this._frame.contentWindow._unusedHeightValue = this._unusedHeight;
+    }
+
+    _measureAsyncTimeout() {
```
rniwa, line 193: Since we're using rAF, we should probably call this _measureAsyncEnd or something.

resources/benchmark-runner.mjs
```diff
@@ -116,6 +116,106 @@ class PageElement {
...
+        this._measureForceLayout();
+        const endTime = performance.now();
+        this.asyncTime = endTime - this._asyncStartTime;
+        performance.mark(this._asyncEndLabel);
+        this._asyncDoneCallback();
+    }
+
+    _measureAsyncRaf() {
+        this._measureForceLayout();
```
rniwa, line 202: I don't think we need to force the layout update here since all browsers implement update the rendering step these days:
https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering

resources/benchmark-runner.mjs
rniwa: Yeah, I think we need to start setTimeout(~, 0) here and measure that as the end of the async time.

resources/benchmark-runner.mjs
```diff
@@ -116,6 +116,106 @@ class PageElement {
     }
 }
 
+class MeasureTask {
```
rniwa, line 119: I'd call this MeasureStep or StepRunner since we'd be running & measuring individual BenchmarkStep here.

resources/benchmark-runner.mjs
rniwa: Conceptually, two rAF callbacks runs back to back. We want to do this so that any async tasks scheduled during the sync step would run before our timer fires.

rAF1 -> measure sync, this can schedule async tasks with setTimeout(~, 0), postMessage, etc...
rAF2 -> setTimeout(~, 0) - end-timer

This guarantees that end-timer runs after all other async tasks scheduled in rAF1.

rniwa:
Cross posting here but I wonder we want to measure the time of subsequent rAF as well since some frameworks are delaying the work to the next rAF. In other words, we do something like this:

rAF1 -> schedules rAF3 and run sync step & start async timer; this can schedule rAF4, etc...
rAF2 -> run setTimeout(~ , 0) to measure the end of async and schedules rAF5
--- the rest of update the rendering happens here ---
timer scheduled by framework, etc... runs here
timer in rAF2 fires here. This is the end of "async" time.

rAF3 -> The start of "subsequent rAF time"
rAF4 -> framework code that supposed to run in rAF runs; this can schedule more async tasks.
...
rAF5 -> schedule setTimeout(~ , 0) to measure the end of "subsequent rAF time"
--- the rest of update the rendering happens here ---
timer scheduled by framework, etc... runs here
timer in rAF5 fires here. This is the end of "subsequent rAF time" time.

smaug----:
Yeah, I've been thinking too that we should perhaps measure start and end of the next animation frame. We'd need to schedule another pair of callbacks and measure time between them.

rniwa:
We had some internal discussion about this, and we're concerned that the proposed change will make CPU throttle back the frequency too much. In practice, this will result in a paradoxical effect of the slower a software is, the faster CPU gets (because longer running task will tend to keep CPUs busier for a longer period of time and therefore keeps it at a higher frequency). We could work around this problem by warming up CPU with some workload but then that's pretty artificial as well and may have other adverse side effects like triggering thermal throttling of CPU.

The current measurement methodology of Speedometer has an advantage that it lets faster browser keep running at a high CPU frequency after finishing preceding workloads. In effect, it lets us measure the throughput of the browser engine, and not the latency for CPU to ramp up.

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,23 @@ const WarmupSuite = {
     ],
 };
 
+class RafBracketedCallbacks {
+    constructor(first_callback, second_callback) {
```
rniwa, line 167: Please use camelCase.

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,23 @@ const WarmupSuite = {
     ],
 };
 
+class RafBracketedCallbacks {
```
rniwa, line 166: rAF should probably capitalized as rAF, not Raf.

resources/benchmark-runner.mjs
```diff
@@ -178,59 +195,44 @@ class MeasureStep {
...
-            requestAnimationFrame(this._measureAsyncRafCallback);
-        } else {
-            throw new Error(`Unknown async measure mode: ${this._asyncMetricMode}`);
+        if (params.stepWaitTime) {
+            await new Promise((resolve) => {
+                setTimeout(resolve, params.stepWaitTime);
+            });
         }
+        if (this._asyncMetricMode === "timeout")
+            setTimeout(this._measureSync.bind(this), 0);
```
rniwa, line 231: Why don't we wrap this in callbacks class like TimerCallbacks which takes two arguments like RafBracketedCallbacks.

resources/benchmark-runner.mjs
```diff
@@ -245,9 +247,12 @@ class MeasureStep {
 
         this.syncTime = syncEndTime - syncStartTime;
 
-        performance.mark(this._asyncStartLabel);
         if (this._asyncMetricMode === "timeout")
             setTimeout(this._measureAsyncTimeoutCallback, 0);
+        else
+            this._rafMeasurement.run();
```
rniwa, line 253: I'm confused. Why do we need to call run again here?

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,166 @@ const WarmupSuite = {
     ],
 };
 
+class TimerCallbacks {
+    constructor(firstCallback, secondCallback) {
+        this._firstCallback = firstCallback;
+        this._secondCallback = secondCallback;
+        this._setTimeoutCallback = this._setTimeout.bind(this);
```
bgrins, line 170: Nit: I'd find it easier to follow these classes if property name was always `_setTimeoutCallback` both in the class declaration and in this statement reassigning to the bound function (`this._setTimeoutCallback = this._setTimeoutCallback.bind(this);`)

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,166 @@ const WarmupSuite = {
...
+        this._frame = frame;
+
+        this._unusedHeight = undefined;
+
+        this._wasRun = false;
+
+        this._asyncMeasurePromise = new Promise((resolve) => {
+            this._asyncDoneCallback = resolve;
+        });
+        this._rAFMeasurement = new RAFBracketedCallbacks(this._measureRafStart.bind(this), this._measureRafEnd.bind(this));
```
smaug----, line 215: Something here is rather confusing. Using RAFBracketedCallbacks in two places with different callbacks.
Why do we need this kind of setup?

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,166 @@ const WarmupSuite = {
     ],
 };
 
+class TimerCallbacks {
+    constructor(firstCallback, secondCallback) {
```
rniwa, line 167: Call these as first & second are rather confusing.
We should probably call them as syncCallback & asyncCallback.

rniwa:
This is all so complicated to review. I think we should split this into two pieces.

1. Refactoring of callbacks
2. Addition of rAF-based measurement

rniwa:
> This is getting all so complicated to review. I think we should split this into two pieces.
> 
> 1. Refactoring of callbacks
> 2. Addition of rAF-based measurement

Alright, I've gone ahead and made a PR: https://github.com/WebKit/Speedometer/pull/164

rniwa:
We've made the equivalent changes in https://github.com/WebKit/Speedometer/pull/173 and https://github.com/WebKit/Speedometer/pull/164

Issue #30, opened 2022-12-09
`benchmark-report.js` is broken


squarewave:
We're looking to get a head start on setting this up in CI, and historically we've used `benchmark-report.js` to drive the execution. However it's broken now because we pulled all of the global functions like `startBenchmark` to be part of `main.js`'s MainBenchmarkClient. Is the plan to add a base class to this that `benchmark-report.(m)js` can subclass and call `startBenchmark` on? It would be nice if we could nail down a plan for this in the near term so that our devs working on CI can get something stood up.

EDIT: I had `benchmark-runner.js` confused with `benchmark-report.js` - fixed now.

housseindjirdeh:
@squarewave Are you planning to set up a unit or integration testing pipeline in CI? I was about to start going down that path, but don't want to duplicate efforts.

squarewave:
We were just planning to get a head start on setting up benchmarking of Firefox in Mozilla's CI. Historically we've used what `benchmark-runner.js` sets up to make extracting scores easier for us.

squarewave:
Woops! I've been typoing the name in this issue. I am actually referring to `benchmark-report.js`.

camillobruni:
Is this still broken for you?

rniwa:
This appears to be functional now. Please comment if it's still broken.

PR #31, opened 2022-12-14
Fix interactive runner

camillobruni:
This partially addresss issue #30

- Make the client callbacks async
- Move the interactive runner into a separate interactive.mjs module


camillobruni:
I will expose the runner again on the window in a follow-up cl for the standard runner as well.

bgrins:
@camillobruni @squarewave as far as I can tell this did fix the InteractiveRunner.html when opening in a browser tab, but https://github.com/WebKit/Speedometer/blob/main/resources/benchmark-report.mjs#L23 seems broken - we're getting `Uncaught ReferenceError: startBenchmark is not defined` when attempting to stand this up with the following patch, then `python -m http.server 8123` and loading `http://0.0.0.0:8123/index.html?raptor`:

```patch
diff --git a/resources/benchmark-report.mjs b/resources/benchmark-report.mjs
index 03831e9..a5e63ce 100644
--- a/resources/benchmark-report.mjs
+++ b/resources/benchmark-report.mjs
@@ -1,7 +1,10 @@
 // This file can be customized to report results as needed.
 
 (function () {
-    if (!window.testRunner && location.search != '?webkit' && location.hash != '#webkit')
+    if (
+        (!window.testRunner && location.search != '?webkit' && location.hash != '#webkit')
+        && location.search != "?raptor"
+    )
         return;
 
     if (window.testRunner)
diff --git a/resources/runner.js b/resources/runner.js
new file mode 100644
index 0000000..7a4faad
--- /dev/null
+++ b/resources/runner.js
@@ -0,0 +1,24 @@
+
+(function () {
+	var PerfTestRunner = {
+		prepareToMeasureValuesAsync: () => {},
+		measureValuesAsync: () => {}
+	}
+
+	PerfTestRunner.finishTesting = function () {
+		window.sessionStorage.setItem(
+			'benchmark_results',
+			JSON.stringify(this.values)
+		);
+	};
+
+	PerfTestRunner.reportValues = function (testInfo, values) {
+		this.values.push([testInfo, values]);
+		console.log("reporting");
+		if (!testInfo.continueTesting) {
+			this.finishTesting();
+		}
+	};
+
+	window.PerfTestRunner = PerfTestRunner;
+})();
```

I'm trying to figure out what's going on here with the `benchmarkClient` global - should benchmark-report be updated to somehow override what's happening in the new interactive.mjs, or does it just need to load that file, or something else?

resources/interactive.mjs
```diff
@@ -0,0 +1,186 @@
...
+function startTest()
+{
+    const queryParam = parseQueryString['suite'];
+    if (queryParam !== undefined)
+        disableAllSuitesExcept(queryParam);
+
+    const benchmarkClient = {
+        stepperPromise: undefined,
+        stepper: undefined,
+        step() {
```
squarewave, line 118: Nit: maybe rename this to `enableStepping`? It's a bit awkward reading this right now.

resources/interactive.mjs
```diff
@@ -0,0 +1,186 @@
...
+        if (keyValue.length == 1)
+            pairs[keyValue[0]] = '';
+        else
+            pairs[keyValue[0]] = decodeURIComponent(keyValue[1].replace(/\+/g, ' '));
+    }
+    return pairs;
+})(window.location.search.substr(1).split('&'));
+
+function disableAllSuitesExcept(suiteName)
+{
```
squarewave, line 102: Nit: I think this styling is inconsistent? I don't particularly care or even have the firmest grasp on what the styling rules are, but once we actually get the code formatting stuff landed it would be nice to minimize noise in the blame record from the autoformatting step.

squarewave (approved):
This looks good to me! Thanks for taking it. I noticed while running it that it behaves slightly different when you click step while in the middle of a run. (This does something, the other one just kept on going.) I really don't think that matters much though, especially assuming we're going to eventually scrap this UI for something less... the way that it is right now.

resources/interactive.mjs
camillobruni: From https://webkit.org/code-style-guidelines/#braces i think functions should have braces on new lines (I might be wrong)

resources/interactive.mjs
camillobruni: refactored the code to be easier to undestand.

jrmuizel:
Does this need something like the following for the disabling of tests to actually work?

```patch
diff --git a/resources/benchmark-runner.mjs b/resources/benchmark-runner.mjs
index a8ddeab..f73dbae 100644
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -188,21 +188,22 @@ export class BenchmarkRunner {
 
     async _runAllSuites()
     {
         this._measuredValues = {tests: {}, total: 0, mean: NaN, geomean: NaN, score: NaN};
 
         this._removeFrame();
         this._appendFrame();
         this._page = new Page(this._frame);
 
         for (const suite of this._suites)
-            await this._runSuite(suite);
+            if (!suite.disabled)
+                await this._runSuite(suite);
 
         // Remove frame to clear the view for displaying the results.
         this._removeFrame();
         await this._finalize();
     }
 
     async _runSuite(suite)
     {
         await this._prepareSuite(suite);
         for (const test of suite.tests)

```

camillobruni:
Correct, I think I accidentally broke this. Adding this piece now.

camillobruni:
Please have a look again.
- Support stepping / running and switching between the two states
- Updated the code to use URLSearchParams instead of custom parsing
- Skipping over disabled stories again

resources/interactive.mjs
rniwa: Yes. Functions should have braces on new line.

resources/interactive.mjs
```diff
@@ -0,0 +1,214 @@
...
+                const resolve = this._stepPromiseResolve;
+                this._stepPromise = this._newStepPromise();
+                resolve();
+            }
+        },
+        _newStepPromise()
+        {
+            return new Promise(resolve => {
+                this._stepPromiseResolve = resolve;
+            })
```
rniwa, line 138: Missing a semicolon here.

resources/interactive.mjs
```diff
@@ -0,0 +1,213 @@
...
+        isStepping: false,
+        willStartFirstIteration()
+        {
+            if (this.isRunning)
+                throw Error('Runner was not stopped before starting;');
+            this.isRunning = true;
+            if (this.isStepping)
+                this._stepPromise = this._newStepPromise();
+        },
+        step()
```
rniwa, line 123: It's a bit confusing that this is a part of `benchmarkClient` since it's not a part of the client interface.
Maybe it's better to define this as a lambda, which knows how to update variables of `benchmarkClient`.
WDYT?

housseindjirdeh (approved):

resources/interactive.mjs
```diff
@@ -0,0 +1,248 @@
+import {BenchmarkRunner} from './benchmark-runner.mjs';
+import {Suites} from './tests.mjs';
+
+class InteractiveBenchmarkRunner extends BenchmarkRunner {
+    _stepPromise = undefined;
+    _stepPromiseResolve = undefined;
+    _isRunning = false;
+    _isStepping = false;
+
+    constructor(suites, iterationCount) {
```
rniwa, line 10: Nit: { should be on new line.

resources/interactive.mjs
```diff
@@ -0,0 +1,248 @@
...
+        super(suites);
+        this._client = this._createClient();
+        if (!Number.isInteger(iterationCount) || iterationCount <= 0)
+            throw Error("iterationCount must be a positive integer.");
+        this._iterationCount = iterationCount;
+    }
+
+    _createClient() {
```
rniwa, line 18: Ditto.

resources/interactive.mjs
```diff
@@ -0,0 +1,248 @@
...
+        return  {
```
rniwa, line 19: Nit: There are two spaces between return and {.

rniwa (approved):

PR #32, opened 2022-12-16
fixed formatting

ivoinestrachan:


bgrins:
We're going to start the process of auto formatting in https://github.com/WebKit/Speedometer/issues/56. So thanks for the interest, but going to close this as we'll ultimately format code using tools set up in the repo rather than one-offs.

PR #38, opened 2023-01-17
Adds unit tests for `BenchmarkRunner`

housseindjirdeh:
This PR does the following:

- Adds a `tests/` directory with a [`benchmark-runner.test.js`](https://github.com/WebKit/Speedometer/compare/main...housseindjirdeh:benchmark-runner-tests?expand=1#diff-95929dff87ba0581fe4fec77ff690ff36a8c126dadedb88fda9d7e198335453b) file that includes unit tests for `BenchmarkRunner`
- Adds a `package.json` file to specify that the application is using ES Module syntax (this is necessary in order for Jest to recognize [ESM](https://jestjs.io/docs/ecmascript-modules))  
- Adds a `jest.config.json` file to define Jest configurations

To run tests locally:

1. `npm install jest jest-environment-jsdom --save-dev`
2. `node --experimental-vm-modules node_modules/jest/bin/jest.js`

---

Note: To make this PR easier to read, I have not included the entire `node_modules/` directory that contains the required `jest` and `jest-environment-jsdom` dependencies. I can include those in a follow up PR.

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -1,0 +1,4 @@
+{
+    "testEnvironment": "jsdom",
+    "testPathIgnorePatterns": ["<rootDir>/resources", "<rootDir>/node_modules/"]
+}
--- a/package.json
+++ b/package.json
@@ -1,0 +1,3 @@
+{
+    "type": "module"
+}
--- a/tests/benchmark-runner.test.js
+++ b/tests/benchmark-runner.test.js
@@ -1,0 +1,332 @@
+import { jest } from "@jest/globals";
+import { BenchmarkRunner } from "../resources/benchmark-runner";
+
+jest.useFakeTimers();
+
+const TEST_FIXTURE = (name) => ({
+    name,
+    run: () => jest.fn(),
+});
+
+const SUITES_FIXTURE = [
+    {
+        name: "Suite 1",
+        url: "todomvc/vanilla-examples/vanillajs/index.html",
+        tests: [
+            TEST_FIXTURE("Test 1"),
+            TEST_FIXTURE("Test 2"),
+            TEST_FIXTURE("Test 3"),
+        ],
+    },
+    {
+        name: "Suite 2",
+        url: "todomvc/vanilla-examples/vanillajs/index.html",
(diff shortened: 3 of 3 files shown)


jest.config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "testEnvironment": "jsdom",
```
rniwa, line 2: I don't think we should be using jsdom.
We should just run the test in a web browser.

housseindjirdeh: Unfortunately Jest can't run in a web browser on its own. There are some options we can consider here (but please let me know if there's any other options I may not be considering):

1. Use something like Jest + JSDOM for small, focused unit tests (like this PR) and then work on to include larger end-to-end integration tests that will run in a web browser.
2. Stick to Jest for small, focused unit tests but remove JSDOM as a test environment. Any tests that need to be run in a browser can be included in our integration testing setup
3. Run all unit tests in the browser (e.g. with something like [`jest-puppeteer`](https://jestjs.io/docs/puppeteer)) regardless of whether we decide to include end-to-end integration tests in the near future
4. Stick to writing larger end-to-end integration tests in a web browser and don't include any small, focused unit tests whatsoever

_Option 1_ was my current understanding after our discussions hence why I decided to use JSDOM for simplicity here (plus the developer experience of Jest is pretty nice especially if we want to include unit testing as step in a CI setup). I personally think that's the best option but would love to hear everyone's thoughts. 

I would be open to considering _Option 3_ but we would likely have to switch to another testing tool (Selenium? Karma?) to ensure that we're not only running in Chromium, but I worry that this might be overkill for all the tiny unit tests that we write **and** it might blur the lines between how we differentiate unit tests and larger end-to-end tests.

rniwa: We should do 3. [mocha.js](https://mochajs.org) works in a browser, and we've [used it](https://github.com/WebKit/WebKit/blob/main/Websites/perf.webkit.org/browser-tests/index.html#L7) in another project in WebKit.

housseindjirdeh: I ported over all the tests to Mocha in #40. If we all agree on 3, than we can close this PR.

PR #40, opened 2023-01-24
Adds unit tests for `BenchmarkRunner` using Mocha

housseindjirdeh:
This PR does the following:

- Adds a `tests/` directory with:
  - [`benchmark-runner-tests.mjs`](https://github.com/WebKit/Speedometer/compare/main...housseindjirdeh:benchmark-runner-tests-mocha?expand=1#:~:text=tests/benchmark%2Drunner%2Dtests.mjs) file that includes unit tests for `BenchmarkRunner`
  - [`tests/index.html`](https://github.com/WebKit/Speedometer/compare/main...housseindjirdeh:benchmark-runner-tests-mocha?expand=1#diff-90b511c392b35b916f8651ba3b27fea4341761dd54aedaec91c52a757e545667) file to run and display tests all in the browser (note: you'll have to spin up a local server to prevent module files from being blocked by CORS)

---

### DIFF
--- a/tests/benchmark-runner-tests.mjs
+++ b/tests/benchmark-runner-tests.mjs
@@ -1,0 +1,323 @@
+import { BenchmarkRunner } from "../resources/benchmark-runner.mjs";
+
+function TEST_FIXTURE(name)
+{
+    return {
+        name,
+        run: sinon.stub(),
+    }
+};
+
+const SUITES_FIXTURE = [
+    {
+        name: "Suite 1",
+        tests: [
+            TEST_FIXTURE("Test 1"),
+            TEST_FIXTURE("Test 2"),
+            TEST_FIXTURE("Test 3"),
+        ],
+    },
+    {
+        name: "Suite 2",
+        tests: [TEST_FIXTURE("Test 1")],
+    },
+];
+
+const CLIENT_FIXTURE = {
+    willRunTest: sinon.stub(),
+    didRunTest: sinon.stub(),
+    didRunSuites: sinon.stub(),
+};
+
+function stubPerformanceNowCalls(syncStart, syncEnd, asyncStart, asyncEnd) 
+{
+    sinon
+        .stub(window.performance, "now")
+        .onFirstCall()
+        .returns(syncStart) // startTime (sync)
+        .onSecondCall()
+        .returns(syncEnd) // endTime (sync)
+        .onThirdCall()
+        .returns(asyncStart) // startTime (async)
+        .returns(asyncEnd); // endTime (async)
+}
+
+describe("BenchmarkRunner", () => {
+    const { spy, stub, assert } = sinon;
+    let runner;
+
+    before(() => {
+        runner = new BenchmarkRunner(SUITES_FIXTURE, CLIENT_FIXTURE);
+    });
+
+    it("should be defined", () => {
+        expect(runner).not.to.be(undefined);
+    });
+
+    describe("_writeMark", () => {
+        it("should fire performance.mark with the correct name", () => {
+            const performanceMarkSpy = spy(window.performance, "mark");
+            runner._writeMark("Button clicked");
+            assert.calledWith(performanceMarkSpy, "Button clicked");
+        });
+    });
+
+    describe("Frame", () => {
+        describe("_removeFrame", () => {
+            let frame, removeChildSpy;
+
+            before(async () => {
+                frame = await runner._appendFrame();
+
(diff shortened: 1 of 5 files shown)


tests/index.html
```diff
@@ -0,0 +1,38 @@
...
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.2.0/mocha.min.css">
+    <style>
+        body {
+            margin: 8px;
+        }
+    </style>
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.2.0/mocha.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/expect.js/0.2.0/expect.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/sinon.js/15.0.1/sinon.min.js"></script>
```
housseindjirdeh, line 16: My initial understanding was that we didn't want a top-level package.json but per our last discussion it sounds like that would be okay? If that's the case, I can install and include these dependencies locally instead of using a CDN.

tests/benchmark-runner-tests.mjs
```diff
@@ -0,0 +1,323 @@
...
+                const createElementSpy = spy(document, "createElement");
+
+                const frame = await runner._appendFrame();
+                expect(frame).to.be.a(HTMLIFrameElement);
+                assert.calledWith(
+                    createElementSpy,
+                    frame.nodeName.toLowerCase()
+                );
+
+                expect(frame.style.width).to.be("800px");
```
rniwa, line 96: Maybe use computedStyle instead?

tests/index.html
bgrins: Yes, correct. I think doing this through npm makes sense at this point

bgrins:
@housseindjirdeh we could move wiring this up to CI into another PR, but relevant to the discussion about cross browser integration tests in CI it appears we can run in macos-latest on github actions (https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/) so presumably we can use Selenium + Mocha in a Github Action to target all three browsers.

In this case I guess we'd be starting a web server and then opening tests/index.html in each browser and checking for some signal of success in the page itself?

housseindjirdeh:
> @housseindjirdeh we could move wiring this up to CI into another PR, but relevant to the discussion about cross browser integration tests in CI it appears we can run in macos-latest on github actions (https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/) so presumably we can use Selenium + Mocha in a Github Action to target all three browsers.
> 
> In this case I guess we'd be starting a web server and then opening tests/index.html in each browser and checking for some signal of success in the page itself?

+1 to wiring it up in CI in a separate PR (and I can start that once this lands). And yep, in the meantime you'll have to start a local web server and open `tests/index.html` on a browser to run and view all test results

tests/benchmark-runner-tests.mjs
```diff
@@ -0,0 +1,323 @@
...
+            let _writeMarkSpy, _testFnSpy, page;
+            const callback = stub();
+            const suite = SUITES_FIXTURE[0];
+
+            before(async () => {
+                page = { _frame: await runner._appendFrame() };
+                _writeMarkSpy = spy(runner, "_writeMark");
+                _testFnSpy = suite.tests[0].run.callsFake(() => null);
+
+                stubPerformanceNowCalls(8000000, 10000000, 12000000, 13000000);
```
rniwa, line 239: Do we really need to use these very large numbers?
How about something like 8000 to make it easier to read?

rniwa (approved):
I'm a bit concerned that these tests might be too granular. I'm afraid It would make refactoring these code harder / a lot of work in the future.

Having said that, we'd have to start from somewhere so this seems okay.

bgrins (approved):

tests/benchmark-runner-tests.mjs
housseindjirdeh: Ah good call, updated!

tests/benchmark-runner-tests.mjs
housseindjirdeh: Yeah definitely not necessary, updated :)

housseindjirdeh:
Yeah I'm totally fine with removing some of these tests and/or making them less granular in the near future, but I thought it would be a good idea to start with more test coverage rather than less :)

PR #41, opened 2023-01-25
Customise viewport size

camillobruni:


camillobruni:
* Increase the viewport size from 800px by 600px to 1500px by 800px.
* Use the rendered size of the viewport when calculating whether the window size is too small

flashdesignory (approved):
LGTM 👍  - it's challenging not to hard-code values in too many places and to be consistent.

bgrins:
I agree with this in spirit, though would like to check with @Bas-moz that this is going to work in our CI environment. Correct me if I'm wrong, but did we decide to switch to 1366x768 to tighten it up a little bit and match a common desktop resolution?

camillobruni:
PTAL again

tests/benchmark-runner-tests.mjs
```diff
@@ -83,7 +83,9 @@ describe("BenchmarkRunner", () => {
         });
 
         describe("_appendFrame", () => {
-            it("should create an absolutely positioned iframe with 800px x 600px dimensions", async () => {
+            const DEFAULT_WIDTH = 1366;
```
flashdesignory, line 86: I like the refactor a lot!!
Could we re-use the params class and add a failsafe for searchParams if needed? 
Alternatively we could export the hardcoded values from that file and have it all live in the same place? 🤷

resources/params.mjs
```diff
@@ -0,0 +1,42 @@
+
+class Params {
+    viewport = {
+        width: 1366,
```
rniwa, line 4: Can we still with 800x600 for now?

bgrins: +1 to decoupling the size change into another issue/PR, then we can get the refactor landed now

tests/benchmark-runner-tests.mjs
camillobruni: DONE: Added a `defaultParams` export on the params.mjs module.

bgrins (approved):

rniwa (approved):

PR #42, opened 2023-01-25
Use Metrics Visualisations from GrandPrix

camillobruni:
- Use bar-chart to show totals of each iteration
- Use scatter-plots for each metrics and submetrcs
- Use more CSS color variables
- Update scrollbar to match existing color scheme


camillobruni:
Example views:
details
metrics

camillobruni:
This should be ready for review now. Happy to take feedback.

rniwa:
"Details" toggle seems to be broken at least in Safari when you click on "Details" although it works when I click on the arrow.

rniwa:
It's a bit confusing that distinct steps and "sync" / "async" times use the same colors. Can we use different colors for "sync" "async" so it's less confusing. Can we also use different shapes for each item so that we can differentiate each step and sync/async in grayscale? Making it totally dependent on color isn't accessible.

It's also a bit confusing that the bar graphs show iterations on x-axis whereas scatter charts underneath it use x-axis as values. Can we stick to either direction representing one thing? e.g. always use x-axis for values.

resources/main.css
```diff
@@ -2,17 +2,41 @@
     --viewport-width: 800px;
     --viewport-height: 600px;
     --foreground: rgb(235, 235, 235);
+    --inactive-color: rgb(128, 128, 128);
```
flashdesignory, line 5: I like the use of custom properties and it seems that we're introducing these progressively. 
Right now it feels slightly inconsistent between colors for example. 
This might be out of scope here, but maybe we can address a consistent use in a separate pr?

resources/metric-ui.mjs
```diff
@@ -0,0 +1,221 @@
...
+        const unit = subMetric.unit;
+        let width = subMetric.delta;
+        let center = mean;
+        if (normalize) {
+            width = (subMetric.delta / mean) * toPercent;
+            center = 0;
+        }
+        let left = center - width / 2;
+        const y = metricIndex;
+        const label = `Mean: ${subMetric.valueString}\n` + `Min: ${subMetric.min.toFixed(2)}${unit}\n` + `Max: ${subMetric.max.toFixed(2)}${unit}`;
```
flashdesignory, line 105: nit: Since we're using template literals, the plus sign can get removed I believe.

resources/metric-ui.mjs
```diff
@@ -0,0 +1,221 @@
...
+        points.push(rect);
+        for (let i = 0; i < values.length; i++) {
+            const value = values[i];
+            let x = value;
+            const normalized = (value / mean - 1) * toPercent;
+            const sign = normalized < 0 ? "" : "+";
+            if (normalize)
+                x = normalized;
+            const y = metricIndex + i / values.length;
+            let label = `Iteration ${i}: ${value.toFixed(2)}${unit}\n` + `Normalized: $mean${sign}${normalized.toFixed(2)}%`;
```
flashdesignory, line 118: nit: Since we're using template literals, the plus sign can get removed I believe.

resources/metric-ui.mjs
```diff
@@ -0,0 +1,221 @@
...
+        const values = subMetric.values;
+        points.push(rect);
+        for (let i = 0; i < values.length; i++) {
+            const value = values[i];
+            let x = value;
+            const normalized = (value / mean - 1) * toPercent;
+            const sign = normalized < 0 ? "" : "+";
+            if (normalize)
+                x = normalized;
+            const y = metricIndex + i / values.length;
```
flashdesignory, line 117: Should this be `(metricIndex + 1) / values.length` ?

resources/metric-ui.mjs
camillobruni: I've added this for future formatting. We're currently at a rather generous 250 char line limit. Probably 80 or 100 are more convenient at which point template-string formatting is not very nice to read.

resources/main.css
camillobruni: Let me create a PR for the last color we're missing.

resources/metric-ui.mjs
camillobruni: This is correct, see the first comment of this method. Sadly auto-formatting removed the explicit parenthesis I put :).
I'll use a separate variable to make this look more intentional.

rniwa:
Hm... the layout is broken in Safari:
Screenshot 2023-02-21 at 10 43 16 AM

camillobruni:
Let me double check, might be that the graphs are a tiny bit too wide with the deafault 800x600 viewport.

rniwa:
I'm getting this error now:
```
TypeError: undefined is not an object (evaluating 'params.suites.length')
```

camillobruni:
- Layout should be fixed now. 
- I couldn't repro the params.suites.length error (I've merged the main branch again, maybe there was something missing)

rniwa:
Can we use a real unit for x-axis instead of %? I'd also prefer if x-axis are all same scale across tests. Otherwise, it's hard to make use of the charts.

camillobruni:
I've discussed this a bit internally since I was also a bit torn between absolute vs. normalized values.
Ultimately I wanted to focus on noise rather than absolute values:

- Given the rather large range of values (EmberJS-Debug vs VanillaJS for instance), absolute values have a bit of downside in terms of comparability (async vs sync has similar issues, and you end up with the values cramped down into each end of plot).
- Given that we use geomean for the total, relative impact of test matters more than it's absolute value, that would be another point in favour of normalised  values.
- Printing a new x-axis scale for each each sub-/metric would use up some additional screen-real-estate
- The tooltip currently shows the absolute values for each point

My plan is to add a separate compare.html helper page to compare the results of two different runs with side-by-side scatter-plots (most likely using absolute values). Would that accommodate your use case better?

camillobruni:
Example with absolute values:
Screenshot 2023-02-21 at 21 16 20
Screenshot 2023-02-21 at 21 16 35

camillobruni:
- Switched to absolute values by default
- Clicking on any charts toggles all of them between absolute and normalized values

rniwa:
> * Switched to absolute values by default
> * Clicking on any charts toggles all of them between absolute and normalized values

Nice. Being able to switch between two modes is useful. Hovering over each data point to see each data point's raw value seems too hard to do in practice though. Many points are overlapping one another or at the edges.

How about toggling between scatter plots & tabular view. We have an internal performance dashboard, which uses horizontal bar graphs to show variance across different subtests:
Screenshot 2023-02-21 at 9 59 55 PM

Clicking on one of those charts expand it to show the data in tabular form:
Screenshot 2023-02-21 at 10 00 48 PM

This UI is more optimized for comparing two results (A vs B here) but I find this way of visualization & ability to drill down to individual iteration's data to be highly useful.

bgrins:
I think this looks like a great improvement, thanks @camillobruni!

> How about toggling between scatter plots & tabular view. We have an internal performance dashboard, which uses horizontal bar graphs to show variance across different subtests

I agree that looks quite nice, but I wouldn't want to prevent incremental improvements from landing before a new UI was developed. Can this be moved into a follow up issue, where we can discuss in more detail (including the idea of a compare UI)?

bgrins:
Another idea we could discuss in a follow up is using a dialog element or another overlay to present the detailed info (so that it's not constrained within the default viewport for the runner itself). I did this in a prototype used to research data grids at https://bgrins.github.io/data-ui-tests/?autorun and found it was a nice way to surface more detailed info than trying to fit it into the rest of the page.

camillobruni:
I like rniwa's suggestion, I definitely use this approach on a separate evaluation page to compare between multiple runs.

Given that we can easily download the data now, the details page should mostly help with assessing the run quality (aka how much noise there is and how data is dstributed).
The scatter plots compress a lot of information into a small area:
- It's easy to see whether we get quantized data
- It's easy to see if you get outliers
- It's easy to see if there is a drift over time

- Maybe a CSV (or tab-separate) download to be easily pasted into a sheet would help with detailed analysis as well?
- I can add a collapsable inline table as well to address the lack of easily reading individual data points

rniwa:
Hm... can we flip the rows & columns of the tables? Test names are too compressed & horizontal scrolling bar is quite painful to use.
 
Screenshot 2023-03-08 at 4 58 45 PM

resources/metric-ui.mjs
```diff
@@ -0,0 +1,292 @@
...
+    const normalizedScatterPlot = renderScatterPlot(scatterPlotParams);
+    scatterPlotParams.values = prepareScatterPlotValues(metrics, false);
+    scatterPlotParams.unit = metrics[0].unit;
+    scatterPlotParams.xAxisLabel = metrics[0].unit;
+    const absoluteScatterPlot = renderScatterPlot(scatterPlotParams);
+    const legend = metrics
+        .map(
+            (metric, i) => `
+                <tr >
+                    <td class=${colors[i % colors.length]} >●</td>
```
rniwa, line 20: Are we sure all these values come from trusted sources and not via query parameters, etc...?
I don't want us to have XSS vulnerabilities.

bgrins:
> Test names are too compressed & horizontal scrolling bar is quite painful to use.

Another thing we could do is when in the details view set `--viewport-width: 90vw; --viewport-height: 90vh`. I did this locally when looking at the PR and it's much easier to use than when stuck in the test frame size.

camillobruni:
- I've now prepared PR #92 to make the details view easier to customise in size (and maximise it)
- I'd prefer the current table layout, so I can easily double check what the outlier values are by scanning across the cell backgrounds
- Hopefully the larger details view should fix this issue

resources/metric-ui.mjs
camillobruni: - I've  now renamed `params` to `viewParams` to make a bit more clear that they are not related.
- params.mjs parses all input data (except for `params.suites` ) to int and bool
- params.suites is checked against a known list of Suites.keys() in runBenchmark

The input vaues to renderMetricView are only extracted from the metrics themselves which have values coming only from the runner.

resources/metric-ui.mjs
```diff
@@ -136,8 +136,8 @@ function renderMetricsTable(metrics, min, max) {
             ${columns}
         </tr>`;
     }
-    return `<table class="metrics-table">
-        <thead>
+    return `<table class="metrics-table" >
+        <thead  onclick="this.classList.toggle('nowrap')" >
```
rniwa, line 140: Nit: two spaces between thead and onclick.

rniwa (approved):
I think we want to do a bunch of style tweaks but that can be done separately.

resources/metric-ui.mjs
flashdesignory: the current format script "should" catch this

camillobruni: DONE

PR #45, opened 2023-01-26
Run mocha test in CI using selenium

bgrins:


.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: CI
+
+on:
+  push:
+    branches: [ "main", "tests-in-ci" ]
```
bgrins, line 5: Would remove this branch before committing - this is allowing results to show up in https://github.com/bgrins/Speedometer/actions

bgrins:
@housseindjirdeh a little rough but this is what I was thinking. Open to suggestion on making the actual runner better but it seems to work alright.

bgrins:
This can be run locally with for example `BROWSER=firefox node tests/unit-test-run.mjs`. Browser specific notes:
- Firefox requires geckodriver https://github.com/mozilla/geckodriver/releases. On mac you can install with `brew install geckodriver` as well.
- Safari requires you to enable safaridriver via https://developer.apple.com/documentation/webkit/testing_with_webdriver_in_safari.
- I think it finds Chrome without any additional install - would be curious if it works for you locally.

tests/unit-test-run.mjs
```diff
@@ -0,0 +1,64 @@
...
+async function test() {
+  let capabilities;
+  switch (process.env.BROWSER) {
+    case "safari":
+      capabilities = Capabilities.safari();
+      capabilities.set("safari.options", { technologyPreview: false });
+      break;
+
+    case "firefox": {
+      capabilities = Capabilities.firefox().setLoggingPrefs({ browser: "ALL" });
```
rniwa, line 24: Doesn't enabling extra logging  affect perf?

tests/unit-test-run.mjs
```diff
@@ -0,0 +1,64 @@
+
+import serve from "./unit-test-server.mjs";
+import {
+  Builder,
```
rniwa, line 4: Please use 4-space indentation per coding style guideline.

tests/unit-test-run.mjs
bgrins: Maybe, but this isn't actually running the benchmark or reporting results, this is only running the unit tests

tests/unit-test-run.mjs
bgrins: OK. Looking forward to auto formatting

tests/unit-test-run.mjs
bgrins: I'll just remove that from all of the browsers so it's consistent though, as I'm not seeing an observable change with or without it

tests/unit-test-run.mjs
```diff
@@ -0,0 +1,67 @@
+import serve from "./unit-test-server.mjs";
+import { Builder, By, Capabilities } from "selenium-webdriver";
```
housseindjirdeh, line 2:
```suggestion
import { Builder, Capabilities } from "selenium-webdriver";
```

tests/unit-test-server.mjs
```diff
@@ -0,0 +1,50 @@
+// Simple server adapted from https://developer.mozilla.org/en-US/docs/Learn/Server-side/Node_server_without_framework:
```
housseindjirdeh, line 1: Nit: I assume we'll likely continue to use our testing setup even when we introduce larger end-to-end tests? If that's the case, maybe we can just rename these files to `run.mjs` and `server.mjs`?

housseindjirdeh (approved):
Just minor nits. Otherwise this looks great 🥳 

> I think it finds Chrome without any additional install - would be curious if it works for you locally.

It does, I just had to update my outdated chromedriver and it worked (`brew upgrade chromedriver`)

tests/unit-test-server.mjs
bgrins: Yeah, we may want to pull in some dependencies if this gets more complex but the generic names make sense.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: CI
+
+on:
+  push:
```
rniwa, line 4: Use 4 space indentation?

rniwa (approved):

.github/workflows/test.yml
bgrins: I just copied this from github actions template, I think 2 space is conventional here

PR #47, opened 2023-01-26
Fix typo in CSS property name, in InteractiveRunner.html

dholbert:
InteractiveRunner.html has a CSS rule which includes a "list-position" property, which is not actually the name of a property.

The intent was surely to use "list-style-position".  Let's just correct it to that.

### DIFF
--- a/InteractiveRunner.html
+++ b/InteractiveRunner.html
@@ -6,7 +6,7 @@
 <style>
 iframe { border: 1px solid black; }
 ol { list-style: none; margin: 0; padding: 0; }
-ol ol { margin-left: 2em; list-position: outside; }
+ol ol { margin-left: 2em; list-style-position: outside; }
 .running { text-decoration: underline; }
 .ran { color: grey; }
 nav { position: absolute; right: 10px; height: 600px; }


dholbert:
Alternately, we could remove the declaration, since it's inert.  Let me know if that's preferred.  I'm not sure it matters too much, since InteractiveRunner.html seems to be a developer-facing quick-and-dirty toolbox, rather than something with pixel-perfect UX.

I'm not entirely sure if/where it's actually intended to be used. Git blame for this line just goes back to https://github.com/WebKit/Speedometer/commit/a1e842164bc1803be6afa658bbb1336ecc726b12 "Add Speedometer 2.1 content", and I couldn't immediately find where to dig through the Speedometer 2.1 change history to satisfy my curiosity about what this CSS was meant to do.

bgrins (approved):
Yeah this makes sense to just fix - I don't see where it's used either. It's likely we'll ultimately migrate the developer UI into the main runner so we can revisit the styling at that point.

bgrins:
(landing since this is a trivial update)

PR #48, opened 2023-01-26
cast iterationCount from url to number in interactive runner. Fixes #46

### LINKED ISSUE #46: Interactive runner doesn't allow specifying the number of iterations
It seems like there's some support for this if you manually specify it in the URL with `iterationCount` but it doesn't work for me.

bgrins:


bgrins:
My understanding is long term we want to migrate the interactive behavior to some developer mode on the main page, but in the meantime here's a fix for #46

resources/interactive.mjs
```diff
@@ -229,7 +229,7 @@ function startTest()
     if (searchParams.has('suite'))
         disableAllSuitesExcept(searchParams.get('suite'));
 
-    const iterationCount = searchParams.get('iterationCount') || 1;
+    const iterationCount = Number(searchParams.get('iterationCount')) || 1;
```
rniwa, line 232: How about just +.

rniwa (approved):

resources/interactive.mjs
bgrins: Doesn't matter to me either way (I think this code will eventually be removed/merged into the main runner)

PR #55, opened 2023-01-30
Dev setup

flashdesignory:
Purpose of this pr is to start setting up a dev environment for our development.
Some notable items:

* Node versions: pinning node versions in the package.json file - I just took the current LTS
* HTTP Server: I added a static http server since we're using JavaScript modules
* Development.md: Readme about development setup
* Testing.md: Readme with testing instructions
* package.json:
  - script to spin up server
  - scripts for testing
  - a few additional standard keys to works towards a complete package file
  - license key: this feels weird but following instructions it should be correct

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict = true
--- a/Development.md
+++ b/Development.md
@@ -17,7 +17,7 @@ In your working directory, open terminal and paste the following commands:
     npm install
 ````
 
-## Run
+## Run Development Server
 1. In your terminal run:
     ```sh
         npm run dev
--- a/Development.md
+++ b/Development.md
@@ -1,0 +1,28 @@
+# Development Environment
+
+## Prerequisite
+Node is required for local development.
+We recommend using a Node version manager to ensure a stable development environment is used. 
+Below are the required node / npm versions:
+```sh
+    "node": "18.13.0",
+    "npm": "8.19.3"
+````
+
+## Installation
+In your working directory, open terminal and paste the following commands:
+```sh
+    git clone https://github.com/WebKit/Speedometer.git
+    cd Speedometer
+    npm install
+````
+
+## Run
+1. In your terminal run:
+    ```sh
+        npm run dev
+    ````
+2. Open your browser of choice and navigate to [http://127.0.0.1:7000](http://127.0.0.1:7000)
+
+## Local Server
+Speedometer uses [http-server](https://github.com/http-party/http-server), which is a static HTTP server. Meaning it does not provide hot-reloading. By default, the dev script disables caching and local changes can be viewed by simply refreshing your browser window. Additional options of the http-server can be found [here](https://github.com/http-party/http-server#available-options).
--- a/Development.md
+++ b/Development.md
@@ -5,8 +5,8 @@ Node is required for local development.
 We recommend using a Node version manager ([nvm](https://github.com/nvm-sh/nvm) for example) to ensure a stable development environment is used. 
 Below are the required node / npm versions:
 ```sh
-    "node": "18.13.0",
-    "npm": "8.19.3"
+    "node": ">18.13.0",
+    "npm": ">8.19.3"
 ````
 
 ## Installation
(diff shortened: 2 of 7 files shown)


Development.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Installation
+In your working directory, open terminal and paste the following commands:
+```sh
+    git clone https://github.com/WebKit/Speedometer.git
+    cd Speedometer
+    npm install
+````
+
+## Run
```
rniwa, line 20: Is this about running the benchmark? If so, we should call that out.

rniwa (approved):

package.json
```diff
@@ -1,8 +1,30 @@
 {
+  "name": "speedometer",
+  "version": "3.0.0",
```
bgrins, line 3: I know we aren't actually publishing this, but perhaps it makes sense to still mark as alpha or some other way to indicate that it's not complete?

flashdesignory: I can change it to "3.0.0-alpha", or remove the key/value.

package.json
```diff
@@ -1,8 +1,30 @@
 {
+  "name": "speedometer",
+  "version": "3.0.0-alpha",
+  "description": "An open source repository for the Speedometer benchmark.",
+  "engines": {
+    "node": "18.13.0",
```
bgrins, line 6: I think this pinned version is too strict for a local development environment that just needs a static server. For example applying the PR locally I get an error:

```
npm ERR! notsup Required: {"node":"18.13.0","npm":"8.19.3"}
npm ERR! notsup Actual:   {"npm":"8.19.3","node":"v19.2.0"}
```

And while I could use nvm to version backwards it just seems like an unnecessary hoop to jump through for something that will surely work fine with either an older or newer node.

I tend to think we should just not set any requirements here, or set whatever minimum is required for the server package to work and allow anything greater than that.

bgrins (changes requested):
This looks like a good set of changes, thanks. I don't think the prerequisite is buying us anything and can probably be dropped, but happy to discuss that further as well.

package.json
flashdesignory: @bgrins  - we could just use a min version that's less restrictive. 
It might be useful to keep the package-lock.json file clean. 

I'm thinking potential issues if someone would use a lower version of node / npm for local development, installs additional packages and commits the updates. 

how about something like:
```
 "engines" : { 
    "npm" : ">=8.19.3",
    "node" : ">=18.13.0"
  }
```

Development.md
```diff
@@ -0,0 +1,28 @@
...
+    cd Speedometer
+    npm install
+````
+
+## Run Development Server
+1. In your terminal run:
+    ```sh
+        npm run dev
+    ````
+2. Open your browser of choice and navigate to [http://127.0.0.1:7000](http://127.0.0.1:7000)
```
bgrins, line 25: Locally `npm run dev` already opens the browser for me. Is that an option on the http-server CLI? I'd prob prefer to match what's described here and not auto open, but if we want to keep the current behavior then let's remove this text

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict = true
```
bgrins, line 1: What does this do? Is it still needed after the update to min versions?

package.json
bgrins: Yep, that works for me. I believe the docs need an update as well.

package-lock.json
```diff
@@ -1,14 +1,22 @@
 {
-  "name": "Speedometer",
+  "name": "speedometer",
+  "version": "3.0.0",
```
bgrins, line 3: Looks like this file should be regenerated with the new alpha version

.npmrc
bgrins: OK I see, this is what caused the bail out to happen https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json. Shame this has to be a separate file in the root of the project, but alright.

Development.md
flashdesignory: Great callout - I'll remove the flag in the script

.npmrc
flashdesignory: If we omit this it won't enforce the min versions I believe.

bgrins (approved):

PR #57, opened 2023-02-03
Add Eslint & Prettier 

flashdesignory:
# Overview
Eslint rules are based on [WebInspectorUI's Eslint implementation](https://github.com/WebKit/WebKit/blob/main/Source/WebInspectorUI/.eslintrc), with a few adjustments (Envs, Globals) to ensure we adapt to Speedometer's needs. 

Prettier has been added to establish a code formatting pattern. Rules have been adjusted to avoid conflicts with the already established rules of the eslintrc file. 

This is the base configuration for both tools, with the goal to avoid additional packages to "glue the two tools together". 
Both can be used separately, with Eslint focusing on code style of JavaScript and Prettier focusing on code formatting for all supported file formats. 

# File changes
The format script (from below) has been run on the codebase.
There are two instances, where I wasn't sure if the changes might have negative impact and therefore I ignored these.
Screen Shot 2023-02-02 at 4 40 30 PM

# Scripts
The following scripts have been added to the package.json file:

```
"lint:check": "eslint **/*.mjs",
"lint:fix": "eslint **/*.mjs --fix",
"pretty:check": "prettier --check ./",
"pretty:fix": "prettier --write ./",
"format": "npm run pretty:fix ; npm run lint:fix",
```

Note: the format script runs both sequentially, even if Prettier's run results in errors. 

# CI
The format script has also been added to the Github workflow and currently runs prior to running our tests.
It is an additional job, mainly since it only needs to run once, and not in all browser we run the tests with.

# Eslint Rules
Please let me know if we want to adjust any rules, some candidates might be:
- object-curly-spacing (do we want a space between?)
- prefer-template (it's a newer convention)


resources/benchmark-report.mjs
```diff
@@ -1,83 +1,107 @@
...
+    if (
+        !window.testRunner &&
+        location.search !== "?webkit" &&
+        location.hash !== "#webkit"
+    )
         return;
 
-    if (window.testRunner)
-        testRunner.waitUntilDone();
+    if (window.testRunner) window.testRunner.waitUntilDone();
```
rniwa, line 11: This is not right either. Each statement should be on its own line:
https://webkit.org/code-style-guidelines/#linebreaking-multiple-statements

resources/benchmark-report.mjs
```diff
@@ -1,83 +1,107 @@
...
         },
         didFinishLastIteration() {
-            document.head.removeChild(document.querySelector('style'));
+            document.head.removeChild(document.querySelector("style"));
 
             const measuredValuesByFullName = {};
             function addToMeasuredValue(value, fullName, aggregator) {
-                const values = measuredValuesByFullName[fullName] || new Array;
+                const values =
+                    measuredValuesByFullName[fullName] || new Array();
```
rniwa, line 57: What's up with this?
Are we trying to set 80 character limit or something.
We certainly shouldn't.

resources/benchmark-runner.mjs
```diff
@@ -5,15 +5,12 @@ export class BenchmarkTestStep {
     }
 }
 
-
 class Page {
-    constructor(frame)
-    {
+    constructor(frame) {
```
rniwa, line 9: This is not right either. Curly braces should be on their own line:
https://webkit.org/code-style-guidelines/#braces-function

resources/benchmark-runner.mjs
```diff
@@ -29,38 +26,33 @@ class Page {
...
-        if (element === null)
-            return null;
+        if (element === null) return null;
         return this._wrapElement(element);
     }
 
-    querySelectorAll(selector)
-    {
-        const elements = Array.from(this._frame.contentDocument.querySelectorAll(selector));
+    querySelectorAll(selector) {
```
housseindjirdeh, line 35: +1 to @rniwa's comment earlier. We'll want to make sure curly braces are in a new line to adhere to https://webkit.org/code-style-guidelines/

.github/workflows/test.yml
```diff
@@ -1,33 +1,51 @@
 name: CI
 
 on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+    push:
```
bgrins, line 4: Is this file formatted by prettier / eslint? I would have expected no changes here

flashdesignory: we can control what files get changed. 
I believe it was prettier that changed this file

bgrins: Alright. I'm fine with just letting it do its thing

InteractiveRunner.html
```diff
@@ -1,22 +1,39 @@
...
+                height: 600px;
+            }
+            nav > ol {
+                height: 100%;
+                overflow-y: scroll;
+            }
+        </style>
+    </head>
+    <body>
+        <script></script>
```
rniwa, line 37: I guess we can get rid of this empty script element? Can be done in a separate PR though.

flashdesignory: might as well do it now 😄

bgrins (approved):
Thanks Thorsten, this looks good to me

resources/benchmark-runner.mjs
rniwa: So are we reverting these curly brace changes??

housseindjirdeh (approved):

resources/benchmark-runner.mjs
rniwa: As I mentioned during today's sync, we're okay with adopting K&R style curly braces everywhere.

tests/benchmark-runner-tests.mjs
```diff
@@ -1,7 +1,6 @@
-import { BenchmarkRunner } from "../resources/benchmark-runner.mjs";
+import {BenchmarkRunner} from "../resources/benchmark-runner.mjs";
```
rniwa, line 1: I think we want to keep a space after { & before }.

flashdesignory: awesome - agreed - it's more readable. 
I'll adjust the rule!

resources/interactive.mjs
```diff
@@ -211,41 +199,34 @@ function createUIForSuites(suites, onStep, onRunSuites)
...
+        disableAllSuitesExcept(searchParams.get("suite"));
 
-    const interactiveRunner =  new window.BenchmarkRunner(Suites, params.iterationCount);
+    const interactiveRunner = new window.BenchmarkRunner(Suites, params.iterationCount);
 
     if (!(interactiveRunner instanceof InteractiveBenchmarkRunner)) {
-        throw Error(
-            "window.BenchmarkRunner must be "+
-            "a subclass of InteractiveBenchmarkRunner");
+        throw Error("window.BenchmarkRunner must be " + "a subclass of InteractiveBenchmarkRunner");
```
rniwa, line 222: Looks like this can just be a single string literal?

rniwa (approved):

resources/interactive.mjs
flashdesignory: nice catch - I can address that with the template literals update in a follow up pr

Issue #53, opened 2023-01-27
Remove "Runs / Minute" in the displayed score


bgrins:
As decided in the meeting on Jan 25. Reason: Although Speedometer includes a scoring system that shows an increase in number as an improvement, “runs per minute” isn’t actually what’s being measured and can be confusing. The decision is to keep the same measurement and overall visualization, but to make the number unitless.

bgrins:
Resolved by https://github.com/WebKit/Speedometer/pull/61

PR #61, opened 2023-02-08
Use Score instead of Runs / Minute

camillobruni:
RPM has always been a bit a confusing term in Speedometer2 as it was never real rpms, rather a scaled value.
To avoid future confusion it is renamed to the more neutral "score".


bgrins:
Thanks - this fixes https://github.com/WebKit/Speedometer/issues/53

bgrins (approved):

rniwa (approved):

PR #63, opened 2023-02-09
Use GrandPrix Metric model

camillobruni:
- Use more explicit Metric objects with parent and child-metrics
- Update statistics.mjs to use more ES6 features


resources/metric.mjs
```diff
@@ -0,0 +1,90 @@
+import * as Statistics from "./statistics.mjs";
+
+/** Number of milliseconds in a single minute. */
+export const MILLIS_PER_MIN = 60 * 1000;
+
+export class Metric {
+    /** @param {string} name */
```
rniwa, line 7: I don't think we should have these comments in the code.

resources/metric.mjs
```diff
@@ -0,0 +1,90 @@
+import * as Statistics from "./statistics.mjs";
+
+/** Number of milliseconds in a single minute. */
+export const MILLIS_PER_MIN = 60 * 1000;
```
rniwa, line 4: Please rename the variable to MILLISECONDS_PER_MINUTE and remove the comment.

resources/metric.mjs
```diff
@@ -0,0 +1,90 @@
...
+    add(value) {
+        if (typeof value !== "number")
+            throw new Error(`Adding invalid value=${value} to metric=${this.name}`);
+        this.values.push(value);
+    }
+
+    /**
+     * Calculate aggregate metrics for collected values.
+     */
+    compute() {
```
rniwa, line 77: Please call this function computeAggregatedMetrics or something and remove the comment.

resources/statistics.mjs
```diff
@@ -75,20 +75,20 @@ export function confidenceIntervalDelta(confidenceLevel, numberOfSamples, sum, s
...
-    var quantile = cdfForProbability[degreesOfFreedom - 1]; // The first entry is for the one degree of freedom.
+    const quantile = cdfForProbability[degreesOfFreedom - 1]; // The first entry is for the one degree of freedom.
     return (quantile * sampleStandardDeviation(numberOfSamples, sum, squareSum)) / Math.sqrt(numberOfSamples);
 }
 
 export function confidenceInterval(values, probability) {
-    var sum = sum(values);
-    var mean = sum / values.length;
-    var delta = confidenceIntervalDelta(probability || 0.95, values.length, sum, squareSum(values));
+    const sum_value = sum(values);
```
rniwa, line 89: Nit: camelCase.

resources/metric.mjs
```diff
@@ -0,0 +1,90 @@
...
+        this.mean = 0.0;
+        this.geomean = 0.0;
+        this.delta = 0.0;
+        this.percentDelta = 0.0;
+
+        this.sum = 0.0;
+        this.min = 0.0;
+        this.max = 0.0;
+
+        /** @type {number[]} */
```
rniwa, line 23: I don't think we should have these comments.

camillobruni: - I will remove them (that's literally just copy-pasted from the grandprix repo)
- Might be something for the future though, the type annotations are quite nice when working with vscode (and we could have typescript verification)

resources/metric.mjs
camillobruni: DONE.

resources/statistics.mjs
camillobruni: Ups, too much python :)

resources/metric.mjs
```diff
@@ -1,7 +1,7 @@
 import * as Statistics from "./statistics.mjs";
 
 /** Number of milliseconds in a single minute. */
```
rniwa, line 3: I don't think we need the comment now that the variable is appropriately named.

resources/benchmark-runner.mjs
```diff
@@ -118,6 +119,11 @@ export class BenchmarkRunner {
         this._suites = suites;
         this._client = client;
         this._page = null;
+        this._metrics = {
+            __proto__: null,
+            Total: new Metric("Total"),
+            Score: new Metric("Score", "rpm"),
```
rniwa, line 125: Why are we using "rpm" here? Didn't we decide to move away from it?

rniwa (approved):

resources/metric.mjs
camillobruni: removed.

resources/benchmark-runner.mjs
camillobruni: Ups, copy-pasta. Changed unit to "score" as well.

camillobruni:
- Use more explicit Metric objects with parent and child-metrics
- Update statistics.mjs to use more ES6 features

PR #65, opened 2023-02-13
Add buttons to download and copy JSON  results

camillobruni:
- Add "Download JSON" button to details view
- Add "Copy JSON" button to details view
- Example JSON result file name: "speedometer_3-2023-02-13T09_18_32.318Z.json"
- Update CSS formatting to make details button all the same width
- Add "title" to results buttons (both summary and details)

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -202,6 +202,13 @@ class MainBenchmarkClient {
         resultsTables[1].innerHTML = "";
         for (; i < formattedValues.length; i++)
             this._addDetailedResultsRow(resultsTables[1], i, formattedValues[i]);
+
+        const jsonData = JSON.stringify(this._measuredValuesList);
+        const blob = new Blob([jsonData], {type: "application/json"});
+        const url = URL.createObjectURL(blob);
+        const link = document.getElementById("download-json");
+        link.href = url;
+        link.setAttribute("download", `speedometer_3-${new Date().toISOString()}.json`);
     }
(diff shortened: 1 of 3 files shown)


resources/main.mjs
```diff
@@ -269,6 +277,23 @@ class MainBenchmarkClient {
...
+    _getFormattedJSONResult()
+    {
+        const indent = "    ";
+        return JSON.stringify(this._measuredValuesList, undefined, indent);
+    }
+
+    copyJsonResults()
+    {
+
+        navigator.clipboard.writeText(this._getFormattedJSONResult());
```
rniwa, line 289: This API doesn't exist in Safari or Firefox.

rniwa (changes requested):
We should probably be using document.execCommand('copy') and event.clipboardData.

resources/main.mjs
camillobruni: It does seem to work both in Firefox 109.0.1 (64-bit) and Safari Version 16.3 on my mac.

camillobruni:
Just look at the [mdn page again](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#browser_compatibility), seems like the only "restriction" is that writeText is only valid during a user interaction.

Might have missed something else?

rniwa (approved):

camillobruni:
Thanks 👍

PR #66, opened 2023-02-13
Adding prefer-template rule to Eslint

flashdesignory:
As discussed, adding rule to prefer template strings.
Details: https://eslint.org/docs/latest/rules/prefer-template

Some additional files got formatted when running our scripts, but those seem straightforward fixes.


resources/statistics.mjs
```diff
@@ -62,23 +62,19 @@ export function supportedConfidenceLevels() {
...
-        throw (
-            "We only support "
-            + supportedConfidenceLevels()
-                .map(function (level) {
-                    return level * 100 + "%";
-                })
-                .join(", ")
-            + " confidence intervals."
-        );
+        throw `We only support ${supportedConfidenceLevels()
```
rniwa, line 65: Can we call this function outside the template literal? This is hard to read.
Maybe something like:
```
const supportedIntervals = supportedConfidenceLevels().map(function (level) { return `${level * 100}%`; });
throw `We only support ${supportedIntervals.join(", ")} confidence intervals.`
```

rniwa (approved):

resources/statistics.mjs
flashdesignory: Agreed - updated the file. Let me know if that works.

rniwa (approved):
r=me

PR #67, opened 2023-02-14
Remove wait from selenium test

flashdesignory:
@bgrins pointed out that we introduced some flakiness in our test by adding a timeout in addition to using the "wait" function from the web driver. 

The approach that I am suggesting here, is to take advantage of the "complete" callback from mocha in the browser, dispatching an even on the window object and listening for that event through "executeAsyncScript" from the web driver, which returns a callback with the stats from Mocha. This gives us a bridge between both scopes and returns the mocha stats results that we need to make the assertions.


rniwa (approved):

bgrins (approved):
Thanks

Issue #51, opened 2023-01-26
Add a svelte todomvc demo


bgrins:
As discussed in #3. Unfortunately there's not an example at https://github.com/tastejs/todomvc/tree/master/examples so we may have to write one from scratch or look around elsewhere

bgrins:
Luckily there's an official example at https://github.com/sveltejs/svelte-todomvc

julienw:
solved by https://github.com/WebKit/Speedometer/pull/135 ?

rniwa:
Indeed.

PR #68, opened 2023-02-16
Add Svelte TodoMVC

bgrins:
This is straightforwardly vendoring in https://github.com/sveltejs/svelte-todomvc, with one dependency update outlined in the SPEEDOMETER_VENDORING.md file. There may be some tweaks needed to bring it in line with the rest of the demos, but opening up a PR for #51 to start that discussion.


resources/todomvc/architecture-examples/svelte/public/todomvc-app-css.css
```diff
@@ -0,0 +1,379 @@
...
+	border: 0;
+	background: none;
+	font-size: 100%;
+	vertical-align: baseline;
+	font-family: inherit;
+	font-weight: inherit;
+	color: inherit;
+	-webkit-appearance: none;
+	appearance: none;
+	-webkit-font-smoothing: antialiased;
```
rniwa, line 19: I don't think these things to be applied only in WebKit/Firefox.

resources/todomvc/architecture-examples/svelte/public/todomvc-app-css.css
```diff
@@ -0,0 +1,379 @@
...
+
+body {
+	font: 14px 'Helvetica Neue', Helvetica, Arial, sans-serif;
+	line-height: 1.4em;
+	background: #f5f5f5;
+	color: #4d4d4d;
+	min-width: 230px;
+	max-width: 550px;
+	margin: 0 auto;
+	-webkit-font-smoothing: antialiased;
```
rniwa, line 31: Ditto.

resources/todomvc/architecture-examples/svelte/public/todomvc-app-css.css
```diff
@@ -0,0 +1,379 @@
...
+	font-family: inherit;
+	font-weight: inherit;
+	line-height: 1.4em;
+	border: 0;
+	color: inherit;
+	padding: 6px;
+	border: 1px solid #999;
+	box-shadow: inset 0 -1px 5px 0 rgba(0, 0, 0, 0.2);
+	box-sizing: border-box;
+	-webkit-font-smoothing: antialiased;
```
rniwa, line 98: Ditto.

resources/todomvc/architecture-examples/svelte/public/todomvc-app-css.css
bgrins: btw the specific css here is coming from npm https://github.com/WebKit/Speedometer/pull/68/files#diff-5df0fe52b53cca989a5f8385d250b1ddbefa104733a1658bf1112320abb8a8d9R33, which appears to be a different version from some of the existing todomvc demos. If we want to fix this maybe we should fork that CSS into this repo and then update all the relevant package.json files / copies of the CSS to point to that. I'm not sure though, curious for thoughts from @housseindjirdeh

.gitignore
```diff
@@ -2,4 +2,5 @@
 *.pyc
 .DS_Store
 .directory
-/node_modules
\ No newline at end of file
+/node_modules
+resources/todomvc/architecture-examples/svelte/node_modules
```
bgrins, line 6: Should we just ignore all node_modules folders instead of one-off. I believe the plan for this is that we check in all package lock files so that we don't need to bloat the repo with the actual modules (relying on build tools to make sure we have a page that can be run from a static server)

bgrins:
After speaking with @flashdesignory - let's pause on reviewing this until we have a chance to research the state of todomvc apps. I'm going to close this for now and we can reopen when that's done.

PR #69, opened 2023-02-16
Use params everywhere

camillobruni:
- Add support for filtering multiple suites
- Support startAutomatically in the default runner
- Fix displaying detailed results with unit=ms
- More error messages on invalid url-params
- Add Suites.enable function instead of two separate implementations per runner

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -6,7 +6,7 @@ import { params } from "./params.mjs";
 
 // FIXME(camillobruni): Add base class
 class MainBenchmarkClient {
-    displayUnit = "runs/min";
+    displayUnit = "score";
     iterationCount = 10;
     stepCount = null;
     suitesCount = null;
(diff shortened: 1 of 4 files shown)


resources/params.mjs
```diff
@@ -15,19 +18,42 @@ class Params {
...
+
+        if (searchParams.has("unit")) {
+            this.unit = searchParams.get("unit");
+            if (!UNITS.includes(this.unit))
+                throw new Error(`Invalid unit=${this.unit}. Valid values are ${UNITS}`);
+            searchParams.delete("unit");
+        }
+
+        if (searchParams.has("suite") || searchParams.has("suites")) {
+            const value = searchParams.get("suites") || searchParams.get("suites");
```
rniwa, line 47: Isn't one of these supposed to be "suite", and not both "suites"?

resources/tests.mjs
```diff
@@ -3,6 +3,21 @@ import { BenchmarkTestStep } from "./benchmark-runner.mjs";
 const numberOfItemsToAdd = 100;
 export const Suites = [];
 
+Suites.enable = function (names) {
+    this.forEach((suite) => {
+        suite.disabled = true;
```
rniwa, line 8: It doesn't seem like this forEach is needed. We can just add else clause in the forEach call below.

resources/tests.mjs
```diff
@@ -443,3 +458,8 @@ Suites.push({
         }),
     ],
 });
+
+Object.freeze(Suites);
```
rniwa, line 462: Let's not do this in this PR?

resources/params.mjs
camillobruni: - I kept "suite" for backwards compatiblity (mayabe that's adding more confusion?)
- I'll add a warning if both "suite" and suites" are specified

rniwa: What I mean is that you have "suites" or "suites". Both of them are suites instead of one being "suite".

camillobruni: Ah sorry, misread. DONE.

resources/tests.mjs
camillobruni: removed.

resources/tests.mjs
camillobruni: DONE

resources/main.mjs
```diff
@@ -19,51 +19,20 @@ class MainBenchmarkClient {
...
-                        }
-                        break;
-                }
+        this.displayUnit = params.unit;
+        this.iterationCount = params.iterationCount;
+        if (params.suites.length > 0) {
+            if (!Suites.enable(params.suites)) {
+                const message = `Suite "${params.suites}" does not exist. No tests to run.`;
+                alert(message);
+                console.error(
```
rniwa, line 29: It seems that this console.error could fit in a single line?

rniwa (approved):

resources/main.mjs
camillobruni: Double checked, `npm run format` somehow want's this on multiple lines :(.

rniwa: Weird.

resources/params.mjs
```diff
@@ -15,19 +18,43 @@ class Params {
...
+        searchParams.delete("startAutomatically");
+        if (searchParams.has("iterationCount")) {
+            this.iterationCount = parseInt(searchParams.get("iterationCount")) || this.iterationCount;
+            if (this.iterationCount <= 1)
+                throw new Error(`Invalid iterationCount param: ${this.iterationCount}`);
+            searchParams.delete("iterationCount");
+        }
+
+        if (searchParams.has("unit")) {
+            this.unit = searchParams.get("unit");
```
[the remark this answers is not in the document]
camillobruni: done.

PR #70, opened 2023-02-16
Use 2x logo and gauge png

camillobruni:
Make sure we load the hi-res background images in all cases:
- Use srcset for logo
- Use image-set for gauge CSS


bgrins (approved):
Nice cleanup

PR #71, opened 2023-02-21
Use more helper methods on PageElement

camillobruni:
Passing in custom event classes to PageElement might become problematic for custom runners.
Let's use custom helper-methods for now to stay runner independent.

- Add mouseUp, mouseDown, mouseMove and scroll helpers
- Use deltaX instead of just delta for wheel events

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -76,20 +76,59 @@ class PageElement {
         this.#node.value = value;
     }
 
-    click() {
-        this.#node.click();
-    }
-
     focus() {
         this.#node.focus();
     }
 
-    dispatchEvent(eventName, options = NATIVE_OPTIONS, eventType = Event) {
+    click() {
+        this.#node.click();
+    }
+
+    mouseMove(clientX = 0, clientY = 0) {
+        this._dispatchMouseEvent("mousemove", clientX, clientY);
+    }
+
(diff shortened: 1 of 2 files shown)


resources/benchmark-runner.mjs
```diff
@@ -76,20 +76,59 @@ class PageElement {
...
+        this.#node.focus();
+    }
+
     click() {
         this.#node.click();
     }
 
-    focus() {
-        this.#node.focus();
+    mouseMove(clientX = 0, clientY = 0) {
```
squarewave, line 87: Nit: why the default values?

squarewave (approved):
This looks good to me. Thanks!

resources/tests.mjs
```diff
@@ -395,51 +370,26 @@ Suites.push({
...
-            event = {
-                clientX: 200,
-                clientY: 200,
-                deltaMode: 0,
-                delta: -10,
-                deltaY: -10,
-                bubbles: true,
-                cancelable: true,
-            };
+                cursor.wheel(10, 10, 650, 200);
```
rniwa, line 390: Did you mean to call scroll? I don't think there is wheel method defined anywhere.

rniwa:
I'm adding similar helper methods in https://github.com/WebKit/Speedometer/pull/156.

PR #76, opened 2023-02-21
Remove _writeMark test

camillobruni:
The method has been removed since we're now using performance.now directly.


rniwa (approved):

camillobruni:
Thanks! I was a bit too quick with landing the previous PR ...

PR #79, opened 2023-02-22
WIP - use performance.measure api instead of manually tracking with performance.now()

bgrins:
This would allow profiler tools to show user timing ranges and makes it easier to focus in on different portions of tests.

@camillobruni what do you think about this? My understanding of this API from https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure is that the `duration` object functionally the same as the existing subtractions using performance.now()

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -230,22 +230,18 @@ export class BenchmarkRunner {
     // This function ought be as simple as possible. Don't even use Promise.
     _runTest(suite, test, page, callback) {
         performance.mark(`${suite.name}.${test.name}-start`);
-        const syncStartTime = performance.now();
         test.run(page);
-        const syncEndTime = performance.now();
         performance.mark(`${suite.name}.${test.name}-sync-end`);
+        const syncTime = performance.measure(`${suite.name}.${test.name}-sync`, `${suite.name}.${test.name}-start`).duration;
 
-        const syncTime = syncEndTime - syncStartTime;
-
-        const asyncStartTime = performance.now();
+        performance.mark(`${suite.name}.${test.name}-async-start`);
         setTimeout(() => {
             // Some browsers don't immediately update the layout for paint.
             // Force the layout here to ensure we're measuring the layout time.
             const height = this._frame.contentDocument.body.getBoundingClientRect().height;
-            const asyncEndTime = performance.now();
-            const asyncTime = asyncEndTime - asyncStartTime;
+            const asyncTime = performance.measure(`${suite.name}.${test.name}-async`, `${suite.name}.${test.name}-async-start`).duration;
+
             this._frame.contentWindow._unusedHeightValue = height; // Prevent dead code elimination.
-            performance.mark(`${suite.name}.${test.name}-async-end`);
             window.requestAnimationFrame(() => {
                 callback(syncTime, asyncTime, height);
             });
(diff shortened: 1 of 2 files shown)


bgrins:
Note: there's a test failure due to calling performance.mark one less time. those could be fixed by either restoring the `${suite.name}.${test.name}-sync-end` mark which I think is basically unnecessary with this change (but can do that if people are relying on it), or updating the assertion to match.

rniwa (commented):
I'd rather not change the way we measure the performance since there could be behavioral difference between calling now() twice vs calling mark() & measure() with respect to jittering added for Spectre/Meltdown mitigation. If we're making this change, then we need to vet the change across browsers on various device types.

LeszekSwirski:
There would be value to us on the chrome side in switching to the performance measure API, since this shows up as a named range in performance traces. I would consider it worth putting in the effort to validate, cc @camillobruni

bgrins:
Agree @LeszekSwirski - this allows the Firefox profiler tooling to show named ranges which makes profiling each step quite a bit more convenient.

julienw:
I don't know for other browsers, but Firefox at least doesn't seem to jitter the timestamp used for `measure` (see [the Firefox code](https://searchfox.org/mozilla-central/rev/a3a9112d4d73d1323eabbc7faa9937cd9aae6465/dom/performance/Performance.cpp#514)). However, using a previous performance timing `mark` name [loops over all timing entries](https://searchfox.org/mozilla-central/rev/a3a9112d4d73d1323eabbc7faa9937cd9aae6465/dom/performance/Performance.cpp#192-196), and therefore this can have a non-constant impact (not on this specific call as `Now` is called before that but on the async measurement after it). So I wouldn't recommend using this change directly.

However we could keep the current code and _in addition_ add `performance.measure` calls after the end of benchmark, using the already present 2 * 2 mark entries:

```js
performance.measure(
  `${suite.name}.${test.name}-sync`,
  `${suite.name}.${test.name}-start`,
  `${suite.name}.${test.name}-sync-end`
);
performance.measure(
  `${suite.name}.${test.name}-async`,
  `${suite.name}.${test.name}-async-start`,
  `${suite.name}.${test.name}-async-end`
);
```


<br>

It's worth noting that `performance.measure` also supports a newer syntax, that allows passing timestamps directly. This could look like this (to be added after the end of the benchmark):
```js
performance.measure(`${suite.name}.${test.name}-sync`, { start: syncStartTime, end: syncEndTime });
performance.measure(`${suite.name}.${test.name}-async`, { start: asyncStartTime, end: asyncEndTime });
```
and the `mark` entries would be removed.
However this is sitll pretty new (for example Firefox supports this just since version 103, so it's not supported by its ESR version) and therefore it's probably not well-suited for this benchmark, so I wouldn't recommend this for now. This could maybe be added in a few months if we want to.

bgrins:
Thanks @julienw! What's the failure case for browsers that don't support the direct timestamp assignment? If it's just that the measure range won't show up I would say that's generally acceptable (status quo with the current runner). That seems marginally better than keeping the custom timings + marks + measure at the end, but I'd also be happy to change it to that, as it gives the same benefit (user timing ranges in profiler tools).

julienw:
This is what we get if we use it unconditionally: `Uncaught DOMException: An invalid or illegal string was specified`. (I tried in Firefox ESR)
We could try/catch it. Or we could feature-detect it at load time.
But given that there's a more compatible solution that gives the same result (albeit with some extra "marks") I think that using it makes more sense.

bgrins:
>  But given that there's a more compatible solution that gives the same result (albeit with some extra "marks") I think that using it makes more sense.

I agree. It seems very unobtrusive, and even though it's a bit more code/bookeeping it will get the same outcome.

rniwa (approved):

PR #80, opened 2023-02-23
remove todoMVC react / bower workload & labs folder

flashdesignory:
Removing the react todoMVC version that is using bower:
https://github.com/WebKit/Speedometer/tree/main/resources/todomvc/labs/architecture-examples/react

Additionally, I don't see a need to have an extra "labs" folder at the moment, but I can put that back if we want to have the folder structure for future use.


bgrins (approved):
As far as I know this is unused, and so we should just remove the code. @rniwa what do you think?

rniwa:
Sure, we can delete unused code.

PR #82, opened 2023-03-01
Todomvc es5 updates

flashdesignory:
- renamed todoMVC workload
- added proper package.json
- updated readme file
- renamed "js" to "src"
- add S3 formatting
- add build script (this moves all necessary files to a dist folder for the tests)
- add dev server (for local preview)
- pin node version (same as s3)
- fixed toggle all 
- removed node_modules folder from github (since we have a package-lock files)

**No refactoring of logic or styling updates have been implemented.**


resources/todomvc/vanilla-examples/javascript-es5/.eslintrc
```diff
@@ -0,0 +1,98 @@
+/*
```
bgrins, line 1: Could we make this inherit eslint and prettier settings from the project root by removing it from the relevant ignore files? That would make each individual patch smaller and central changes easier.

flashdesignory: ah I see what you mean. 
I like that each workload is independent at the moment. 

The potential disadvantage of having S3 format all workloads is that it might surface issues on workloads that I am not currently working on.

resources/tests.mjs
```diff
@@ -18,8 +18,8 @@ Suites.enable = function (names) {
 };
 
 Suites.push({
-    name: "VanillaJS-TodoMVC",
-    url: "todomvc/vanilla-examples/vanillajs/index.html",
+    name: "TodoMVC-JavaScript-ES5",
```
bgrins, line 21: I like the suggestion since it will allow us to namespace ScenarioName-WorkloadName when looking at results and the codebase. Though this will likely require some kind of migration for any tooling that's tracking subtests so let's make sure everyone is OK with that (cc @rniwa, and I'll confirm with our team)

resources/todomvc/vanilla-examples/javascript-es5/.eslintrc
bgrins: I would opt in folder by folder - see for example https://github.com/WebKit/Speedometer/pull/81/files#diff-b640b344ee7f3f03d2a443795a5d0708ef50e2e6e34214109ab2aad13ad6ba98L3

resources/tests.mjs
bgrins: Confirmed this would be fine for us to change

resources/todomvc/vanilla-examples/javascript-es5/index.html
```diff
@@ -1,20 +1,23 @@
-<!doctype html>
-<html lang="en" data-framework="javascript">
+<!DOCTYPE html>
+<html lang="en" data-framework="javascript" data-version="es5" data-features="">
```
bgrins, line 2: What are data-features and data-version for?

resources/todomvc/vanilla-examples/javascript-es5/src/helpers.js
```diff
@@ -0,0 +1,52 @@
+/* global NodeList */
```
bgrins, line 1: I wonder why this wasn't detected as a file move... I assume this is the same content as before but let me know if it isn't

.gitignore
```diff
@@ -2,4 +2,4 @@
 *.pyc
 .DS_Store
 .directory
-/node_modules
\ No newline at end of file
+*node_modules
```
bgrins, line 5: Aren't there other node_modules folders still in the tree? I assume we only want to remove this individual test's (as long as the root one) for now

bgrins (changes requested):
This looks good to me. My main request is to remove the linting/formatting machinery (in package.json and associated files) and rely on the config at the project root by modifying prettier ignore

flashdesignory:
> This looks good to me. My main request is to remove the linting/formatting machinery (in package.json and associated files) and rely on the config at the project root by modifying prettier ignore

@bgrins - sounds good - i'll make the change 😄

resources/todomvc/vanilla-examples/javascript-es5/index.html
flashdesignory: I have to admit, that I saw "data-framework" in the html files and thought that it's an interesting feature to have, although it doesn't look like we're doing anything with it. 
Maybe that could be useful in the future if we want to query during the tests what framework, version and features are currently tested? 

I can totally remove all of those data attributes though

resources/todomvc/vanilla-examples/javascript-es5/src/helpers.js
flashdesignory: 🤷

.gitignore
flashdesignory: good point - will adjust

resources/todomvc/vanilla-examples/javascript-es5/.eslintrc
flashdesignory: done 👍

.gitignore
flashdesignory: done 👍

bgrins (approved):
LGTM. Given the suite name change I'd like @rniwa to sign off before landing though.

resources/tests.mjs
rniwa: This seems fine to me.

rniwa:
Can we have before/after scores across browsers?

flashdesignory:
> Can we have before/after scores across browsers?

https://gist.github.com/flashdesignory/42ac2fbabb18b4c527672509f157442a

rniwa:
Thanks. With those numbers, I think we're good to go!

rniwa (approved):

PR #86, opened 2023-03-07
Fix score view

camillobruni:
- Use params directly in MainBenchmarkClient
- Remove params.unit
- Update summary only with score
- Update details view with ms value
- Fix gauge background on chrome by also adding `-webkit-image-set(..)`

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -6,8 +6,6 @@ import { params } from "./params.mjs";
 
 // FIXME(camillobruni): Add base class
 class MainBenchmarkClient {
-    displayUnit = "score";
-    iterationCount = 10;
     stepCount = null;
     suitesCount = null;
     _measuredValuesList = [];
@@ -20,8 +18,6 @@ class MainBenchmarkClient {
     }
 
     startBenchmark() {
-        this.displayUnit = params.unit;
-        this.iterationCount = params.iterationCount;
         if (params.suites.length > 0) {
             if (!Suites.enable(params.suites)) {
                 const message = `Suite "${params.suites}" does not exist. No tests to run.`;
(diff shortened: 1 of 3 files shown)


resources/main.mjs
```diff
@@ -78,18 +74,18 @@ class MainBenchmarkClient {
...
-        document.getElementById("result-number").textContent = results.formattedMean;
-        if (results.formattedDelta)
-            document.getElementById("confidence-number").textContent = `\u00b1 ${results.formattedDelta}`;
+        const scoreResults = this._computeResults(this._measuredValuesList, "score");
+        this._updateGaugeNeedle(scoreResults.mean);
+        document.getElementById("result-number").textContent = scoreResults.formattedMean;
+        if (scoreResults.formattedDelta)
+            document.getElementById("confidence-number").textContent = `\u00b1 ${scoreResults.formattedDelta}`;
 
+        const results = this._computeResults(this._measuredValuesList, params.unit);
```
bgrins, line 84: The only issue I have with this commit is that I don't know if anyone looking at the detailed view would want to see it displayed as "score" - and this changes the default to do that. My hypothesis is that the details view is only useful for browser engineers - who would all like to see the results in ms. So I wonder if in practice what we want is to have a param like  `showDetails` instead of `unit` which only controls which panel is visible by default and then we always have Summary = score and Details = ms.

I could be convinced otherwise if you feel strongly about this though, and will also ask a few of our engineers for thoughts.

camillobruni: humm you're right, there is little added value from having params.unit.
- Let me remove params.unit
- Summary view always shows score
- Detailed view always shows ms

bgrins (approved):
Thanks. Maybe we separately introduce a `?dev` or similar param that can control going straight to details (and layering in checkbox selection feature from InteractiveRunner)

camillobruni:
Adding `&dev` sounds good.

bgrins:
I filed https://github.com/WebKit/Speedometer/issues/88 so we don't lose track of that

PR #89, opened 2023-03-08
Initial dev mode param implementation

### LINKED ISSUE #88: Add a `dev` param
This would initially control the default view (to go straight to details), and could also be used to add in a test selection list (and thus allow us to remove InteractiveRunner)

bgrins:
Fixes #88. For now this just switches to the details view. I'd like to expand it over time so it has enough features to remove InteractiveRunner.html.

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -87,11 +87,10 @@ class MainBenchmarkClient {
         this._populateDetailedResults(results.formattedValues);
         document.getElementById("results-with-statistics").textContent = results.formattedMeanAndDelta;
 
-        if (this.developerMode) {
+        if (this.developerMode)
             this.showResultsDetails();
-        } else {
+        else
             this.showResultsSummary();
-        }
     }
 
     _computeResults(measuredValuesList, displayUnit) {
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -6,7 +6,7 @@ import { params } from "./params.mjs";
 
 // FIXME(camillobruni): Add base class
 class MainBenchmarkClient {
-    dev = false;
+    developerMode = false;
     stepCount = null;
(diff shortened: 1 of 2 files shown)


flashdesignory (approved):

resources/main.mjs
```diff
@@ -6,6 +6,7 @@ import { params } from "./params.mjs";
 
 // FIXME(camillobruni): Add base class
 class MainBenchmarkClient {
+    dev = false;
```
rniwa, line 9: "dev" is a really generic name. Can we instead call it `isInDeveloperMode`?

bgrins: Maybe we meet in the middle - `devMode` or `developerMode`? More descriptive than `dev` but also still <= 2 words like other params.

bgrins: Renamed to developerMode in the latest push - lmk if that works for you

resources/main.mjs
```diff
@@ -85,7 +87,11 @@ class MainBenchmarkClient {
         this._populateDetailedResults(results.formattedValues);
         document.getElementById("results-with-statistics").textContent = results.formattedMeanAndDelta;
 
-        this.showResultsSummary();
+        if (this.dev) {
```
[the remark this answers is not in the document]
bgrins: Ah I had been running `npm run format` and it didn't make this change. If I run `npm run pretty:fix` it does but also a bunch of other changes which makes me think maybe the format command needs to be running lint first and then prettier or something? cc @flashdesignory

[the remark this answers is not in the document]
bgrins: I'll fix this manually in the commit and we can figure out the tooling separately

rniwa (approved):

PR #90, opened 2023-03-09
Improve runner

camillobruni:
- Pre-allocate all mark-label strings to avoid potential gc's during sensitive measurements
- Move code to new separate `_recordTestResults` method in preparation for pr #28

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -211,33 +211,28 @@ export class BenchmarkRunner {
                 await this._client.willRunTest(suite, test);
 
             setTimeout(() => {
-                this._runTest(suite, test, this._page, async (syncTime, asyncTime) => {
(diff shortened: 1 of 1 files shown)


resources/benchmark-runner.mjs
```diff
@@ -211,33 +211,28 @@ export class BenchmarkRunner {
...
-                });
+                this._runTest(suite, test, this._page, resolve);
             }, 0);
         });
     }
 
     // This function ought be as simple as possible. Don't even use Promise.
-    _runTest(suite, test, page, callback) {
-        performance.mark(`${suite.name}.${test.name}-start`);
+    _runTest(suite, test, page, test_done_callback) {
```
rniwa, line 220: Please use camelCase for testDoneCallback.

camillobruni: argh, sorry. Too much python 😛

PR #98, opened 2023-03-13
Todomvc react update

flashdesignory:
- Update version to 17.0.2 
- Add updated readme
- Enforce style guide.
- Rewrite & cleanup app
- Use functional components & hooks
- Break up components into separate building blocks - MV(*) 
- Use Webpack 5 + Babel 
- Ensure Toggle for all todos is displaying correctly

numbers:
https://gist.github.com/flashdesignory/dfe0158371f7c6052dbfe4ba8767a96b

cc: @kara


resources/todomvc/architecture-examples/react/src/todo/components/Input.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+    const handleBlur = (e) => {
+        if (onBlur) {
+            onBlur();
+        }
+    };
+
+    return (
+        <form onSubmit={handleSubmit}>
+            <input className="new-todo" id="todo-input" type="text" data-testid="text-input" autoFocus placeholder={placeholder} defaultValue={defaultValue} onBlur={handleBlur} />
```
julienw: In react I believe it's usually advised to implement as a "controlled input", where the value is kept in the local state. Here instead you implemented an uncontrolled input. I believe this works and I'll leave it up to you to decide what style you prefer, but I wanted to point it out. For example a controlled component might make it easier to access the value in `handleSubmit`.

Controlled components: https://reactjs.org/docs/forms.html#controlled-components
Uncontrolled components: https://reactjs.org/docs/uncontrolled-components.html

This is the old doc, it looks like that the new documentation website is less assertive about which style to use.

julienw (commented):
I looked at this commit by commit, and while it gave a good insight in your thought process, it didn't make it very easy to review. I believe that otherwise all my comments to the preact PR also apply here, especially related to the use of `useCallback` and `useMemo`, so I won't repeat them. In the commits I see that you introduced them after some feedback, so I'd love to know more :-)
In case you'd like to keep them, it would be good to document the rationale using comments in the code.

Compared to the preact PR, this time the numbers are much lower, have you profiled to understand what could have changed?

flashdesignory:
@julienw - I did and I am still comparing both the react and preact versions to determine what affects the scores. 
It's hard to compare both previous builds, since they are very different. 

Overall, the biggest impact in the new builds was to wrap the <Item/> with a memo call, to ensure items that didn't change won't re-render. You can observe that behavior in both builds if you run them locally with the dev tools. 

The react version went through some internal feedback rounds initially and therefore you can see some changes in the commits. 

Regarding the inputs, I can see that there might be some discussions and I can switch the input to a controlled component if there are strong feelings. I honestly didn't see a need to keep additional state, if we can just read the value on the event handler.

resources/todomvc/architecture-examples/react/src/todo/components/Input.jsx
flashdesignory: yes - I used a form before 😄 
It does solve some problems, but even with a form I honestly wouldn't keep local state.
I could still read the value from the event (e.target.elements["todo-input"].value).

rniwa (approved):

resources/tests.mjs
```diff
@@ -104,8 +104,8 @@ Suites.push({
 });
 
 Suites.push({
-    name: "React-TodoMVC",
-    url: "todomvc/architecture-examples/react/index.html",
+    name: "TodoMVC-React",
+    url: "todomvc/architecture-examples/react/dist/index.html",
     async prepare(page) {
         const element = await page.waitForElement(".new-todo");
         element.focus();
```
julienw, line 111: This is a comment for the line `dispatchEvent("input")` below, and this can be handled in a separate pull request.

Current versions of react (since ~v16) have some code that tracks the `value`. As a result the `input` event doesn't do anything now:
* when `setValue` is called, react saves the `value`
* when the event is dispatched, react compares the current `value` with the saved one, they are the same value => as a result it swallows the `input` event. See https://github.com/facebook/react/issues/10135 for more discussion about this.

The solution that the `testing-library` library uses is this `setNativeValue` function that we might want to steal to replace our current implementation of `setValue`:
https://github.com/testing-library/dom-testing-library/blob/fb069c93983bc0300a6e1c91bdec5bf9443b5286/src/events.js#L106-L123

This has no impact to _this_ version of todomvc because this version doesn't use the `input` or `change` event (as opposed to the previous one), that's why it can be handled separely.

julienw:
Today I tried to _only_ update react to v17 on the old app. You can see the result in this branch => https://github.com/WebKit/Speedometer/compare/main...julienw:Speedometer:only-update-react?expand=1
(I didn't check in the changes to node_modules so make sure to run `npm i` if you want to try it -- also my editor automatically formatted the benchmark-runner so don't pay attention to that)

In short: in the context of this app, the React v17 update seems to be just a little bit faster on chrome, and similarly fast in Firefox. (I can't check Safari).

flashdesignory:
@julienw - that's a great comparison. 
How do you measure the speed and are the comparisons against the "old" react workload or the proposed react workload?

julienw:
> Ah ok, and with those tests, which version was the fastest?

Your new version, by a large margin!

I also profiled (both in chrome and Firefox) to see if anything was outside of the performance marks (that could explain the timings change); but no, everything is still inside the performance marks. So the app is probably just faster due to how it's built. I believe you're right that by memozing the `Item` component we're cutting out a lot of the work. Maybe it would be even more visible if we'd use more than 100 items -- maybe we should to better show how the browsers deal with bigger arrays (?) (but that would need to be discussed of course).

resources/todomvc/architecture-examples/react/src/todo/components/input.jsx
```diff
@@ -26,7 +26,7 @@ export function Input({ onSubmit, placeholder, label, defaultValue, onBlur }) {
 
     const handleKeyDown = useCallback(
         (e) => {
-            if (e.keyCode === 13) {
+            if (e.key.match(/Enter/i)) {
```
julienw, line 29: I noticed that the value for `key` we get from the runner is `ENTER`, but it should really be `Enter` ([see MDN](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values)). IMO the runner should be fixed at https://github.com/WebKit/Speedometer/blob/6523f664973df6bffa2dd24836f8b479f1d0f241/resources/benchmark-runner.mjs#L110

resources/todomvc/architecture-examples/react/webpack.common.js
```diff
@@ -0,0 +1,36 @@
...
+            {
+                test: /\.(js|jsx)$/,
+                exclude: /nodeModules/,
+                use: {
+                    loader: "babel-loader",
+                    options: {
+                        presets: [
+                            ["@babel/preset-env", { targets: "defaults" }],
+                            ["@babel/preset-react", { runtime: "automatic" }],
+                        ],
```
julienw, line 26: I see it's already in `babel.config.js`, is it necessary to have both?
Also it's not clear to me what environement babel targets here... It looks like it targets some recent version of JS but it would be good to explicit this, especially that [babel seems to say otherwise](https://babeljs.io/docs/options#no-targets)...

flashdesignory: thanks - I removed the babel.config file

resources/todomvc/architecture-examples/react/src/todo/components/input.jsx
flashdesignory: merged main with the fix

flashdesignory:
@julienw  - please let me know where we are with this pr 😄

bgrins (approved):

resources/todomvc/architecture-examples/react/package.json
```diff
@@ -1,20 +1,40 @@
...
+        "mini-css-extract-plugin": "^2.7.2",
+        "style-loader": "^3.3.1",
+        "webpack": "^5.75.0",
+        "webpack-cli": "^5.0.1",
+        "webpack-dev-server": "^4.11.1",
+        "webpack-merge": "^5.8.0"
+    },
+    "dependencies": {
+        "classnames": "^2.2.5",
+        "react": "^17.0.2",
```
gsathya, line 35: Apologies for the drive-by question: but why pick v17 and not v18 (current stable)?

julienw: I believe there are 2 reasons:
* v17 is still the most used version in the web currently.
* v18 is much more asynchronous, so it's difficult (if possible) to determine when it finished rendering. We'll have to eventually solve this problem though...

tetsuharuohzeki: > * so it's difficult (if possible) to determine when it finished rendering. We'll have to eventually solve this problem though...

At API Level, React ~v17 also has similar problem that is hard to know when the "rendering" finished and the 3rd argument of [`render()`](https://react.dev/reference/react-dom/render) has been used for that. 

After v18, these guide might be helpful

- https://github.com/reactwg/react-18/discussions/5
- https://react.dev/blog/2022/03/08/react-18-upgrade-guide

housseindjirdeh: FYI: Some more context on why v17 was chosen is in this [document](https://docs.google.com/document/d/1MXxk9DryJmvAmPMt2uMneaS3J5Y-dMFGt4EGvMzhhJE/edit?disco=AAAAsVSN7aM) that Thorsten put together.

josephsavona: +1 to what @gsathya mentioned. NPM data suggests that 18 is already the most popular version, and in-the-wild numbers will naturally trail installs. It's really important that we're all optimizing for the right version and I'm very concerned about the choice of 17 here. There is a very big difference between 17 and 18, and optimizations that benefit one may not help the other.

If the question is about how to measure accurately in 18, we can help with that :-)

julienw: Thanks for the useful feedback, folks. I agree that v18 will likely have different performance characterstics compared to v17. It would make sense to include a version of the benchmark with v18.
But also please note that we also intend to update the benchmark more often than what we did so far.

PR #99, opened 2023-03-13
Todomvc react reducer update

flashdesignory:
- Update React-Redux to 7.2.6
- Update React / ReactDom to 16.14.0
- Use Webpack 5 + Babel 
- Add updated readme
- Enforce style guide.
- Keep React Class Components where applicable

numbers:
https://gist.github.com/flashdesignory/31770eb9ad1b55bde168cc59e57eea2e


bgrins:
Would it be too much of a pain to split this two separate PRs - one where you update tooling / style guide / etc which creates a large but mostly mechanical diff, and then one where you bump the library versions including any necessary API changes and design changes.

bgrins:
Same question for other PRs that include functional changes as well - which I believe is the case for 101, 100, and 98. Just scanning this one I do think it's possible to review as-is if it's going to be painful to split up (so just say so if that's the case), but it'd probably be easier if we could focus just on the functional changes.

flashdesignory:
@bgrins - I see what you mean, we have a lot of files that changed. 
It would take a while to take all prs apart and rebuild in phases in order to create smaller prs for review. 

If possible, I'd like to avoid a rebuild of the changes and rather keep that in mind for the remaining versions, unless it's really hard to review correctly otherwise. 

Some ways that might make it easier: 
- take a look at the standalone branch directly:
https://github.com/flashdesignory/Speedometer/tree/todomvc-react-redux/resources/todomvc/architecture-examples/react-redux

- global runner changes for all prs:
renamed the test
always pointing to a dist folder
update gitignore, eslintignore, prettierignore

- global build changes for all prs:
using webpack with similar setup to have consistency in build steps
force style-guide
update readme

The remainder is functional changes when updating the version of the libraries and to reflect modern approaches that are common for the libraries used.

bgrins:
Alright, let's see what @rniwa thinks - and @julienw is going to take a look at this set of PRs for us so his opinion once he can take a look at them will matter more than mine :).

I think Backbone (#101) will not need reorganization regardless of the outcome on the others, since there are very limited functional changes there.

rniwa (approved):

resources/todomvc/architecture-examples/react-redux/src/components/text-input.jsx
```diff
@@ -0,0 +1,55 @@
...
+import classnames from "classnames";
+import PropTypes from "prop-types";
+
+export default class TextInput extends Component {
+    static propTypes = {
+        onSave: PropTypes.func.isRequired,
+        text: PropTypes.string,
+        placeholder: PropTypes.string,
+        editing: PropTypes.bool,
+        newTodo: PropTypes.bool,
```
julienw, line 11: It's not obvious what  `newTodo` does, I think a comment would be useful for this property.

resources/todomvc/architecture-examples/react-redux/src/components/text-input.jsx
```diff
@@ -0,0 +1,55 @@
...
+            <input
+                className={classnames({
+                    edit: this.props.editing,
+                    "new-todo": this.props.newTodo,
+                })}
+                type="text"
+                data-testid="text-input"
+                placeholder={this.props.placeholder}
+                autoFocus
+                value={this.state.text}
```
julienw, line 48: Interesting that we have a controlled component in this version :-) I'm not against it, it's interesting that we have both approaches in the various benchmarks.

resources/todomvc/architecture-examples/react-redux/webpack.common.js
```diff
@@ -0,0 +1,36 @@
...
+                test: /\.(js|jsx)$/,
+                exclude: /nodeModules/,
+                use: {
+                    loader: "babel-loader",
+                    options: {
+                        presets: [
+                            ["@babel/preset-env", { targets: "defaults" }],
+                            ["@babel/preset-react", { runtime: "automatic" }],
+                        ],
+                    },
```
julienw, line 27: Do we need these options if we have `babel.config.js`? It looks confusing to have both.

resources/todomvc/architecture-examples/react-redux/src/components/main.jsx
```diff
@@ -0,0 +1,50 @@
...
+        todos: PropTypes.array.isRequired,
+        actions: PropTypes.object.isRequired,
+        location: PropTypes.object.isRequired,
+    };
+
+    render() {
+        const { todos, actions, location } = this.props;
+
+        const filteredTodos = todos.filter(TODO_FILTERS[location.pathname]);
+        const completedCount = todos.reduce((count, todo) => (todo.completed ? count + 1 : count), 0);
```
julienw, line 25: So we're not memoizing here. Would it make sense to memoize using selectors, so that this is more redux-y?

We'd need a component that would listen for location changes and update the redux state with this info, so that the selector could pick it up.

Then with these selectors we could connect more components. For example `Footer` could be connected and get `completedCount` from the selector (as well as the `clearCompleted` action).

julienw (commented):
Thanks for this patch, this is a lot of work! Also everything seems to work properly!

That said I'm surprised that there's only one connected component. I would have expected that more components would be connected.

For example, I would have expected that `Header` and `Main` would be connected by themselves instead of getting the actions (for both components) and `todos` (for Main) from the `App` component.

Similarly I would have accepted some selectors to compute the data consumed by the components. For example `Footer` as I commented.

I'm not so sure for `Item`, I guess both connected and not connected work for me. If it's connected I would just get the actions through the connection, and still pass the todo through props.

Tell me what you think :-)

resources/todomvc/architecture-examples/react-redux/src/components/main.jsx
julienw: BTW, by memoizing using selectors, I mean using the `reselect` library, like most apps do. (@redux/toolkit uses `reselect` under the hood for example)

resources/todomvc/architecture-examples/react-redux/src/components/text-input.jsx
flashdesignory: added a note 😄

resources/todomvc/architecture-examples/react-redux/webpack.common.js
flashdesignory: added a note in the babel.config.js file, which is only used for the Jest tests. 
I kept those tests since they already existed.

resources/todomvc/architecture-examples/react-redux/src/components/main.jsx
flashdesignory: I tried to keep the previous implementation intact and therefore tried to avoid making changes the way redux was implemented. We could add another redux version of the todo app that uses a more modern approach and with hooks / functional components maybe?

julienw: That said, I understand that the existing version in the tree doesn't do that either. So I'm OK if you want to discuss this topic first in the group, and do it in a follow-up then.

flashdesignory: @julienw - let's make things better! How about this approach:
* I'll update this version and use connected components closer where the todos / dispatch is used
* I'll add some selectors (and look into reselect) 
We just want to ensure that the performance doesn't degrade in any way with these updates.

Additionally, we could discuss adding a "modern" redux version that uses the currently recommended approach with @reduxjs/toolkit / functional components / hooks.

My personal goal for this version was to use a more "classic" version of redux in combination with React 16 and using class components / hocs, ect...

flashdesignory:
@julienw - I pushed an update - let me know how that feels. 
updated scores: https://gist.github.com/flashdesignory/97e8fc22fd451f7bc7f302fec3a902ef

resources/todomvc/architecture-examples/react-redux/src/components/main.jsx
julienw: I like this plan! Maybe just React 17 instead of 16? There's not much difference between 16 and 17 though.
I think we definitely need at least one class component-based implementation, so this is totally fine for me.
About the "modern" redux version my understanding is that we'll need to trade one benchmark for one, because we don't want to add more. But if we have good arguments in favor of this new version, maybe we'll get a pass ;-)

Call me radical :-) but TBH I don't understand this concern of "ensuring that the performance doesn't degrade", as long as the benchmark is close to what we see in the wild. If the performance degrades then this would mean that the real-life websites don't work performantly, and this needs to be fixed in these browsers.

julienw:
> @julienw - I pushed an update - let me know how that feels. updated scores: https://gist.github.com/flashdesignory/97e8fc22fd451f7bc7f302fec3a902ef

Wow the numbers are so much better!

resources/todomvc/architecture-examples/react-redux/src/containers/main-container.jsx
```diff
@@ -0,0 +1,23 @@
...
+    const visibleTodos = getVisibleTodos(todos, location.pathname);
+    const completedCount = getCompletedTodos(todos).length;
+    const activeCount = todos.length - completedCount;
+
+    return { todos, completedCount, activeCount, visibleTodos };
+};
+
+const mapDispatchToProps = (dispatch) => ({
+    actions: bindActionCreators(TodoActions, dispatch),
+});
```
julienw, line 21: optional nit: react-redux knows how to bind automatically and wrap `dispatch` (like you did for the header container above). I believe this would work (haven't checked):
```suggestion
const mapDispatchToProps = actions;
```

resources/todomvc/architecture-examples/react-redux/src/components/item.jsx
```diff
@@ -0,0 +1,62 @@
...
+import classnames from "classnames";
+import PropTypes from "prop-types";
+import TextInput from "./text-input";
+
+export default class Item extends PureComponent {
+    static propTypes = {
+        todo: PropTypes.object.isRequired,
+        editTodo: PropTypes.func.isRequired,
+        deleteTodo: PropTypes.func.isRequired,
+        toggleTodo: PropTypes.func.isRequired,
```
julienw, line 11: I considered about connecting `Item`, but I'm not so sure... We can look at it in a follow-up possibly.

resources/todomvc/architecture-examples/react-redux/webpack.common.js
flashdesignory: fixed the babel.config file

resources/todomvc/architecture-examples/react-redux/src/test/setup.js
```diff
@@ -0,0 +1,9 @@
+import "@testing-library/jest-dom/extend-expect";
+
+import crypto from "crypto";
+
+Object.defineProperty(globalThis, "crypto", {
+    value: {
+        randomUUID: () => crypto.randomUUID(),
+    },
+});
```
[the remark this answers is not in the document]
flashdesignory: this was just for the Jest test environment

resources/todomvc/architecture-examples/react-redux/src/containers/main-container.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default withRouter(connect(mapStateToProps, mapDispatchToProps)(Main));
```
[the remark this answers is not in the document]
flashdesignory: removed in Main

resources/todomvc/architecture-examples/react-redux/src/selectors/filters.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+}
+
+export function getVisibleTodos(todos, route) {
+    return getFilteredTodos(todos, route);
+}
+
+export function getCompletedTodos(todos) {
+    return getFilteredTodos(todos, SHOW_COMPLETED);
+}
```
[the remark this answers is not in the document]
flashdesignory: awesome - let's save it for later

resources/todomvc/architecture-examples/react-redux/src/containers/header-container.jsx
```diff
@@ -0,0 +1,5 @@
+import { connect } from "react-redux";
+import Header from "../components/header";
+import { addTodo } from "../actions";
+
+export default connect(null, { addTodo })(Header);
```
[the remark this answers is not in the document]
flashdesignory: nope - I could have done that too - just preference.

resources/todomvc/architecture-examples/react-redux/src/components/main.jsx
bgrins: >  Additionally, the todomvc-react version is using 17.0.2. It's nice to have two different versions as workloads that represent what's being used.

I agree with this reasoning - if 16 is more commonly used with the patterns used in this workload makes sense to stick with it.

PR #100, opened 2023-03-13
Todomvc preact update

flashdesignory:
- Update version to 10.11.3
- Add updated readme
- Enforce style guide.
- Remove one of the bundlers (we don’t need both)
- Rewrite & cleanup app
- Use functional components & hooks

numbers:
https://gist.github.com/flashdesignory/b5b326a74189ff1109545bc1d8616762

cc: @kara


resources/todomvc/architecture-examples/preact/src/todo/components/item.jsx
```diff
@@ -0,0 +1,51 @@
...
+import { Input } from "./input";
+
+export const Item = memo(function Item({ todo, dispatch }) {
+    const [isWritable, setIsWritable] = useState(false);
+    const { title, completed, id } = todo;
+
+    const toggleItem = useCallback(() => dispatch({ type: "TOGGLE_ITEM", payload: { id } }), [dispatch]);
+    const removeItem = useCallback(() => dispatch({ type: "REMOVE_ITEM", payload: { id } }), [dispatch]);
+    const updateItem = useCallback((id, title) => dispatch({ type: "UPDATE_ITEM", payload: { id, title } }), [dispatch]);
+
```
julienw, line 13: This comment applies to other similar places where `useCallback` is used too.

It's probably not necessary to use `useCallback` (and even could have negative implications to the performance of the app because of the unneeded memoization). Indeed `useCallback` is useful when you want to keep the same instance of the function. In this case it's not needed because they're used either on plain dom nodes or Components that aren't using `memo`.

Although we may want to have some of them memoized just for the sake of being closer to real-life applications? But in that case please add comments stating it explicitely.

resources/todomvc/architecture-examples/preact/src/todo/components/item.jsx
```diff
@@ -0,0 +1,51 @@
...
+    const handleDoubleClick = useCallback(() => {
+        setIsWritable(true);
+    });
+
+    const handleBlur = useCallback(() => {
+        setIsWritable(false);
+    });
```
julienw, line 20: In these calls, because there's no dependency parameter, my understanding is that `useCallback` will always return the new function, so it's pretty much useless. Probably you wanted to pass an empty array instead, but I'd suggest to just get rid of `useCallback` usage at all here... unless we want to recreate development mistake on purpose (?).

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
```diff
@@ -18,6 +18,16 @@ const hasValidMin = (value, min) => {
 };
 
 export function Input({ onSubmit, placeholder, label, defaultValue, onBlur }) {
+    const inputRef = useRef(null);
+
+    useEffect(() => {
```
julienw, line 23: can you please add a comment explaining what this `useEffect` does? For example `Focus and put the cursor at the end of the input when the input appears` (if I understand this well).

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
```diff
@@ -0,0 +1,58 @@
...
+    }, [inputRef.current])
+
+    const handleBlur = useCallback(() => {
+        if (onBlur)
+            onBlur();
+    }, [onBlur]);
+
+    const handleKeyDown = useCallback(
+        (e) => {
+            if (e.keyCode === 13) {
```
julienw, line 38: nit: I would suggest to use `e.key === "Enter"` instead of `keyCode` that is deprecated.

Another option would be to use a `form` and attach the event handler to the form's `submit` event. My understanding is that this is better supported by assistive technologies, but I might be wrong on that :-)

resources/todomvc/architecture-examples/preact/.babelrc
```diff
@@ -1,9 +1,9 @@
...
-    ["transform-react-jsx", { "pragma":"h" }]
-  ]
+    "presets": [
+        "@babel/env",
+        "@babel/preset-react"
+    ],
+    "plugins": [
+        "@babel/plugin-proposal-class-properties"
+    ]
 }
```
julienw, line 9: Do we want an explicit browserslist query?

julienw (commented):
This looks pretty good to me. However I see that the numbers are up in nearly all browsers, and especially in Firefox, by chance do you know why this would happen?

resources/todomvc/architecture-examples/preact/src/todo/components/main.jsx
```diff
@@ -0,0 +1,42 @@
+import { useMemo, useCallback } from "preact/hooks";
+import { getCurrentUrl } from "preact-router";
+
+import { Item } from "./item";
+
+export function Main({ todos, dispatch }) {
+    const route = getCurrentUrl();
+
+    const visibleTodos = useMemo(
```
julienw, line 9: It's probably unnecessary to `useMemo` here because that's just 100 items which isn't much. Moreover in this app probably at each action we're doing `todos` changes, so this makes it even more unneeded.

_That said_ it's also pretty common in real-life apps, so I'm inclined to keep it, but with a comment explaining the trade-off. What do you think?

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
flashdesignory: will add!

resources/todomvc/architecture-examples/preact/src/todo/reducer.js
```diff
@@ -0,0 +1,22 @@
...
+        case "TOGGLE_ITEM":
+            return state.map((todo) => (todo.id === action.payload.id ? { ...todo, completed: !todo.completed } : todo));
+        case "REMOVE_ALL_ITEMS":
+            return [];
+        case "TOGGLE_ALL":
+            return state.map((todo) => (todo.completed !== action.payload.completed ? { ...todo, completed: action.payload.completed } : todo));
+        case "REMOVE_COMPLETED_ITEMS":
+            return state.filter((todo) => !todo.completed);
+        default:
+            return state.slice();
```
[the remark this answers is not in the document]
flashdesignory: makes sense - will change

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
julienw: BTW I see that hte react PR simply uses `autoFocus` for this, doesn't that work here?

resources/todomvc/architecture-examples/preact/.babelrc
flashdesignory: we actually don't need the .babelrc file - will remove

resources/todomvc/architecture-examples/preact/src/todo/components/item.jsx
flashdesignory: the Item is wrapped in a memo function

resources/todomvc/architecture-examples/preact/src/todo/components/item.jsx
flashdesignory: will adjust

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
flashdesignory: that works for the main input in the header, but doesn't work for the input in each item when you try to edit the todo

resources/todomvc/architecture-examples/preact/src/todo/components/item.jsx
flashdesignory: @julienw Preact might be different in this regard (due to the lack of a synthetic event system), but passing functions to DOM nodes isn't free in React. Negligible cost? Probably. But not free. Internally, plain DOM nodes in React have a sort of memo wrapper, and changing event handlers deopts them and causes some reconciliation work that is otherwise unnecessary.

rniwa:
So we're seeing 50% slowdown in Safari and 20% slowdown in Chrome & Firefox. We need to understand why there is such a divergence between browser engines.

flashdesignory:
small update, removing useCallback, in the Item didn't change much. 
There's a better performance in Firefox, but other browsers are similar to before:
https://gist.github.com/flashdesignory/b441fa7c01d10c5d8f72d8954a9b53d7

Additional test with removing memo(<Item/>) looks like there was a tiny performance hit, so I left that in there. 
Also re-verified that wrapping the item with memo did prevent re-rendering of all items when a single item changed.

julienw:
FTR I'm totally fine with keeping the `useCallback` usage, as long as the reasons are documented. My reason is that this _is_ used in real-world apps and therefore it's kinda important that it's part of the benchmark.

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
julienw: It's probably not a big deal for the benchmark itself, all these events should fire synchronously together.

resources/todomvc/architecture-examples/preact/src/todo/components/input.jsx
julienw: That's weird :-) can you please add this explanation to the comment?

rniwa:
I don't think we should take this change without understanding where the performance disparity comes from. We should probably roll back to the old version and make smaller changes incrementally instead.

flashdesignory:
> I don't think we should take this change without understanding where the performance disparity comes from. We should probably roll back to the old version and make smaller changes incrementally instead.

Agreed - I'll keep looking into the issue.

flashdesignory:
Closing this pr in favor of a new phased approach.

flashdesignory:
@julienw new pr is here: https://github.com/WebKit/Speedometer/pull/104

PR #101, opened 2023-03-13
Todomvc backbone update

flashdesignory:
- Add updated readme
- Update to 1.4.1
- Enforce style guide.
- Ensure initial view is consistent with other frameworks

numbers:
https://gist.github.com/flashdesignory/3468e3754ac6d4331ec186b3475e8b31

cc: @kara


flashdesignory:
> FYI I see that `node_modules` is part of this PR :-) As we agreed it would be good to remove it here as well!

uh - yeah - thanks 😓 
- removed

resources/todomvc/architecture-examples/backbone/scripts/build.js
```diff
@@ -0,0 +1,56 @@
...
+        const fileName = filesToMove[i].split("/").pop();
+        await copy(filesToMove[i], `${targetDirectory}/${fileName}`);
+    }
+
+    // read html file
+    let html = await fs.readFile(`${targetDirectory}/${htmlFile}`, "utf8");
+
+    // remove base paths from files to move
+    for (let i = 0; i < filesToMove.length; i++) {
+        const basePath = `${filesToMove[i].split("/").slice(0, -1).join("/")}/`;
```
bgrins, line 41: It seems like something where we replaced i.e. `"node_modules/jquery/dist/jquery.min.js"` with `"jquery.min.js"` would be better than replacing `"node_modules/jquery/dist/"` with `""`. Haven't worked it all the way out but I feel like there could be a bug with subsequent paths in nested subdirectories otherwise. Something like this perhaps (instead of calculating basePath)

```
html = html.replace(filesToMove[i], filesToMove[i].slice(-1))
```

resources/todomvc/architecture-examples/backbone/scripts/build.js
```diff
@@ -0,0 +1,56 @@
...
+        html = html.replace(basePath, "");
+    }
+
+    // remove basePath from source directory
+    const basePath = `${sourceDirectory.split("/")[1]}/`;
```
bgrins, line 46: Globally replacing `src/` in the file feels like it could cause a bug in the general case (I realize we are working in a pretty constrained environment here - but if we want a script like this for other apps it's probably worth attention). Perhaps replacing `="src/` and/or `="./src/` would prevent false positives? Don't feel too strongly here - at some point it may make sense to just use webpack/parcel/etc as well, but I'm OK with a simple script for the Backbone case since that's may be how these apps are deployed.

bgrins:
I made a couple comments on the build script but I'm basically OK with it otherwise

resources/todomvc/architecture-examples/backbone/scripts/build.js
flashdesignory: we could re-use line 33 and update it to:

```
const fileName = filesToMove[i].split("/").pop();
html = html.replace(filesToMove[i], fileName);
```

bgrins: yeah something like that seems fine

flashdesignory: done

resources/todomvc/architecture-examples/backbone/scripts/build.js
flashdesignory: The script has limitations, but so far worked as expected for me. 
If we need something more flexible, we can always replace it with something you mentioned above.

rniwa (approved):

PR #104, opened 2023-03-14
Todomvc preact - Phase 1

flashdesignory:
Speedometer updates:
* gitignore
* eslintignore
* prettierignore
* renamed test name

Preact updates:
* updated preact to latest versions
* cleaned up package.json
* cleaned up Rollup

**Scores**: https://gist.github.com/flashdesignory/ba322dfe48885f2e11011d026af28d42

**Notes**: I compared Rollup to Webpack and it looks like Rollup produces more optimized files that perform better. 
I have to look further into the various output options in Rollup vs Webpack, but since we want to match the scores as closely as possible I kept Rollup. Locally, there was about a 1-2 second increase when testing webpack. 

**TodoMVC specific**: I still want to refactor the app, to reflect a more modern style - that will be phase 2

cc: @kara


flashdesignory:
Additional note regarding the initial scores:
All scores stayed the same expect an increased performance in Firefox: 18 -> 15.
That jump happened when I updated Rollup and cleaned up the configuration file.

rniwa (approved):

bgrins:
Thanks, this looks good to me but please give @julienw a chance to look at it before landing

resources/todomvc/architecture-examples/preact/rollup.config.js
```diff
@@ -1,27 +1,53 @@
...
+import resolve from "@rollup/plugin-node-resolve";
+import commonjs from "@rollup/plugin-commonjs";
+import babel from "@rollup/plugin-babel";
+import css from "rollup-plugin-import-css";
+import copy from "rollup-plugin-copy";
+
+// `npm run build` -> `production` is true
+// `npm run dev` -> `production` is false
+const production = !process.env.ROLLUP_WATCH;
+console.log(production, process.env.ROLLUP_WATCH);
```
julienw, line 11: If we want to keep this log, what do you think of making it somewhat prettier with some prefix?
Or remove it :-)

julienw:
This looks pretty good to me! Thanks for the cleanup :-)

resources/todomvc/architecture-examples/preact/rollup.config.js
flashdesignory: ha - let's remove - thanks for catching this 😄

bgrins (approved):

PR #108, opened 2023-03-16
Todomvc preact 02

flashdesignory:
Last set of updates 🤞 !

* used functional components & hooks
* overall cleanup
* added header component - purely presentational
* added main component - to be able to manage local state separately
* kept model pretty much the same as before 

scores: https://gist.github.com/flashdesignory/212dea03f01a9fc0f6106199aadbe005

cc: @kara


bgrins:
cc @julienw

resources/todomvc/architecture-examples/preact/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": ["eslint:recommended", "plugin:react/recommended"],
+    "rules": {
+        "no-extra-parens": 0,
+        "react/react-in-jsx-scope": 0,
+        "react/no-unknown-property": 0,
+        "react/prop-types": 0
+        // "react/jsx-max-props-per-line": [1, { "maximum": 2 }]
+    }
```
flashdesignory, line 9: local override to mute some noise that the main eslint rule produces otherwise.

resources/todomvc/architecture-examples/preact/src/app/app.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return route;
+};
+
+export default function App() {
+    const [, setUdpatedAt] = useState(Date.now());
+    const [route, setRoute] = useState("all");
+
+    function update() {
+        setUdpatedAt(Date.now());
```
flashdesignory, line 28: since model updates a local array and doesn't use state, the update function that gets called after each model change simply triggers a re-render (**similar to the previous version**).

resources/todomvc/architecture-examples/preact/src/app/item.js
```diff
@@ -1,85 +1,49 @@
...
-            </li>
-        );
+    function handleRemove() {
+        onRemove(todo);
     }
+
+    return (
+        <li class={cx({ completed: todo.completed, editing })}>
+            <div class="view">
+                <input class="toggle" type="checkbox" checked={todo.completed} onChange={handleToggle} />
```
flashdesignory, line 42: yes - uncontrolled on purpose. 
There was a slight increase in performance when I removed local state.

rniwa (approved):

resources/todomvc/architecture-examples/preact/src/app/item.js
```diff
@@ -1,85 +1,49 @@
...
-    handleEdit = () => {
-        let { onEdit, todo } = this.props;
-        onEdit(todo);
-        this.setState({ editText: todo.title });
-    };
+    function handleKeyDown(e) {
+        // prettier-ignore
+        if (e.key === "Escape" || e.key === "ESCAPE")
+            setEditing(false);
+        else if (e.key === "Enter" || e.key === "ENTER")
```
julienw, line 22: same comment as above: can you please add a link to that bug number to fix the benchmark runner?

resources/todomvc/architecture-examples/preact/src/app/model.js
```diff
@@ -1,54 +1,69 @@
...
-    clearCompleted() {
-        this.todos = this.todos.filter( todo => !todo.completed );
-        this.inform();
+    function clearCompleted() {
+        todos = todos.filter((todo) => !todo.completed);
+        inform();
     }
+
+    function getTodos() {
+        return [...todos];
```
julienw, line 57: Is there a good reason to get a copy of the todos information besides defensive programming?

Note that we're calling it twice in `app.js`.

resources/todomvc/architecture-examples/preact/src/app/app.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    const model = TodoModel(update);
+
+    useEffect(() => {
+        function handleHashChange() {
+            setRoute(getRoute());
+            update();
```
julienw, line 36: My understanding is that the `update` is unnecesary here because `setRoute` will already retrigger a render.

resources/todomvc/architecture-examples/preact/src/app/app.js
julienw: Here is another solution to force an update: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate
It's more elegant because it doesn't need to call `Date.now()`, which can by itself be slow on some systems.

Alternatively, you could use `useState` with an integer state and call `setValue(val => val + 1)` but I find this less explicit.

resources/todomvc/architecture-examples/preact/src/app/main.js
```diff
@@ -0,0 +1,21 @@
...
+import TodoItem from "./item";
+import { FILTERS } from "./utils";
+
+export default function TodoMain({ onChange, onToggle, onRemove, onSave, todos, route }) {
+    const visibleTodos = todos.filter(FILTERS[route]);
+    const activeTodoCount = todos.filter(FILTERS["active"]).length;
+
+    return (
+        <section class="main">
+            <input class="toggle-all" type="checkbox" onChange={onChange} checked={activeTodoCount === 0} />
```
julienw, line 13: This doesn't seem to be displayed, or am I missing something?

julienw (commented):
I'm a bit sad because I liked the previous version you wrote (with `useReducer`) :-) I would love to understand why that version was that much slower, despite having some more memoization... Or was it _because_ of the memoization?

This version looks a lot more like the previous implementation and seems to behave somewhat similarly. One difference being that there's no more bound functions like we had before (I know Firefox was suffering from those). For example I think that things like `handleSubmit = () => {` were generating bound functions after transpiling.
It would be good to make sure we have at least one react-like implementation with bound functions because we have them in real-life apps for sure.

Otherwise I left a few comments, tell me what you think :-)

resources/todomvc/architecture-examples/preact/src/app/item.js
```diff
@@ -1,85 +1,49 @@
...
+            handleSubmit(e);
+    }
+
+    function handleDoubleClick() {
+        setEditing(true);
```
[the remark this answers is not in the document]
julienw: We also miss the behavior for the `blur` event. But that can be done in a follow-up if you prefer. (same for autofocus)

resources/todomvc/architecture-examples/preact/src/app/main.js
flashdesignory: @julienw - regarding the use of Date.now() - I tried previously two things:
**the above mentioned forceUpdate solution from the doc** 
using useReducer comes with a cost and performance degraded in preact

**using something simpler like val +1**
this worked when running the workload by itself, but when running it in a test in speedometer it broke.

My assumption is that Date.now() produces a more unique number and ensures that useState doesn't get called with the same value for some reason and therefore prevents an update.

I could be totally off though.

julienw: I think you replied to the wrong comment :)

>  using useReducer comes with a cost and performance degraded in preact

🤔 this is strange but I don't know the hooks enough (nor preact) to understand why this might be.

> using something simpler like val +1

This should work if you use the function style of setting the state: `setValue(val => val+1)` (in this case React will call the function with the value as a parameter) instead of `setValue(value + 1)` where `value` comes from `useState` (in this case we may have an outdated value because of some race).

But no need to spend too much time on this, this is some details. This could also be looked at in follow-ups, we don't need to make it perfect in this PR.

resources/todomvc/architecture-examples/preact/src/app/item.js
```diff
@@ -1,85 +1,60 @@
...
+export default function TodoItem({ onSave, onRemove, onToggle, todo }) {
+    const [editing, setEditing] = useState(false);
+
+
+    /**
+     * useEffect keeps track of the 'editing' state change.
+     * If the input field is present, we set focus programmatically.
+     */
+    useEffect(() => {
+        const input = document.querySelector(".editing");
```
flashdesignory, line 14: @julienw - I tested using useRef and it increased the time. 
This works, but it's not recommended maybe? 

We can either:
1. useRef and be ok with the increased time
2. keep this

resources/todomvc/architecture-examples/preact/src/app/app.js
flashdesignory: added a note

resources/todomvc/architecture-examples/preact/src/app/model.js
flashdesignory: Just for the reason you mentioned. 
At least in our tests I didn't see a difference in performance when I just returned the array vs returning a copy.

resources/todomvc/architecture-examples/preact/src/app/app.js
```diff
@@ -0,0 +1,69 @@
...
+
+    // prettier-ignore
+    if (!FILTERS[route])
+        route = "all";
+
+    return route;
+};
+
+export default function App() {
+    const [, setUdpatedAt] = useState(Date.now());
```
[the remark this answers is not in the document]
flashdesignory: fixed - thanks 😄

resources/todomvc/architecture-examples/preact/src/app/app.js
flashdesignory: removed - thanks 😄

resources/todomvc/architecture-examples/preact/src/app/app.js
```diff
@@ -0,0 +1,69 @@
...
+            setRoute(getRoute());
+            update();
+        }
+
+        addEventListener("hashchange", handleHashChange);
+        handleHashChange();
+    }, []);
+
+    function handleKeyDown(e) {
+        if (e.key === "Enter" || e.key === "ENTER") {
```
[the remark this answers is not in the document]
flashdesignory: merged main with the update

resources/todomvc/architecture-examples/preact/src/app/item.js
flashdesignory: merged main with the update

resources/todomvc/architecture-examples/preact/src/app/app.js
```diff
@@ -0,0 +1,69 @@
...
+
+    useEffect(() => {
+        function handleHashChange() {
+            setRoute(getRoute());
+            update();
+        }
+
+        addEventListener("hashchange", handleHashChange);
+        handleHashChange();
+    }, []);
```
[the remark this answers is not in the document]
flashdesignory: done

bgrins (approved):
Thanks again! There are a few small changes / follow ups called out in the thread. Could you please make sure we don't lose track of those?

bgrins:
To be clear, this can land without addressing them - just want to be sure we get to some kind of resolution to them before we ship. The ones I'm thinking of in particular are 

* Adding a comment https://github.com/WebKit/Speedometer/pull/108#discussion_r1138475797
* Figure out what to do with blur https://github.com/WebKit/Speedometer/pull/108#discussion_r1139096411
* Connect label and input for accessibility  https://github.com/WebKit/Speedometer/pull/108#discussion_r1138648818

Thorsten and I briefly chatted about this and it may make sense to do an accessibility audit for all of the todo apps (as opposed to one-off fixing this issue in the Preact variation). That would of course be done as a follow up.

flashdesignory:
Issue filed: https://github.com/WebKit/Speedometer/issues/112

PR #109, opened 2023-03-16
Allow iterationCount=1

### LINKED ISSUE #107: iterationCount should probably be able to be 1
We currently don't allow `<= 1` (https://github.com/WebKit/Speedometer/blob/6c31329c83815ebb55c0f2cece170a2985b77afc/resources/params.mjs#L31). Obviously niche, but I was just wanting to test a single iteration run to compare with normal runs. @camillobruni is there a reason to prevent 1?

camillobruni:
- Don't calculate standard deviation for single values
- Center single data points on scatter plots to fix track alignment

This resolves #107


bgrins (approved):
Thanks @camillobruni!

PR #110, opened 2023-03-16
Use "Enter" instead of "ENTER" for runner KeyboardEvent

camillobruni:


camillobruni:
I've ran in Safari / Chrome / Firefox to check that it works.

flashdesignory (approved):

PR #113, opened 2023-03-17
TodoMVC - run format from root

flashdesignory:
Clean-up:
Some formatting on recently updated todomvc apps. 

This ensures that when running format from root doesn't introduce file changes in unrelated workloads. 

@kara

### DIFF
--- a/resources/todomvc/architecture-examples/backbone/src/sync/backbone.sync.js
+++ b/resources/todomvc/architecture-examples/backbone/src/sync/backbone.sync.js
@@ -31,6 +31,8 @@ Backbone.sync = (function (method, model, options) {
             size += 1;
 
             // if a success callback was provided, execute it.
+
+            // prettier-ignore
             if (options.success)
                 options.success(model, model.toJSON(), options);
 
@@ -45,6 +47,8 @@ Backbone.sync = (function (method, model, options) {
                 cache[model.id] = model;
 
                 // if a success callback was provided, execute it.
+
+                // prettier-ignore
                 if (options.success)
                     options.success(model, model.toJSON(), options);
 
@@ -52,6 +56,7 @@ Backbone.sync = (function (method, model, options) {
 
                 // if this model doesn't exist yet, we can't update it
             } else {
+                // prettier-ignore
                 if (options.error)
                     options.error(model, "Model not found");
                 deferred.reject(model);
@@ -62,12 +67,15 @@ Backbone.sync = (function (method, model, options) {
             // as long as it exists
             if (cache[model.id]) {
                 // if a success callback was provided, execute it.
+
+                // prettier-ignore
                 if (options.success)
                     options.success(model, cache[model.id].toJSON(), options);
 
                 // resolve
                 deferred.resolve(model);
             } else {
+                // prettier-ignore
                 if (options.error)
                     options.error(model, "Model not found");
                 deferred.reject(model);
@@ -83,6 +91,8 @@ Backbone.sync = (function (method, model, options) {
                 // and trigger the success callback. Note we're passing an
                 // empty object as the second argument, because a deletion
                 // would result in an empty return from the server.
+
+                // prettier-ignore
                 if (options.success)
                     options.success(model, {}, options);
 
@@ -91,6 +101,7 @@ Backbone.sync = (function (method, model, options) {
 
                 // otherwise, error that the model doesn't exist.
             } else {
+                // prettier-ignore
                 if (options.error)
                     options.error(model, "Model not found");
                 deferred.reject(model);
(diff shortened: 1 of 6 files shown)


resources/todomvc/architecture-examples/backbone/src/sync/backbone.sync.js
```diff
@@ -31,6 +31,8 @@ Backbone.sync = (function (method, model, options) {
             size += 1;
 
             // if a success callback was provided, execute it.
+
+            // prettier-ignore
```
bgrins, line 35: I'm happy that this change allows us to run prettier again but it's unfortunate that single line conditions require header text to work (and likely hard to remember when writing patches). What are our options here? I could imagine some alternatives

1) Some tooling changes allow us to reformat single line conditions after prettier runs (maybe an eslint plugin etc)
2) We change the formatting guidelines to put brackets around single line conditions
3) We stop using prettier on js files and only use eslint

I think Thorsten has looked into (1) a bit already and this would be my least favorite option as it sounds like it will take engineering time to solve (as opposed to a configuration change).

(2) would be the simplest technically and is my personal preference - it's a style decision we could decide on and then just re-run prettier as is. I know we've had discussions in the past about changes from the webkit guidelines around bracket style with functions, but we didn't extend that to this single-line conditional rule. I would have brought it up at the time when but didn't realize there was still a difference with what tooling was producing. Perhaps @rniwa is open to making this change?

What are the implications of (3)? I see prettier fixed some whitespace in main-container.jsx but given that the majority of this patch is undoing these conditions maybe it isn't buying us a lot over what's currently getting written?

rniwa: > I'm happy that this change allows us to run prettier again but it's unfortunate that single line conditions require header text to work (and likely hard to remember when writing patches). What are our options here? I could imagine some alternatives
> 
> 1. Some tooling changes allow us to reformat single line conditions after prettier runs (maybe an eslint plugin etc)
> 2. We change the formatting guidelines to put brackets around single line conditions
> 3. We stop using prettier on js files and only use eslint
> 
> I think Thorsten has looked into (1) a bit already and this would be my least favorite option as it sounds like it will take engineering time to solve (as opposed to a configuration change).
> 
> (2) would be the simplest technically and is my personal preference - it's a style decision we could decide on and then just re-run prettier as is. I know we've had discussions in the past about changes from the webkit guidelines around bracket style with functions, but we didn't extend that to this single-line conditional rule. I would have brought it up at the time when but didn't realize there was still a difference with what tooling was producing. Perhaps @rniwa is open to making this change?

I'd prefer anything but (2). I don't want our style to keep deviating from WebKit's coding style guideline since that'll make contributing to Speedometer harder for WebKit contributors.

bgrins: @flashdesignory actually I think the eslint fixing is working fine as-is with the "curly" and "nonblock-statement-body-position" rules. The problem seems to be that it's only running on mjs files (from the package.json `eslint **/*.mjs --fix`). We can see all the files that are requiring the prettier ignore in this script are .js or .jsx. Perhaps we can fix this by changing the task to include more file extensions?

flashdesignory: ha - will do 😄

flashdesignory: @bgrins @rniwa - adjusted the settings a bit and now opting in js, mjs and jsx files.

bgrins: nice cleanup! the lint job is failing in CI https://github.com/WebKit/Speedometer/actions/runs/4493309669/jobs/7904354382?pr=113 but otherwise LGTM

bgrins (approved):

resources/todomvc/vanilla-examples/javascript-es5/src/store.js
```diff
@@ -43,20 +43,20 @@
...
             this,
             todos.filter(function (todo) {
                 for (var q in query) {
-                    if (query[q] !== todo[q]) {
+                    if (query[q] !== todo[q])
                         return false;
-                    }
                 }
+
+
```
rniwa, line 59: This looks like an extra blank line?

rniwa (approved):

resources/todomvc/vanilla-examples/javascript-es5/src/store.js
flashdesignory: removed - ran format on root again and the extra space didn't re-appear. 
must have been a manual clean-up that introduced it.

PR #116, opened 2023-03-21
Add RAF and setTimout based testRun initiation

camillobruni:
Seems like a lot of the perf different isn't coming actually from the  change in the async measurements but rather in using RAF to start the test.
This PR adds `params.testInitiator` to change the mode for local investigation.

RAF-based test initiation seems to decrease the score by 15% on Chrome, by 25% on Safari and 4% on Firefox, which is rather unexpected.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -213,10 +213,12 @@ export class BenchmarkRunner {
         return new Promise(async (resolve) => {
             if (this._client?.willRunTest)
                 await this._client.willRunTest(suite, test);
+            const runTest = this._runTest.bind(this, suite, test, this._page, resolve);
+            if (params.testInitiator === "raf")
+                window.requestAnimationFrame(runTest);
+            else
+                setTimeout(runTest, 0);
 
-            setTimeout(() => {
-                this._runTest(suite, test, this._page, resolve);
-            }, 0);
         });
     }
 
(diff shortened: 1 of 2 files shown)


resources/benchmark-runner.mjs
```diff
@@ -213,10 +213,12 @@ export class BenchmarkRunner {
         return new Promise(async (resolve) => {
             if (this._client?.willRunTest)
                 await this._client.willRunTest(suite, test);
+            const runTest = this._runTest.bind(this, suite, test, this._page, resolve);
+            if (params.testInitiator === "raf")
+                window.requestAnimationFrame(runTest);
```
smaug----, line 218: This isn't doing what we were discussing, or am I missing something. The idea was to have two raf callbacks for the same raf. The first one would run the sync part of the test, the second one would capture the end part of the sync test (that way promise callbacks are part of the sync testing) and then the second raf callback would trigger setTimeout(,0) which records async time.

camillobruni: correct, this is separate. I just realised that this part of my original PR had a massive performance impact, and I'd like to understand why. I think it would make sense to start each sub-tests with an empty task / microtask queue to limit the unmeasured overhead. This I think using an RAF to initiate the test might be a good idea.

camillobruni:
Merging everything back to PR #28 .

PR #117, opened 2023-03-22
Todomvc ember step 01

flashdesignory:
Started from a fresh boilerplate created with ember-cli and therefore there is a large amount of changed files. 
A follow up pr will clean up some files that we don't need (removing lint, so that we can opt into the root format).

For the upgrade and new boilerplate code, it made more sense to rewrite right away, since it's easier to have ember generate files as you build the app. This guarantees that all files are compliant with the desired version. 

most importantly, scores:
https://gist.github.com/flashdesignory/650016ce6b404b4aeacab65995b700bf

@kara

### DIFF
--- a/resources/todomvc/architecture-examples/emberjs/README.md
+++ b/resources/todomvc/architecture-examples/emberjs/README.md
@@ -8,7 +8,15 @@ This application uses Ember 3.28.11 to implement a todo application.
 
 ## Implementation details
 
-[TBD]
+Ember:\
+Model: todo-data.js service\
+View: Ember templates\
+controller: Ember routes and components
+
+MVC:\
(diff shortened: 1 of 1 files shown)


rniwa (approved):

resources/todomvc/architecture-examples/emberjs/.ember-cli
```diff
@@ -5,5 +5,5 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": true
```
bgrins, line 8: Huh. So was this making network requests before?

bgrins (approved):

PR #120, opened 2023-03-24
Todomvc jquery update - Wip

flashdesignory:
Similar to other updates, I opted to rebuild, since upgrading and fixing things that I plan to rewrite doesn't feel to be a time saver. This one is pretty straight forward though and hopefully it's easy to follow along.

- upgraded jquery and dependencies. 
- used latest todo styles 
- updated readme
- updated package.json
- opted into format
- added a build folder (for consistency of having a dist folder to point to)

rebuild specific:
- added an mvc pattern: this was done to have a similar structure as other apps and to have more a "real world" example how a larger app would potentially be structured. 

scores:
https://gist.github.com/flashdesignory/b44b5374e88042d134d6543f6bc30e67

I am trying to pinpoint, why Safari doesn't have as much of a boost as the other browsers. 
The main difference in rendering is that I am not rebuilding the whole list with each update, which I assumed would also benefit safari. 

couple things I tried:
pre-compile handlebars - no difference
html() vs append() - no difference
toggling completed directly without re-rendering the whole todo item - no difference

@kara


rniwa:
Ugh... adding items got slower in Safari but got twice as fast in Chrome & Firefox. I don't think we can accept this level of discrepancy without fully understanding the reason behind it. My suggestion is to start with the existing code and make incremental changes instead of rewriting from scratch.

flashdesignory:
@rniwa - small update with huge difference... 
https://gist.github.com/flashdesignory/0d60a2d16549bfc57e3e001005b14ac2

only thing added was:
`
#main,
            #footer {
                display: none;
            }
`

bgrins:
> only thing added was:
> #main, #footer { display: none; }

If this is modifying the shared stylesheet I would prefer we don't do that - each workload should have the same shared styles and we shouldn't tweak them in one of the tests to change its performance characteristics. I realize this is a tricky one to move forward when combining this feedback with rniwa's comment - sorry about that.

I think the most clear path forward on this one is splitting it - first a PR that does all the paperwork  (latest jQuery and stylesheet, cleaned up build directories etc). If that one is slower in any browser I think we should have good confidence that those are things the browser should be optimizing - similar to the Vue PR. And if it's not (IOW if the rebuild to MVC is causing the discrepancy) we can then have a more focused discussion about how that should move forward.

flashdesignory:
@bgrins I understand and just to ensure we are all aware, the previous todomvc apps are NOT consistent with styles. 
I ensured that what I merged was using the same styles so far for other libraries / frameworks. 

This particular case is tricky, since we are comparing the previous implementations against my updates and the previous implementation had exactly the styles mentioned above:
https://github.com/WebKit/Speedometer/blob/main/resources/todomvc/architecture-examples/jquery/css/app.css

If I remove these styles, we already know that the scores are not exactly what is desired. 

I already tried just doing the upgrade, using consistent styles and removing the above "fix"  (without the js refactor) and that version performs 3x slower in all browsers.

bgrins:
> I understand and just to ensure we are all aware, the previous todomvc apps are NOT consistent with styles.

Yes, 100% and very much appreciate your effort to unify these things here. But just to confirm - as you've been migrating new versions you have been unifying styles in the process, right?

It sounds like your observation is that the performance discrepancy isn't due to the version update or the MVC refactor, but due to the style changes? If that's the case then maybe we can unblock this by having a PR that first does only the version update and project cleanups (linting/formatting/builds/etc).

That should be non controversial to land and then allow a more focused discussion on the two remaining changes - the MVC refactor and the style update to be in line with the other workloads. Does that seem reasonable to you?

flashdesignory:
small variations without custom styles... basically just adding main and footer if there are todos. 
https://gist.github.com/flashdesignory/95bc7296f0be7fae4807c6fabf21c8aa

bgrins:
Will take a closer look this afternoon (including the staged PR, which I guess should continue to include the custom styles as per the discussion above)

flashdesignory:
@bgrins @rniwa - since the phased version is blocked due to the performance of the simple version update, but this version gives us a performance boost that we want to see(although it's a large amount of changes), do we feel comfortable moving forward with this one? 

scores: https://gist.github.com/flashdesignory/95bc7296f0be7fae4807c6fabf21c8aa

rniwa:
I'm getting a bit confused with all the changes being proposed here. What are our options?

bgrins:
> @bgrins @rniwa - since the phased version is blocked due to the performance of the simple version update, but this version gives us a performance boost that we want to see(although it's a large amount of changes), do we feel comfortable moving forward with this one?

FWIW I actually prefer we start with #125 and have r+ed it, as it is a straightforward version update with very few moving parts. Then we can separately look at logic / application changes if any are deemed necessary. It's not clear to me if that issue is truly blocked and and if so why, until someone from WebKit weighs in on it.

flashdesignory:
> I'm getting a bit confused with all the changes being proposed here. What are our options?

This pr has the most optimized and performant upgrade of the jquery workload, but the changes include upgrading version and rewriting the todomvc implementation, which makes it harder to review. 

The reason I broke up the pr into "version 01" was to have a phased implementation:
01. upgrade version only
02. optimize build

Since the [first step](https://github.com/WebKit/Speedometer/pull/125) doesn't give us the performance we want, I believe we are hesitant in moving forward. 

I believe this pr makes the most sense to use since it shows the whole picture of what is intended.

bgrins:
> Since the https://github.com/WebKit/Speedometer/pull/125 doesn't give us the performance we want, I believe we are hesitant in moving forward.

In what way? I think the main concern in https://github.com/WebKit/Speedometer/pull/120#issuecomment-1487963940 was resolved when you re-added the style override (which has also been done in #125).

flashdesignory:
I believe the scores are a factor in approving - at least that's how it seems to me. 

Looking just at the version upgrade the scores are slightly worse, compared to what's in the main branch:
https://gist.github.com/flashdesignory/0fcf2573ff5cf1c086a5b3614c76067e

Looking at the scores from this pr - with the todomvc rebuild - the scores are dramatically better:
https://gist.github.com/flashdesignory/0fcf2573ff5cf1c086a5b3614c76067e

flashdesignory:
closing this pr and follow up with more jquery updates separately

PR #129, opened 2023-04-05
Todomvc angular 01

flashdesignory:
- updated version
- fixed a few typescript issues
- no changes to the todomvc app

scores:
https://gist.github.com/flashdesignory/f85130bf080ef4e46d422aab2613acd5

@kara


resources/todomvc/architecture-examples/angular/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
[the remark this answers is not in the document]
flashdesignory: fixed

resources/todomvc/architecture-examples/angular/README.md
```diff
@@ -1,11 +1,27 @@
...
-## Test locally
+This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 15.2.2.
 
-1. `npm install`
-2. `npm run start-prod`
-3. Navigate to `http://localhost:4200/`
+## Development server
 
-## Build (do this before committing!)
+Run `ng serve` for a dev server. Navigate to `http://localhost:4200/`. The application will automatically reload if you change any of the source files.
```
julienw, line 7: What do you think of refering to the npm scripts rather than `ng`?

flashdesignory: yeah - the whole readme was generated by the angular cli - it's temp for now

resources/todomvc/architecture-examples/angular/tsconfig.json
```diff
@@ -1,20 +1,33 @@
...
-    "moduleResolution": "node",
-    "emitDecoratorMetadata": true,
+    "downlevelIteration": true,
     "experimentalDecorators": true,
-    "target": "es5",
-    "typeRoots": [
-      "node_modules/@types"
-    ],
+    "moduleResolution": "node",
+    "importHelpers": true,
```
julienw, line 18: do you know why this is needed?

bgrins:
I occasionally see a flash of unstyled content when running this locally (in Firefox and Chrome). Have attached a quick screen recording of it. I assume this isn't affecting measured time but I wonder if there's some kind of config that's causing that.

https://user-images.githubusercontent.com/95570/230434472-126fd7e0-f0a2-489f-9eab-4c8d1aafb76e.mov

flashdesignory:
> I occasionally see a flash of unstyled content when running this locally (in Firefox and Chrome). Have attached a quick screen recording of it. I assume this isn't affecting measured time but I wonder if there's some kind of config that's causing that.

oh that's interesting - I didn't notice it locally on my machine.

bgrins:
I wonder if it's something to do with the generated `<link rel="stylesheet" href="styles.5fd101737c0aaaef.css" media="print" onload="this.media='all'">`, I hadn't seen that before. Searching around I found https://stackoverflow.com/questions/67565858/angular-12-css-optimization-inline-event-handler-with-content-security-policy/67582075#67582075 which suggests a configuration that might fix it (`"inlineCritical": false`).

flashdesignory:
> I wonder if it's something to do with the generated `<link rel="stylesheet" href="styles.5fd101737c0aaaef.css" media="print" onload="this.media='all'">`, I hadn't seen that before. Searching around I found https://stackoverflow.com/questions/67565858/angular-12-css-optimization-inline-event-handler-with-content-security-policy/67582075#67582075 which suggests a configuration that might fix it (`"inlineCritical": false`).

oh nice - I'll try it out

rniwa:
Hm... there seems to be a discrepancy between browsers. The new score is roughly 13% slower on both Safari & Chrome but it's 30% slower in Firefox. We should figure out why such a big discrepancy exists.

flashdesignory:
> Hm... there seems to be a discrepancy between browsers. The new score is roughly 13% slower on both Safari & Chrome but it's 30% slower in Firefox. We should figure out why such a big discrepancy exists.

This pr simply updates the version of angular and doesn't touch the todomvc app. 

I was under the impression that updating versions of the frameworks intend was to use workloads that represent what is being used on the web today, independent from a performance impact that might cause. 

Investigating what causes such impact should be maybe a separate task?

rniwa:
> > Hm... there seems to be a discrepancy between browsers. The new score is roughly 13% slower on both Safari & Chrome but it's 30% slower in Firefox. We should figure out why such a big discrepancy exists.
> 
> This pr simply updates the version of angular and doesn't touch the todomvc app.
> 
> I was under the impression that updating versions of the frameworks intend was to use workloads that represent what is being used on the web today, independent from a performance impact that might cause.

"Use workloads that represent what is being used on the web today" is the goal. But in order to achieve that, we need to be mindful of what kind of selection bias we'd be introducing the set of frameworks and their versions we use as well as how those frameworks are used.

It's unclear to me whether the kind of performance difference we're seeing is inherent to angular version difference(s) and if so, whether we're using the most popular version of angular or not. None of that information has been presented as far as I can tell.

bgrins:
> It's unclear to me whether the kind of performance difference we're seeing is inherent to angular version difference(s) and if so, whether we're using the most popular version of angular or not. None of that information has been presented as far as I can tell.

From my perspective - we've been reviewing each change with criteria like that, regardless of scores:

* Analysis of the ecosystem and framework versions - Thorsten has documented and shared a plan for each of these Todo changes (including this one) ahead of the PRs in https://docs.google.com/document/d/1MXxk9DryJmvAmPMt2uMneaS3J5Y-dMFGt4EGvMzhhJE/edit#heading=h.hthjv64z0rm0.
* Correctness / functional changes (as per the FOUC issue we highlighted in this and similar issues in other PRs)
* Usage patterns for a particular version compared with the original implementation (which is not easily quantifiable and requires research / discretion for each, but in general I think taking the existing code through framework migration instructions is a good default)

When there are differential score changes across engines it's a reasonable signal for some scrutiny as it may be the case that something is off. But it's not really surprising that there are perf changes inherit to framework updates, and in that case it should be landed and optimized in the engines.

I think this is the case with Vue (https://github.com/WebKit/Speedometer/pull/114#issuecomment-1480293835) and that it should be landed. We're still taking a look at this PR but I suspect it's probably the case here as well.

bgrins:
@flashdesignory looking at the CLI downloads graph in your doc it seems that 14.2 is more popular than 15.2. Is that a good proxy for popularity in the wild, and if so should we be migrating to that version instead?

FWIW I did some testing with `npx @angular/cli@14 new angular-14-project` and got something running (https://github.com/bgrins/Speedometer/tree/angular-14). It doesn't seem particularly different in practice based on a quick check.

flashdesignory:
@bgrins - downgraded to 14.
Scores:
https://gist.github.com/flashdesignory/3cede61c5b66d0aefc3500f3cd36a028

resources/todomvc/architecture-examples/angular/dist/index.html
```diff
@@ -1,13 +1,12 @@
...
-<html lang="en">
-<head>
+<!DOCTYPE html><html lang="en"><head>
   <meta charset="utf-8">
   <title>Angular</title>
   <base href="/resources/todomvc/architecture-examples/angular/dist/">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" type="image/x-icon" href="favicon.ico">
-<link rel="stylesheet" href="styles.5fd101737c0aaaef.css"></head>
+<style>@charset "utf-8";html,body{margin:0;padding:0}body{font:14px Helvetica Neue,Helvetica,Arial,sans-serif;line-height:1.4em;background:#f5f5f5;color:#111;min-width:230px;max-width:550px;margin:0 auto;-webkit-font-smoothing:antialiased;-moz-osx-font-smoothing:grayscale;font-weight:300}</style><link rel="stylesheet" href="styles.5fd101737c0aaaef.css" media="print" onload="this.media='all'"><noscript><link rel="stylesheet" href="styles.5fd101737c0aaaef.css"></noscript></head>
```
bgrins, line 7: The media=print stylesheet appears to have been restored

rniwa:
> @bgrins - downgraded to 14. Scores: https://gist.github.com/flashdesignory/3cede61c5b66d0aefc3500f3cd36a028

Thanks. There is still discrepancy in scores but I suppose this is inherent to the framework version upgrade since PR doesn't do anything else at this point.

rniwa (approved):

resources/todomvc/architecture-examples/angular/dist/index.html
flashdesignory: removed again 😄

bgrins (approved):

PR #131, opened 2023-04-07
Add a method to the Page object so that we can force a reflow in a frame

julienw:
Some tests may need to force a reflow between steps, so I added this capability to the "page" tool and used it in the runner.

I tested in Firefox and Chrome, but I don't have Mac or Windows so I can't test Safari and Edge. It would be good if somebody could test it as well.

What do you think?

[results.txt](https://github.com/WebKit/Speedometer/files/11180313/results.txt)

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -57,6 +57,10 @@ class Page {
         return null;
     }
 
+    forceReflowAndGetHeight() {
+        return this._frame.contentDocument.body.getBoundingClientRect().height;
+    }
+
     _wrapElement(element) {
         return new PageElement(element);
     }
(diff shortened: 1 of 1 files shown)


resources/benchmark-runner.mjs
```diff
@@ -57,6 +57,10 @@ class Page {
         return null;
     }
 
+    forceReflowAndGetHeight() {
```
rniwa, line 60: I know reflow is the term of art for Firefox but WebKit/Blink uses Layout for the same concept.
How about being more explicit and calling it forceUpdateBoundingBoxes?

But this code may not be necessary after https://github.com/WebKit/Speedometer/pull/28 since all browsers implement HTML5 event loop's update the rendering step these days:
https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering

julienw: The goal is really to be able to use this in tests.

+1 for changing its name.

rniwa: I'd hate for browsers to start adding crazy optimizations around document.body.getBoundingRect though. It's not far fetched with the level of optimizations browsers are implementing these days. My preference is to defer to rAF to measure these things.

julienw: I believe you're right.

Some colleagues mentioned another approach of relying on rAF (that would also have the nice benefit of allowing for Paint, which is what happens in real websites so the benchmark would be closer to these), but that would need the asynchronous runner.

PR #132, opened 2023-04-07
Use act in stockcharts to reduce the issue of missing work happening in rAF callbacks

julienw:
Here is an example of the problem when running the stockcharts workload: https://share.firefox.dev/3KikmR9
At the top of the window is the category graph. Here you can see the brown markers for `requestAnimationFrame` callbacks. They should be included in the `React-Stockcharts.PanTheChart-async` `UserTiming` markers. But from this recording, it's clear that this doesn't always happen. Out of 10 runs, this works as expected just 3 times!

The problem is that React runs some code in a rAF callback (I believe especially since v16 but I'm not so sure about that). This problem might be present in other workloads too.

This patch tries to reduce this problem by using React's [`act`](https://it.reactjs.org/docs/testing-recipes.html#act) function. This should at least make React run its own normally-asynchronous code before returning `act`.

Here is a profile after this change: https://share.firefox.dev/3MqwrXm
Now we see that there's all `React-Stockcharts.PanTheChart-async` `UserTiming` markers` look correct.

There are still `requestAnimationFrame` calls in stockcharts itself:
https://github.com/rrag/react-stockcharts/blob/a32e65681e5d74c4d2008c21c71670c724992df5/src/lib/ChartCanvas.js#L551-L566
https://github.com/rrag/react-stockcharts/blob/a32e65681e5d74c4d2008c21c71670c724992df5/src/lib/ChartCanvas.js#L783-L787
I believe that the latter especially is exercized in our panning steps. When working on this I saw that we were still missing _some_ work sometimes even after this patch. However I don't reproduce this issue today...

It would be good if you could look whether you're seeing this issue on your respective browsers.

I'm not super proud of the code, but I couldn't find another solution. Happy to get your feedback about that!

__Possible alternative__:
1. use an asynchronous runner
2. so that we'd be able to _wait_ until _something_ happens on the screen, as a result of the action
But this would change the performance characteristics of the workload a load.


resources/tests.mjs
```diff
@@ -350,6 +350,7 @@ Suites.push({
     async prepare(page) {
         await page.waitForElement("#render");
     },
+    withReactAct: true,
```
julienw, line 353: When this is set to true, then the runner will wrap the user actions in a call to an exposed `ReactAct` by the workload.

resources/tentative/react-stockcharts/src/index.js
```diff
@@ -38,3 +39,5 @@ render(
 	<ChartComponent />,
 	document.getElementById("root")
 );
+
+window.ReactAct = act;
```
julienw, line 43: The workload has to expose its `act` function so that we can use it from the runner.

rniwa:
Do we need this even after https://github.com/WebKit/Speedometer/pull/28 ? I guess we do if React is pushing all the work until the next rAF? I wonder if a better thing to do is to measure one subsequent rAF.

e.g.
rAF1 -> schedules rAF3 and run sync step & start async timer; this can schedule rAF4, etc...
rAF2 -> run setTimeout(~ , 0) to measure the end of async and schedules rAF5
--- the rest of update the reddening happens here ---
timer scheduled by framework, etc... runs here
timer in rAF2 fires here. This is the end of "async" time.

rAF3 -> start of "subsequent rAF time"
rAF4 -> framework code that supposed to run in rAF runs; this can schedule more async tasks.
rAF5 -> schedule setTimeout(~ , 0) to measure the end of "subsequent rAF time"
--- the rest of update the reddening happens here ---
timer scheduled by framework, etc... runs here
timer in rAF5 fires here. This is the end of "subsequent rAF time" time.

julienw:
I won't say that I understand everything about this, but I know that `act` is used in react-specific test libraries such as react-testing-library for this reason.

julienw:
This was fixed differently, by mocking rAF.

PR #125, opened 2023-03-30
Todomvc jquery update version

flashdesignory:
WIP - I don't think we should merge this:

step 1 - upgraded package versions 
https://gist.github.com/flashdesignory/0fcf2573ff5cf1c086a5b3614c76067e

step 2 - removed custom styles (literally, just deleting the css folder and removing the reference in the html file)
https://gist.github.com/flashdesignory/7a498b928d06b4357f1e9f0bb2f14e93

additional small clean-up tasks that were needed: 
a) remove node_modules folder from the repo
b) add node_modules to gitignore
c) add package-lock file to repo
d) director update needed to point to new module folder

@kara

PR #133, opened 2023-04-07
Todomvc jquery 02

flashdesignory:
Here is the follow up from the jquery version update. 
Main differences are within the src folder, which contains a rebuild of the todomvc application.

Overall scores show a pretty big improvement:
https://gist.github.com/flashdesignory/e20f6ae94ee69bb9c1b3d944fdf00e15

**Changes**

Speedometer
- opt into format from root for jquery

cleanup of workload
- clean up package.json
- clean up readme
- clean up index.html
- address a few ally related items

workload specific
- rebuild todo app with a traditional mvc pattern
- remove director

Biggest performance boost is also a result of not re-rendering the whole todo-list on updates. 
These changes can be seen in the view and template files.

@kara


bgrins:
Just to set expectations up front I’m not sure if we will want to accept this whole change. The challenge to me is that there’s a lot of open space in design decisions here given the diversity of ways people use jQuery (esp compared with some of the other frameworks). This makes it harder to assess whether an update like this is “representative” compared to the implementation today. Not saying the current implementation is great, just that it's already there & had at least some validation as being conventional from the original landing upstream. This type of question wasn't present with a migration to the more popular version of jQuery (#125) which we did support.

I wonder if we might instead take some of the smaller cleanups here, but use the rewritten MVC code for the basis of a Lit implementation rather than changing the jQuery one. Am open to discussion / suggestion here.

flashdesignory:
@bgrins - I'm open to change the implementation. 

There are a few reasons I like this approach: Performance increase and a pattern that represents more accurately how an mvc application can look like. 

The philosophy of the todomvc apps is in my opinion to showcase how a developer can implement a todo app with a mv(*) pattern in mind using various libraries and frameworks. Since jQuery is more like a library of utility tools, it doesn't give directions or sets up implementation details within. Therefore giving an example how one can achieve an mvc pattern can be useful in that perspective.

bgrins:
And just to be clear, when I say "some of the smaller cleanups" I mean everything else in your list other than the rewrite to MVC (formatting, "cleanup of workload", and removing director which AFAICT is unused).

bgrins:
Having quickly looked at this PR I agree that it's a better representation of the MVC approach and definitely a more efficient implementation. Though I don't think Speedometer is the place people should look for reference / showcase implementations as our project goals can be in tension with that (the upstream TodoMVC repo could be updated to that achieve that end and would be a better place to have a detailed discussion on the merits of the change from an architectural standpoint).

Since it does happen that AFAIK we do need new code written for another utility-style library which may not dictate the architecture / implementation details (Lit), it seems feasible to bring some opinionated architecture there. I'm not sure if there are strong conventions in that ecosystem that would dictate a different implementation. May depend on if we used the basic html template strings feature or the higher level Custom Element wrapper classes.

bgrins:
And as per performance in particular, I commented on a similar point on the Vue 03 PR: https://github.com/WebKit/Speedometer/pull/134#issuecomment-1502026442. I don't think that should be a determining factor in the workloads themselves as long as they are within reasonable bounds.

PR #134, opened 2023-04-10
Todomvc vue 03 - Cleanup

flashdesignory:
Some cleaning up and re-org of the todomvc application. 

Added a header and footer component, but kept the todo-items in the todo-list. 

From reading the vue documentation, it is recommended to use stateful component and emitting events to the parent, but in this case this would definitely increase the scores for all browsers (tested locally and seeing about 5+ increase).

Scores stayed the same:
https://gist.github.com/flashdesignory/428115d114549a3adb2cc8d46fb94576 

cc: @kara


resources/todomvc/architecture-examples/vue/README.md
```diff
@@ -1,24 +1,37 @@
...
-```
-npm install
-```
+## Description
 
-### Compiles and hot-reloads for development
-```
-npm run serve
-```
+This application uses React 3.2.47 to implement a todo application.
```
bgrins, line 5: Typo: React -> Vue

flashdesignory: thanks - fixed

bgrins:
Seems like a pretty sensible set of cleanups to me. @julienw, do you have any feedback on it?

> From reading the vue documentation, it is recommended to use stateful component and emitting events to the parent, but in this case this would definitely increase the scores for all browsers (tested locally and seeing about 5+ increase).

Generally speaking, if there is a change to bring an app more in line with recommended practices for a framework it seems fine for scores to increase as that would provide the testcase and incentive to make that recommended practice faster in browsers. There's obv important discretion here in terms of how ubiquitous the pattern is actually used in places that are todo-mvc-shaped. But I think that the total runtime is a relatively unimportant variable in either direction as long as it's within reasonable bounds.

flashdesignory:
Maybe this pr is the next step of changes and once approved I could push up a new pr just with the the addition of todo-item components?

bgrins:
Yeah for sure, I don't think we need to bundle here

resources/todomvc/architecture-examples/vue/README.md
```diff
@@ -1,24 +1,37 @@
...
+Vue.js is focused on the ViewModel layer of the MVVM pattern. It connects the View and the Model via two way data bindings. Actual DOM manipulations and output formatting are abstracted away into Directives and Filters.
+
+MVC:\
+Model: maintains the data and behavior of an application\
+View: displays the model in the ui\
+Container: serves as an interface between view & model components
+
+## Built steps
+
+To build the static files, this application utilizes webpack. It minifies and optimizes output files and copies all necessary files to a `dist` folder.
```
julienw, line 20: What do you think of adding the command used to build?

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -61,50 +59,48 @@ export default {
...
         }
     },
     computed: {
-        remaining() {
-            return this.todos.filter(todo => !todo.completed).length
-        },
-        pluralizedWord() {
-            return this.remaining === 1 ? 'item' : 'items'
+        activeTodos() {
+            return filters['active'](this.todos);
```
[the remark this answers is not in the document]
julienw: Also, why don't you use the simpler `filters.active` instead of `filters['active']`? (same comment for `completedTodos` below)

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -61,50 +59,48 @@ export default {
...
     methods: {
-        addTodo() {
+        addTodo(value) {
             this.todos.push({
                 completed: false,
-                title: this.newTodo
-            })
-            this.newTodo = ''
+                title: value,
+                id: uuid()
```
julienw, line 66: Am I right that the previous version had no `id` value despite that we were refering to it in the loop?
(nothing to do here, thanks for the change!)

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -17,36 +13,38 @@
...
+        <TodoFooter :todos="todos" @delete-completed="deleteCompleted" :remaining="activeTodos.length" :completed="completedTodos.length" :route="route"/>
     </section>
 </template>
 
 <script>
 import { nextTick } from 'vue';
+import TodoHeader from './TodoHeader.vue';
+import TodoFooter from './TodoFooter.vue';
+
+const uuid = () => crypto.randomUUID();
```
julienw, line 32: I noticed recently that this isn't working when accessing the app using http on the network, for example when testing from Android to a local dev server.
What do you think of using a simple locally incremented var?

(this could be a change accross the other implementations in an upcoming PR)

julienw (approved):
Thanks, this all look good to me.

I think there are a few comments that are worth handling now, others that are worth discussing later. I'll let you decide what you keep for later :-)

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -17,36 +13,38 @@
...
                     <div class="view">
                         <input type="checkbox" v-model="todo.completed" class="toggle">
-                        <label @dblclick="editTodo(todo)">{{ todo.title }}</label>
+                        <label @dblclick="startEdit(todo)">{{ todo.title }}</label>
                         <button class="destroy" @click.prevent="deleteTodo(todo)"></button>
                     </div>
-                    <input type="text" class="edit" v-model="todo.title" @keyup.enter="doneEdit" @blur="doneEdit"
-                        v-todoFocus="todo === editing" />
+                    <input type="text" class="edit" v-model="todo.title" @keyup.enter="finishEdit" @blur="finishEdit" v-todoFocus="todo === editing" />
                 </li>
```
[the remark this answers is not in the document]
julienw: Ah I just see your line in the PR description and the follow-up discussion!
I'd be happy to see a PR with just this change, so that we can study it in isolation.

resources/todomvc/architecture-examples/vue/README.md
flashdesignory: ✅

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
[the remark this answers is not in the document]
flashdesignory: both done - was a simple opt-in for the format command

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
flashdesignory: oh interesting - changed it to the other uuid() method that we are using. 
I'll open a separate pr that'll remove the other usages throughout the todomvc apps.

bgrins (approved):

PR #135, opened 2023-04-11
Todomvc svelte 01

flashdesignory:
First implementation (or phase) of Svelte TodoMVC app. 
It's based off of this pr: https://github.com/WebKit/Speedometer/pull/68

Main difference:
- it's using the same styles as all the other todo apps.
- it's using same functionality as the other todo apps
- some minor clean up

It's not using an mvc pattern, but it's pretty close the the official app and hopefully that makes it easy to review. 

I know you all will ask, so here are scores (it's the comparison from the original pr to this pr):
https://gist.github.com/flashdesignory/05b4c2ee2449bc24aee262f4d3b0327f

@kara


resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        editing = null;
+    }
+
+    function toggleAllItems(event) {
+        items = items.map((item) => ({
+            id: item.id,
+            description: item.description,
+            completed: event.target.checked,
```
julienw, line 47: nit: I'd use 
```js
{
  ...item,
  completed: event.target.checked,
}
```
so that we don't have to change it if for any reason the structure of an item changes in the future.

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,121 @@
...
+        }));
+    }
+
+    function removeCompletedItems() {
+        items = items.filter((item) => !item.completed);
+    }
+
+    function handleEdit(event) {
+        if (event.key === "Enter") event.target.blur();
+        else if (event.key === "Escape") editing = null;
```
julienw, line 57: thanks for using event.key 👍

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,121 @@
...
+    <section class="main">
+        <input id="toggle-all" class="toggle-all" type="checkbox" on:change={toggleAllItems} checked={numCompleted === items.length} />
+        <label for="toggle-all">Mark all as complete</label>
+
+        <ul class="todo-list">
+            {#each filtered as item, index (item.id)}
+                <li class="{item.completed ? 'completed' : ''} {editing === index ? 'editing' : ''}">
+                    <div class="view">
+                        <input class="toggle" type="checkbox" bind:checked={item.completed} />
+                        <!-- svelte-ignore a11y-label-has-associated-control -->
```
julienw, line 88: I like that svelte flags all the small accessibility issues that we've seen in other implementations in the past :-) (not to be changed here)

resources/todomvc/architecture-examples/svelte/src/utils.js
```diff
@@ -0,0 +1,7 @@
+export function uuid() {
+    return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, function (c) {
+        var r = (Math.random() * 16) | 0,
+            v = c === "x" ? r : (r & 0x3) | 0x8;
+        return v.toString(16);
+    });
+}
```
julienw, line 7: have you taken this from somewhere? If yes, can you please add a comment with a pointer? Is it worth replacing with an incremented local variable instead?

resources/todomvc/architecture-examples/svelte/README.md
```diff
@@ -0,0 +1,36 @@
+# Speedometer 3.0: TodoMVC: Svelte
```
julienw, line 1: it could be good to mention that this has been heavily inspired by https://github.com/sveltejs/svelte-todomvc

resources/todomvc/architecture-examples/svelte/src/utils.js
flashdesignory: yes, from here: https://github.com/sveltejs/svelte-todomvc/blob/master/src/TodoMVC.svelte#L64

I wanted to keep this pr as close to the original version as possible, without introducing too many changes.

resources/todomvc/architecture-examples/svelte/src/App.svelte
flashdesignory: ✅

resources/todomvc/architecture-examples/svelte/src/utils.js
julienw: yeah I don't mind to keep it. It seemed a bit overkill to me at first, but rhere's also value in having a string.replace with a function callback in the benchmark IMO.

julienw (approved):

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,120 @@
+<script>
+	import { tick } from 'svelte';
```
rniwa, line 2: Tab character?

PR #137, opened 2023-04-11
Use metrics for CSV

camillobruni:
- Enforce better default order of metrics
- Introduce Metric.separator to be used by metrics-ui and CSV generation


resources/main.mjs
```diff
@@ -266,48 +269,19 @@ class MainBenchmarkClient {
     }
 
     _getFormattedCSVResult() {
-        let tests = [];
         // The CSV format is similar to the details view table. Each measurement is a row with
         // the name and N columns with the measurement for each iteration:
         // ```
         // Measurement,#1,...,#N
         // TodoMVC-JavaScript-ES5/Total,num,...,num
```
bgrins, line 276: I did a quick check with this applied and it drops the "/Total" from the subtest total string. I've been using that string to make analysis easy - is it possible to keep that?

camillobruni: I'm a bit torn between that. I'd prefer having it all consistent between the metrics view and the exported data (minus the backwards-compatible JSON).
I'd be tempted to add `/Total` to any aggregated metric.

 - `TodoMVC-React-Redux` => `TodoMVC-React-Redux/Total`
 - `TodoMVC-React-Redux/Adding100Items` => `TodoMVC-React-Redux/Adding100Items/Total`
 
 WDYT?

bgrins: Sure, I can some up with another heuristic for splitting (ends in "/Total" and has a single "/" or something). Do you prefer your suggested approach here (/Total for any aggregate) or something else?

rniwa:
Looks like this CSV doesn't have geomean or score for the entire thing?

bgrins:
That data is shaped differently than the existing csv so it would be awkward to fit in I think. Plus at least for my use case the score isn’t useful, and I doubt the audience for this raw export is much broader than the folks in this thread.On May 15, 2023, at 4:04 PM, Ryosuke Niwa ***@***.***> wrote:﻿
Looks like this CSV doesn't have geomean or score for the entire thing?

rniwa:
> That data is shaped differently than the existing csv so it would be awkward to fit in I think.

I'd definitely would like to see the geomean & score in these CSV.

camillobruni:
Should we append them as an additional col to the Total and Score lines?
Alternatively (and likely easier) I can add a `Total geomean` and `Score geomean` line with a single entry.

rniwa:
> Should we append them as an additional col to the Total and Score lines? Alternatively (and likely easier) I can add a `Total geomean` and `Score geomean` line with a single entry.

You mean additional rows? That would make most sense to me.

camillobruni:
```
Name,#1,#2
Metric,1.1,2.2
Total,1.1,2.2,1.556
```

vs.
```
Name,#1,#2
Metric,1.1,2.2
Total,1.1,2.2,
Total geomean,1.556
```

camillobruni:
Oh, we can simply do the following, which is more consistent:

```
Name,#1,#2,mean
Metric,1.1,2.2,1.65
Total,1.1,2.2,1.65
```

camillobruni:
Screenshot 2023-05-16 at 00 12 03

rniwa:
It would be nice if we can have Geomean instead of / in addition to Total.

camillobruni:
The "Total" should have been the geomean, I've only fixed the but in #146 (added the fix here as well).

Screenshot 2023-05-16 at 09 58 37

camillobruni:
Added the minimal changes here to use "Geomean" instead of "Total".

PR #139, opened 2023-04-11
Todomvc vue 04 - use component for item

flashdesignory:
As mentioned in a [previous pr](https://github.com/WebKit/Speedometer/pull/134), this update uses a separate stateful component for the todo-item, which emits events on toggle, update, remove.

scores:
https://gist.github.com/flashdesignory/74861a421134ce3a4a6c5c12303f3a4a

@kara


resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -72,18 +55,18 @@ export default {
...
             });
         },
+        toggleTodo(todo, value) {
+            todo.completed = value;
+        },
         deleteTodo(todo) {
             this.todos = this.todos.filter((t) => t !== todo);
         },
+        editTodo(todo, value) {
+            if (todo) todo.title = value;
```
rniwa, line 65: todo.title should be on a new line (indented).

resources/todomvc/architecture-examples/vue/src/components/TodoItem.vue
```diff
@@ -0,0 +1,81 @@
...
+    methods: {
+        startEdit() {
+            this.editing = true;
+            nextTick(() => {
+                this.focusEditInput();
+            });
+        },
+        finishEdit() {
+            this.editing = false;
+            if (this.editText.trim().length === 0) {
```
rniwa, line 41: No curly braces around single line statements.

flashdesignory: same as above

rniwa (approved):

julienw:
Trying this locally I realize that `package-lock.json` is missing and should be committed, so that the build is reproducible.

flashdesignory:
> Trying this locally I realize that `package-lock.json` is missing and should be committed, so that the build is reproducible.

added

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -7,24 +7,17 @@
...
             <ul class="todo-list">
-                <li class="todo" :key="todo.id" v-for="todo in filteredTodos" :class="{ completed: todo.completed, editing: todo === editing }">
-                    <div class="view">
-                        <input type="checkbox" v-model="todo.completed" class="toggle" />
-                        <label @dblclick="startEdit(todo)">{{ todo.title }}</label>
-                        <button class="destroy" @click.prevent="deleteTodo(todo)"></button>
-                    </div>
-                    <input type="text" class="edit" v-model="todo.title" @keyup.enter="finishEdit" @blur="finishEdit" v-todoFocus="todo === editing" />
-                </li>
+                <TodoItem v-for="todo in filteredTodos" :key="todo.id" v-memo="[todo.completed, todo.title]" :todo="todo" @delete-todo="deleteTodo" @edit-todo="editTodo" @toggle-todo="toggleTodo" />
```
julienw, line 10: It would be good to see the numbers without `v-memo`. My understanding is that it is superfluous. Indeed according to https://vuejs.org/guide/best-practices/performance.html#props-stability Vue memoizes the props already, so if `todo` stays the same (and I believe it changes only when we edit a property thanks to the internal use of proxies) the update will be skipped already.
Therefore I believe that using `v-memo` adds work for no benefit. It would be good to check :-)

julienw (approved):
This looks reasonable to me, except for the use of `v-memo` as I explain below. Please tell me what you think!

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
flashdesignory: @julienw - let me remove the v-memo call. There wasn't any measurable difference in using it and I was going to remove it as well.

julienw:
> > Trying this locally I realize that `package-lock.json` is missing and should be committed, so that the build is reproducible.
> 
> added

Thanks, now I when I run `npm i` locally there's no local change anymore.

resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
flashdesignory: done

PR #142, opened 2023-04-14
Todomvc angular update / cleanup

flashdesignory:
The previous pr was strictly focused on updating the framework version, without considering how Angular developers actually build a typical Angular app. 

This pr is a slight rebuild of the todo application with that in mind, which is a better representation of how an Angular app looks like.

- readme update
- package.json update
- index.html cleanup
- adding proper routing
- splitting up the single component into smaller components

Scores:
https://gist.github.com/flashdesignory/4241450e74506982693ce9131d911692

@kara


resources/todomvc/architecture-examples/angular/src/app/todo-header/todo-header.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, ChangeDetectionStrategy } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { TodosService } from '../todos.service';
+
+@Component({
+  selector: 'app-todo-header',
+  templateUrl: './todo-header.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
```
julienw, line 8: Can you please explain in a comment why you need to change the `changeDetection` parameter here?

resources/todomvc/architecture-examples/angular/src/app/todo-item/todo-item.component.ts
```diff
@@ -0,0 +1,58 @@
+import { Component, Input, Output, EventEmitter, ViewChild, ElementRef, AfterViewChecked, ChangeDetectionStrategy } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { Todo } from '../todo';
+
+@Component({
+  selector: 'app-todo-item',
+  templateUrl: './todo-item.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class TodoItemComponent implements AfterViewChecked{
```
julienw, line 10: nit: space before `{`

Is prettier running on this codebase?

resources/todomvc/architecture-examples/angular/src/app/app.module.ts
```diff
@@ -1,18 +1,26 @@
...
+    TodoFooterComponent,
+    TodoItemComponent
   ],
   imports: [
     BrowserModule,
-    FormsModule,
-    HttpClientModule
+    AppRoutingModule,
+    ReactiveFormsModule
   ],
```
julienw, line 24: Note: it looks like that newer versions of Angular do not need NgModule anymore, see https://angular.io/guide/standalone-components

I wonder if we shouldn't migrate to that.

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.todos = this.todos.map((todo) => todo !== todoToToggle ? todo : { ...todo, completed: !todo.completed });
+  }
+
+  removeItem(todo: Todo):void {
+    this.todos = this.todos.filter((t) => t !== todo);
+  }
+
+  updateItem(todoToSave: Todo, title: string) {
+    this.todos = this.todos.map((todo) => todo !== todoToSave ? todo : { ...todo, title });
+  }
```
julienw, line 42: It would be good to check whether angular would work with mutations rather than generating the whole list (and of course look at the performance difference if this works). I'm under the impression that Angular works from the user inputs rather than the values, but I merely skimmed through the documentation.

If mutations work, it would be a good idea to use mutations when possible.

julienw (commented):
I'm far from knowing Angular much, but here are some thoughts reading through this update. Tell me what you think :-)

resources/todomvc/architecture-examples/angular/README.md
```diff
@@ -1,27 +1,44 @@
...
+Angular:\
+Model: todos.service\
+View: angular component's html\
+ViewModel: angular component's ts
 
-## Build
+MVC:\
+Model: maintains the data and behavior of an application\
+View: displays the model in the ui\
+Container: serves as an interface between view & model components
```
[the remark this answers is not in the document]
flashdesignory: uhm, yeah

resources/todomvc/architecture-examples/angular/src/app/app.component.css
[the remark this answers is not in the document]
flashdesignory: will remove

resources/todomvc/architecture-examples/angular/src/app/todo-header/todo-header.component.ts
flashdesignory: removed from header, after running tests with no measurable difference

resources/todomvc/architecture-examples/angular/src/app/todo-item/todo-item.component.ts
```diff
@@ -0,0 +1,58 @@
+import { Component, Input, Output, EventEmitter, ViewChild, ElementRef, AfterViewChecked, ChangeDetectionStrategy } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { Todo } from '../todo';
+
+@Component({
+  selector: 'app-todo-item',
+  templateUrl: './todo-item.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
```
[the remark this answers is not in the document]
flashdesignory: Actually, here it is useful and you can see a difference in scores. 
I'll add some comments on this one

resources/todomvc/architecture-examples/angular/src/app/todo-item/todo-item.component.ts
flashdesignory: will check

resources/todomvc/architecture-examples/angular/src/app/app.module.ts
flashdesignory: hmm will look into it.

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: will check

resources/benchmark-report.mjs
```diff
@@ -16,7 +16,7 @@
     document.head.appendChild(styleElement);
 
     let createTest;
-    const valuesByIteration = new Array();
+    const valuesByIteration = [];
```
flashdesignory, line 19: root format changed this - I can revert if we want to keep the original way.

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: interesting - I made the change to use mutation for toggle item and update item. 
Seems like there's a slightly better performance:
https://gist.github.com/flashdesignory/80970c02d90a8b29e611fe8bfb53e5ba

resources/todomvc/architecture-examples/angular/src/app/todo.ts
```diff
@@ -1,9 +1,3 @@
...
-  id: number = 0;
-  title: string = '';
-  complete: boolean = false;
-
-  constructor(values: Object = {}) {
-    Object.assign(this, values);
-  }
-}
\ No newline at end of file
+    constructor(public id: string, public title: string, public completed: boolean) { }
```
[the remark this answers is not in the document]
flashdesignory: This was somewhat outdated and I am using now a simple Todo interface for the todo-item.

resources/todomvc/architecture-examples/angular/src/app/todo-item/todo-item.component.ts
flashdesignory: opted into root format 😄

resources/benchmark-report.mjs
julienw: Both are identical to me, so let's keep the root format :-)

resources/todomvc/architecture-examples/angular/src/app/todo-item/todo-item.component.ts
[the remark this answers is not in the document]
julienw: I asked around, and I've been answered that OnPush is actually a best practice these days.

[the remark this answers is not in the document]
julienw: If you're interested, my friend told me about this: https://github.com/angular/angular/discussions/49685
This is a discussion about what they intend to do to track changes. 
But interesting for us is the Appendix that is "how things work currently".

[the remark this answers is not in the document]
julienw: (thanks @nud)

flashdesignory:
@bgrins @rniwa @julienw - any thoughts?

rniwa:
Would it be possible to reach out to the folks working on angular & ask which version is more representative? It's hard for me to judge which version is more representative.

flashdesignory:
> Would it be possible to reach out to the folks working on angular & ask which version is more representative? It's hard for me to judge which version is more representative.

@rniwa - Yes, this pr was already reviewed internally by someone who worked on the Angular team!

rniwa (approved):
Sounds good then.

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
```diff
@@ -0,0 +1,55 @@
...
+            id: uuid(),
+            title,
+            completed: false,
+        };
+
+        this.todos = this.todos.concat(todo);
+    }
+
+    removeItem(todo: Todo): void {
+        const index = this.todos.findIndex((t) => t.id === todo.id);
```
julienw, line 33: I wonder if `this.todos.indexOf(todo)` would work. Have you tried?

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    todos: Todo[] = [];
+
+    addItem(title: string): void {
+        const todo: Todo = {
+            id: uuid(),
+            title,
+            completed: false,
+        };
+
+        this.todos = this.todos.concat(todo);
```
julienw, line 29: Can we just push to the same array instead of creating a new one here?

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
```diff
@@ -0,0 +1,55 @@
...
+        const index = this.todos.findIndex((t) => t.id === todo.id);
+        this.todos.splice(index, 1);
+    }
+
+    clearCompleted(): void {
+        this.todos = this.todos.filter((todo) => !todo.completed);
+    }
+
+    toggleAll(completed: boolean): void {
+        this.todos = this.todos.map((todo) => ({ ...todo, completed }));
```
julienw, line 42: Can we toggle `completed` by looping instead of creating a new array?
For example we could use `for of`, which would also have the nice side effect of exercizing a different part of the JS engine (the iterable protocol).

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    getItems(type = "all"): Todo[] {
+        switch (type) {
+            case "active":
+                return this.todos.filter((todo) => !todo.completed);
+            case "completed":
+                return this.todos.filter((todo) => todo.completed);
+        }
+
+        return [...this.todos];
```
julienw, line 53: Why don't you just `return this.todos` here?

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    clearCompleted(): void {
+        this.todos = this.todos.filter((todo) => !todo.completed);
+    }
+
+    toggleAll(completed: boolean): void {
+        this.todos = this.todos.map((todo) => ({ ...todo, completed }));
+    }
+
+    getItems(type = "all"): Todo[] {
```
julienw, line 45: I was curious, so I added a log here, and I noticed that this is called 14 times (!!) with  very often (for example on blur, or for each typed character in the header).
Can you please check if there's a way of _not_ doing that? Is there something native to do it only when needed, or maybe can we cache the result, invalidating the cache when some changes are made to the todos array?

julienw: I think this would be a great use of a pipe (https://angular.io/guide/pipes), which AFAIU is similar to Redux selectors.
However my understanding is that with pipes we either need to go back at the immutable way of doing things: use map, filter, etc, (in that case please disregard my other comments above + change the `removeItem`) or mark the pipe as "impure", and in that case it would be evaluated once for every keystroke and other events (which would still be an improvement over the current situation!).

Tell me what you think :-)

julienw (commented):
This looks good to me, however I think we're calling the filtering functions in the service too much. Can we try to do something about it? I've put some thoughts below but tell me what you think about those!

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: What's the benefit to use `indexOf` over `findIndex`?

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: similar answer as I had for other prs 😄 

I'll change it though so we can get this merged

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: I'll take a look - thanks for surfacing. 
At this point I rather fix what we have though instead of refactoring again.

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: since the item uses `ChangeDetectionStrategy.OnPush`, this wouldn't work

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
julienw: `concat` creates a new array, `push` doesn't

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
julienw: I think this is faster: `indexOf` would be implemented completely in c++, while with `findIndex` a JS function has to be called repeatidly.
Using findIndex is the correct solution if the `todo` parameter isn't the same as the one in the array though.
I'm defering to your judgment!

resources/todomvc/architecture-examples/angular/src/app/todos.service.ts
flashdesignory: @julienw what do you want me to try? 

I did test out the pipe idea, but didn't see any noticeable performance benefit. 
Here is the test branch I was using: 
https://github.com/flashdesignory/Speedometer/blob/todomvc-angular-test/resources/todomvc/architecture-examples/angular/src/app/todos.pipe.ts

julienw: Thanks for the pointer! I'll have a look tomorrow, to see if there's anything obvious.
Otherwise I wanted to implement some cache like mentioned in the previous comment. I can also have a look at it of you want, see if this changes things.

flashdesignory: Oh absolutely - it's a team effort and I welcome direct contributions 😄

julienw: With your code, the pipe was still running as often as before. But I made it work properly by replacing `getTodos` to return `this.todos` directly instead of slicing it (in `todos.service`).

But... this made no difference to the benchmark result. My guess is that we don't exercise this code much in the benchmark as we assign the values directly to the `value` attribute, so the fact that it runs for each keystroke doesn't matter.
So even with this change, like you I see that the pipe implementation (along with using immutable structures) is slower than the current code.

I also tried to implement a cache "manually" but the fact that we change the `completed` value directly on the `todo` item makes things a bit messy.

All in all I think we went through all paths now and I'm happy about landing this version.

julienw (approved):

bgrins (approved):
Setting flag based on Julien's review

PR #144, opened 2023-04-17
Add a new charting workload using observable plot

julienw:
Hey folks,

This is a new charting workload using observable plot, and therefore d3 under the hood. It contains 3 different graphs:
* stacked bars
* grouped bars
* a dotted graph

The grouped bars is currently disabled in the benchmark steps because we found it doesn't bring a lot of value compared to the stacked bars.

Please tell me what you think!

Here are the runs in various browsers, courtesy of @flashdesignory (thanks!): https://gist.github.com/julienw/ebdf1b3a01a8e28212a1ebf15df5ad45. It looks like the first run always has some penalty.

image


julienw:
I'm on Linux so I can't run the 3 browsers on my computer. I'd appreciate if somebody would do it.

flashdesignory (commented):
Thanks for all the comments throughout the code. That makes it a lot simpler to follow along. I still have to run it locally, but looks really clean.

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,253 @@
...
+    document.querySelector("#chart").append(Plot.plot(options));
+}
+
+function addGroupedBars() {
+    if (!isReady())
+        throw new Error("Please prepare the data first.");
+
+    const options = {
+        width: 2000,
+        height: 1000,
```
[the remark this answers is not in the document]
julienw: I should write in a comment that they are really max values, the width is adjusting with the available widths.

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,253 @@
...
+        const infoForDestinationAirport = flightsByAirport.get(destination) ?? { origin: 0, destination: 0, total: 0 };
+        infoForDestinationAirport.destination += intCount;
+        infoForDestinationAirport.total += intCount;
+        flightsByAirport.set(destination, infoForDestinationAirport);
+    }
+
+    return flightsByAirport;
+}
+
+function isReady() {
```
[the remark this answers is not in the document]
julienw: Yeah that would be better indeed

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,253 @@
...
+            // horizontal bottom line
+            Plot.ruleY([0]),
+        ],
+    };
+
+    document.querySelector("#chart").append(Plot.plot(options));
+}
+
+function reset() {
+    document.querySelector("#chart").textContent = "";
```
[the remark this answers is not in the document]
julienw: copy-pasta programming :-)
I have a ROOT constant now, so that I don't have to run `querySelector` each time.

julienw:
I added a README file.

julienw:
I added timings from @flashdesignory's computer to the PR description.

resources/tentative/charts/dist/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
rniwa, line 1: Can we rename this to developer.html or something?

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+};
+
+let preparedData;
+function prepare() {
+    /**
+     * AirportInformation: { state, iata, name, city, country, latitude, longitude }
+     * airports: Array<AirportInformation>
+     * flights: Array<{ origin, destination, count }>
+     */
+    const { airports, flights } = parseAirportsInformation(airportsString);
```
rniwa, line 27: This function doesn't seem to take any arguments?

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+
+    /* Array<[state, AirportInformation[]]> */
+    const airportsGroupedByStateArray = d3Array.groups(airports, (d) => d.state);
+
+    /* DescSortedArray<[{ state: string, total: number, mostUsedAirportsInState: AirportInformation[] }]> */
+    const stateInformationSortedArray = airportsGroupedByStateArray
+        .map(([state, airportsInState]) => {
+            const totalFlightsInState = d3Array.sum(airportsInState, ({ iata }) => flightsByAirport.get(iata)?.total);
+            const sorted = d3Array.sort(airportsInState, ({ iata }) => -flightsByAirport.get(iata)?.total);
+            const mostUsedAirportsInState = sorted.slice(0, 6);
```
rniwa, line 45: I'm a bit confused here. The charts in your screenshot seems to have 7 items, not 6.
That seems rather arbitrary anyway though. Why not top 10 for example instead?
Maybe this should be another configurable parameter across test runs.

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+        y: { grid: true, tickFormat: "~s" },
+        marks: [
+            // stacked bars
+            Plot.barY(preparedData.plotData, {
+                x: "state",
+                y: "total",
+                fill: "index",
+                title: (d) => `${d.iata === "Other" ? "Other" : `${d.name}, ${d.city} (${d.iata})`}\n${d3Format(",")(d.total)} flights`,
+            }),
+            // labels
```
rniwa, line 144: This comment seems unnecessary? The code below is pretty self-explanatory.

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+            Plot.text(preparedData.stateInformationSortedArray, { x: "state", y: "total", text: (d) => d3Format(".2~s")(d.total), dy: -10 }),
+            // horizontal bottom line
```
rniwa, line 146: Ditto.

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+
+    const options = {
+        ...DEFAULT_OPTIONS,
+        color: { type: "categorical" },
+        x: {
+            domain: preparedData.stateSortedArray,
+        },
+        y: { grid: true, tickFormat: "~s" },
+        marks: [
+            // stacked bars
```
rniwa, line 137: This comment seems unnecessary given this function is called addStackedBars.

resources/tests.mjs
```diff
@@ -344,6 +344,31 @@ Suites.push({
...
+    async prepare(page) {},
+    tests: [
+        new BenchmarkTestStep("Prepare", (page) => {
+            page.querySelector("#prepare").click();
+        }),
+        new BenchmarkTestStep("Stacked", (page) => {
+            page.querySelector("#reset").click();
+            page.querySelector("#add-stacked-chart-button").click();
+        }),
+        /* Skip the Grouped Chart because it doesn't bring a lot of value compared to Stacked
```
rniwa, line 359: I think we should just remove the code.

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+    };
+
+    ROOT.append(Plot.plot(options));
+}
+
+function reset() {
+    ROOT.textContent = "";
+}
+
+async function runAllTheThings() {
```
rniwa, line 246: Do we need this function / feature?

resources/tentative/charts/observable-plot.js
```diff
@@ -0,0 +1,265 @@
...
+            // labels
+            Plot.text(preparedData.stateInformationSortedArray, { x: "state", y: "total", text: (d) => d3Format(".2~s")(d.total), dy: -10 }),
+            // horizontal bottom line
+            Plot.ruleY([0]),
+        ],
+    };
+    ROOT.append(Plot.plot(options));
+}
+
+// Note that this function is currently commented out in the benchmark. We may
```
rniwa, line 153: I think we should just delete it. It's easy enough to add back if we see the need for it.

resources/tentative/charts/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
rniwa, line 1: Can we call this one developer.html or something and the other one index.html?

julienw: I can rename this one, but I'd prefer to keep the other one with the original name, so that we can add more html files for more workloads (like I did for chart.js).

resources/tentative/charts/observable-plot.js
julienw: This is useful when developing.

resources/tentative/charts/observable-plot.js
julienw: Indeed, thanks. I moved back and forth with different styles.

resources/tentative/charts/observable-plot.js
julienw: Now I'm confused too 🤔 I'll look at it closer.

resources/tentative/charts/observable-plot.js
julienw: This specifically means that the next line is about the bars themselves, as opposed to the following lines.

resources/tentative/charts/observable-plot.js
julienw: Well, _I_ don't find this self-explanatory as somebody who's not used to this much. I'd prefer to keep the comments if that's OK. I can expand them a bit to make them more useful.

resources/tentative/charts/observable-plot.js
julienw: I don't find that "easy" myself :D 
but otherwise yeah, I wanted to keep it for the sake of discussing about including it, and I'm happy to remove it if we agree that we don't need it.

resources/tentative/charts/index.html
julienw: Also note that I followed the same structure than in `editors`.

resources/tentative/charts/observable-plot.js
julienw: Oh yes I know: the 7th item is "Others", look at the push operation in line 71.

I picked 6 in an arbitrary way. I can extract it to a constant. What do you mean by "across test runs", do you mean we could vary it in subsequent runs? Why not...

julienw: But TBH I don't see what varying this value would bring to the benchmark compared to just keeping it at a fixed value.

rniwa: Wouldn't varying the number force more DOM updates?

julienw: If the initial rendering takes n milliseconds, I believe that rendering it again will take n milliseconds again, using the same operations. So a slower browser would still be slower when rendering again. And this won't test other parts of the browser engine.

But I'm in no way a browser engine expert, so I may be off here. If we do what you suggest, would that change the performance characteristics in some way (maybe running the JS code in a compiled mode?)? I'm curious now.

resources/tentative/charts/observable-plot.js
rniwa: I'd strongly prefer not having any "what" comments as is the case with the rest of WebKit codebase.

julienw: I removed all the comments now.

resources/tentative/charts/observable-plot.js
julienw: I checked locally on Firefox: if I bump this value from 6 to 20, the "prepare" step doesn't change much, but "stacked chart" step increases by about 33%. But 

Would that be OK to land this as is, but that I work on a separate PR on making this value configurable, and make it vary in the steps, so that we could look at this separately?

julienw: Landing this would also make it possible to look at the other PR more easily.

rniwa: Sounds reasonable.

rniwa (approved):

flashdesignory (approved):

PR #145, opened 2023-04-17
Todomvc svelte 02 - Optional

flashdesignory:
The current Svelte version feels like a good representation for a typical Svelte app, but it doesn't take in consideration how a Svelte app performs when introducing more components. 

This would be comparable to other frameworks' todoMVC implementations that use multiple components instead of just one file. 

I marked this update optional, since there's also a good argument to keep the current version as is. 
On the other hand, this update could be a good indicator and opportunity to optimize for any performance drop.

Score:
https://gist.github.com/flashdesignory/72c8ed2061852b0b5a41058dabd77ed4

@kara


resources/todomvc/architecture-examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,51 @@
...
+        <input class="toggle" type="checkbox" on:change={toggleItem} checked={item.completed} />
+        <!-- svelte-ignore a11y-label-has-associated-control -->
+        <label on:dblclick={startEdit}>{item.description}</label>
+        <button on:click={removeItem} class="destroy" />
+    </div>
+
+    {#if editing === index}
+        <div class="input-container">
+            <!-- svelte-ignore a11y-autofocus -->
+            <input value={item.description} id="edit-todo-input" class="edit" on:keydown={handleEdit} on:blur={updateItem} use:focusInput />
```
julienw, line 47: maybe we can use `bind:value` on `item.description` and avoid all the churn around the `blur` event. The drawback is that the item would be changed at each keystroke, so I'm not sure this would work. Also I'm not sure how to handle the `Escape` key.

resources/todomvc/architecture-examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,51 @@
...
+        dispatch('toggleItem', { index, checked: event.target.checked });
+    }
+
+    async function focusInput(element) {
+		await tick();
+		element.focus();
+	}
+
+    export let item;
+    export let editing;
```
julienw, line 32: It's not clear to me why the `editing` state is handled outside of this component. I would have used a boolean handled at the component level, and not exposed it to the parent. The `index` wouldn't be necessary.

Note also that with the current implementation all items are rerendered if the "editing" state changes for one (because `editing` changes for all). This sounds suboptimal.

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -81,40 +85,10 @@
...
-
-                    {#if editing === index}
-                        <div class="input-container">
-                            <!-- svelte-ignore a11y-autofocus -->
-                            <input value={item.description} data-index={index} id="edit-todo-input" class="edit" on:keydown={handleEdit} on:blur={updateItem} use:focusInput />
-                            <label class="visually-hidden" for="edit-todo-input">Edit Todo Input</label>
-                        </div>
-                    {/if}
-                </li>
+                <Item item={item} editing={editing} index={index} on:removeItem={removeItem} on:updateItem={updateItem} on:startEdit={startEdit} on:handleEdit={handleEdit} on:toggleItem={toggleItem} />
```
julienw, line 88: instead of passing the `index` to the component that doesn't need it (see below), you can use arrow functions as handler that would take the `index` directly here.

julienw (commented):
I find this interesting. I left a few comments.

I believe the "editing" state needs to be done differently, and it would be good to use some svelte magic binding as I'm suggesting below.
Tell me what you think :-)

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -12,54 +15,59 @@
...
-                description: text,
-                completed: false,
-            });
-            event.target.value = "";
-        }
+        items = items.concat({
+            id: uuid(),
+            description: event.detail.text,
+            completed: false,
+        });
```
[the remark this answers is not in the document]
flashdesignory: reading the link you provided, we could also do:

```
items = [...items, {
            id: uuid(),
            description: event.detail.text,
            completed: false,
}];
```

resources/todomvc/architecture-examples/svelte/src/Header.svelte
```diff
@@ -0,0 +1,19 @@
...
+            text: e.target.value
+        });
+        e.target.value = "";
+    }
+</script>
+
+<header class="header">
+    <h1>todos</h1>
+    <!-- svelte-ignore a11y-autofocus -->
+    <input class="new-todo" on:keydown={addItem} placeholder="What needs to be done?" autofocus />
```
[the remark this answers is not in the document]
flashdesignory: Forms do simplify things for sure and remove some boilerplate code. 
We didn't do this with the other implementations so far and the [official svelte example](https://github.com/sveltejs/svelte-todomvc/blob/master/src/TodoMVC.svelte#L90) doesn't do it either. 

This is something I'd like to revisit "globally", with some other a11y related issues we found throughout the update process of all todo apps.

resources/todomvc/architecture-examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,51 @@
...
+    <div class="view">
+        <input class="toggle" type="checkbox" on:change={toggleItem} checked={item.completed} />
+        <!-- svelte-ignore a11y-label-has-associated-control -->
+        <label on:dblclick={startEdit}>{item.description}</label>
+        <button on:click={removeItem} class="destroy" />
+    </div>
+
+    {#if editing === index}
+        <div class="input-container">
+            <!-- svelte-ignore a11y-autofocus -->
```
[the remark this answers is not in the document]
flashdesignory: not anymore - will remove 😄

resources/todomvc/architecture-examples/svelte/src/Item.svelte
flashdesignory: The parent component evaluates the key and decides what to do with the escape key.

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -12,54 +15,59 @@
...
     }
 
-    function removeItem(index) {
+    function removeItem(e) {
+        const index = e.detail.index;
         items = items.slice(0, index).concat(items.slice(index + 1));
```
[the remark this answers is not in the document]
flashdesignory: Just from a testing perspective, I couldn't determine if the change impacts performance. 
Maybe the list is too small to measure.

resources/todomvc/architecture-examples/svelte/src/Item.svelte
julienw: I mean, I'm not sure how to handle the Escape key with the change I'm proposing :-)

resources/todomvc/architecture-examples/svelte/src/App.svelte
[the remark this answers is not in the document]
julienw: Yes but this is very similar to .concat, iterating on the old array and creating a new array. While `push` is not used a lot in current benchmarks as far as I know.

resources/todomvc/architecture-examples/svelte/src/Header.svelte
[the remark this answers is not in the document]
julienw: This example is pretty old, were bidirectional binding existing at that time?

resources/todomvc/architecture-examples/svelte/src/App.svelte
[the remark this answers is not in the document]
flashdesignory: got it - sounds good. Score wise, I don't really see a difference, but it's good to have.

resources/todomvc/architecture-examples/svelte/src/App.svelte
flashdesignory: removed index param

resources/todomvc/architecture-examples/svelte/src/Header.svelte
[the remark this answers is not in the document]
flashdesignory: I'm not sure, but I still feel that the introduction of forms would be beneficial for all todo apps and could be handled as a separate effort.

resources/todomvc/architecture-examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,55 @@
...
+
+    function updateItem(event) {
+        if (!editing) return;
+        dispatch('updateItem', { text: event.target.value });
+        editing = false;
+    }
+
+    function toggleItem(event) {
+        dispatch('toggleItem', { checked: event.target.checked });
+    }
```
julienw, line 31: I believe that the toggle operation could be done directly here.

resources/todomvc/architecture-examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,55 @@
...
+            event.target.blur();
+        else if (event.key === "Escape") 
+            editing = false;
+    }
+
+    function updateItem(event) {
+        if (!editing) return;
+        dispatch('updateItem', { text: event.target.value });
+        editing = false;
+    }
```
julienw, line 27: I think that the update operation could be done directly here. If the value is emtpy we could call `removeItem`.

julienw (commented):
Reading the code again, I thought of some more simplification. Tell me what you think!
After that (and if this works as expected) I think we should be good to go.
Sorry for the additional delay.

resources/todomvc/architecture-examples/svelte/src/Item.svelte
flashdesignory: @julienw - do you mean an inline function like this in the item component?
`<input class="toggle" type="checkbox" on:change={(e) => item.completed = e.target.checked } checked={item.completed} />`

I tried that and it didn't work, maybe since I don't re-assign the item data in the parent?

flashdesignory: yeah, seems like we'd need to force an update in the parent and that would make the toggle update mood...
thanks for the above 😄

julienw:
I'm fine with the current code. @bgrins what do you think?

resources/todomvc/architecture-examples/svelte/src/Item.svelte
bgrins: Probably just leave it as-is then? I assume the "correct" todomvc behavior is to only change on blur / change and not on keypress

bgrins:
> I'm fine with the current code. @bgrins what do you think?

I'll defer to your review. Generally speaking, splitting this up into the component structure proposed here makes sense to me as I assume it captures some typical overhead and API usage (since presumably most Svelte applications are not implemented in a single component). So as long as we have enough collective confidence that the new implementation is following conventional Svelte usage I think it would be good to land.

bgrins (approved):

rniwa:
Yeah, it's hard to say which version is more representative. Would it be possible to reach out to Svelte community and see what they say / prefer?

flashdesignory:
> Yeah, it's hard to say which version is more representative. Would it be possible to reach out to Svelte community and see what they say / prefer?

I'll reach out and try to find out.

flashdesignory:
Simon from the Svelte team at Vercel was kind enough to provide feedback. 

Screenshot 2023-04-28 at 11 49 14 AM

My motivations to split the app are below - and I think they are still valid points:
1.) show off what the framework can do
2.) be representative of how a larger app most likely will look like.

I did address all items besides the `autofocus` (this didn't work for me, because we are adding the input conditionally and the state has to reflect first).

rniwa (approved):

resources/todomvc/architecture-examples/svelte/src/App.svelte
```diff
@@ -81,40 +56,10 @@
...
-
-                    {#if editing === index}
-                        <div class="input-container">
-                            <!-- svelte-ignore a11y-autofocus -->
-                            <input value={item.description} data-index={index} id="edit-todo-input" class="edit" on:keydown={handleEdit} on:blur={updateItem} use:focusInput />
-                            <label class="visually-hidden" for="edit-todo-input">Edit Todo Input</label>
-                        </div>
-                    {/if}
-                </li>
+                <Item bind:item editing={editing} on:removeItem={() => removeItem(index)} />
```
julienw, line 59: :+1:

resources/todomvc/architecture-examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,56 @@
...
+        } else {
+          removeItem();
+        }
+        editing = false;
+    }
+
+    async function focusInput(element) {
+		await tick();
+		element.focus();
+	}
```
julienw, line 39: nit: indentation

julienw (approved):
This looks pretty good to me, thanks for the changes and for checking with a community member!

PR #146, opened 2023-04-19
Add detailed JSON downloads

camillobruni:
For the upcoming comparison view we will need the separate data points which are not available in the existing JSON format.
- Add `Download classic JSON` button
- `Download JSON` and `Copy JSON` now provide a serialized version of the new metrics
-  Make sure there are not loops in the metrics structure to JSON.stringify them directly


resources/benchmark-runner.mjs
```diff
@@ -289,9 +287,16 @@ export class BenchmarkRunner {
             await this._client.didRunSuites(this._measuredValues);
         }
     }
+
+    getIterationTotalMetric(i) {
```
rniwa, line 291: WebKit's naming convention is not to use "get" prefix in these cases:
https://webkit.org/code-style-guidelines/#names-setter-getter

So this can be just iterationTotalMetric(i)

resources/main.mjs
```diff
@@ -260,8 +268,10 @@ class MainBenchmarkClient {
         this._showSection("#details");
     }
 
-    _getFormattedJSONResult() {
+    _getFormattedJSONResult(modern = false) {
```
rniwa, line 271: Can we use a dictionary instead so that the call site would be `_formattedJSONResult({modern: true});` instead of `_getFormattedJSONResult(true);`

rniwa (approved):

camillobruni:
Added `geomeanToScore` helper function to make sure we get the same values for old-style _measuredValues and new-style metrics.

rniwa:
Still seeing a massive discrepancy between JSON & summary:
Screenshot 2023-05-15 at 3 58 46 PM

camillobruni:
I think you're looking a the "classic" JSON, no? In which case the "total" is the total of an iteration, not all iterations.

Classic:
Screenshot 2023-05-16 at 09 51 36

New:
Screenshot 2023-05-16 at 09 53 06

PR #149, opened 2023-05-02
Add a workload with Chart.js

julienw:
This PR adds a new workloads based on Chart.js. As a reminder Observable Plot (#144) is d3-based, and therefore SVG-based, but Chart.js is canvas-based, so these 2 workloads show different performance characteristics.

According to our [research document](https://docs.google.com/document/d/1-KGxek_9LKOmIVVIZpA9YzR0fJoXKSJUrJ0P2h7GKIY/edit), d3 and chart.js are the 2 most used charting libraries. That's why there's an interest to include them in the benchmark, either as a replacement or an addition to the stockcharts-based workload.

On top of the library, this workload also features a different graph than in #144, that involves some trigonometry calculation to compute the distance between two GPS coordinates. So this could be interesting for us.

~~Finally I'd like to highlight that the last step in the benchmark (showing a tooltip) isn't captured by the current synchronous runner, because the work happens in a rAF callback that goes off after the end of the async step (at least on my computer). I hope that the work on the runner will help with that.~~ Following a suggestion from Ryosuke, now this patch mocks rAF with a setTimeout(0) call.

[live demo](https://chartjs--speedometer-preview.netlify.app/?suite=Charts-chartjs)

image


resources/tentative/charts/chartjs.html
```diff
@@ -0,0 +1,19 @@
...
+  <body>
+    <div id="app">
+      <button id='run-all' type='button'>Run all scenario (development only)</button>
+      <button id="prepare" type="button">Prepare data</button>
+      <button id="add-scatter-chart-button" type="button">Add a scatter chart</button>
+      <button id="open-tooltip" type="button">Open tooltips</button>
+      <button id="reset" type="button">Reset</button>
+    </div>
+    <canvas id="chart"></canvas>
+    <script type="module" src="/chartjs.js"></script>
```
rniwa, line 17: Can we use a relative path instead?

rniwa:
> Please look at the last commit only, as this PR is based on #144.
> 
> This PR adds a new workloads based on Chart.js. As a reminder Observable Plot (#144) is d3-based, and therefore SVG-based, but Chart.js is canvas-based, so these 2 workloads show different performance characteristics.
> 
> According to our [research document](https://docs.google.com/document/d/1-KGxek_9LKOmIVVIZpA9YzR0fJoXKSJUrJ0P2h7GKIY/edit), d3 and chart.js are the 2 most used charting libraries. That's why there's an interest to include them in the benchmark, either as a replacement or an addition to the stockcharts-based workload.
> 
> On top of the library, this workload also features a different graph than in #144, that involves some trigonometry calculation to compute the distance between two GPS coordinates. So this could be interesting for us.
> 
> Finally I'd like to highlight that the last step in the benchmark (showing a tooltip) isn't captured by the current synchronous runner, because the work happens in a rAF callback that goes off after the end of the async step (at least on my computer). I hope that the work on the runner will help with that.

Can we manually force-render that step?

julienw:
> > Finally I'd like to highlight that the last step in the benchmark (showing a tooltip) isn't captured by the current synchronous runner, because the work happens in a rAF callback that goes off after the end of the async step (at least on my computer). I hope that the work on the runner will help with that.
> 
> Can we manually force-render that step?

This happens because all events are throttled in https://github.com/chartjs/Chart.js/blob/83ba8fd77689645d0f915a6c8b4881b882be5276/src/platform/platform.dom.js#L238

This is where rAF is called in the `throttled` function:
https://github.com/chartjs/Chart.js/blob/83ba8fd77689645d0f915a6c8b4881b882be5276/src/helpers/helpers.extras.ts#L38

We can't mock it easily (even if we wanted) because it's kept in a local variable: https://github.com/chartjs/Chart.js/blob/83ba8fd77689645d0f915a6c8b4881b882be5276/src/helpers/helpers.extras.ts#L13-L20

And because all events are throttled (even click) we can't try to use another one sadly.

So I don't have a good idea to make it work. Do you have one?

resources/tentative/charts/chartjs.html
julienw: The Vite bundler makes it relative in the dist files, so I think we're good. Please tell me if I'm missing something :-)

julienw: Ah, this is because we specified `base: "./"` in the configuration for vite. I just copy pasted the configuration from the `editors` workload without thinking about it too much, but this looks like the good thing to do!

rniwa:
> > > Finally I'd like to highlight that the last step in the benchmark (showing a tooltip) isn't captured by the current synchronous runner, because the work happens in a rAF callback that goes off after the end of the async step (at least on my computer). I hope that the work on the runner will help with that.
> > 
> > 
> > Can we manually force-render that step?
> 
> This happens because all events are throttled in https://github.com/chartjs/Chart.js/blob/83ba8fd77689645d0f915a6c8b4881b882be5276/src/platform/platform.dom.js#L238
> 
> This is where rAF is called in the `throttled` function: https://github.com/chartjs/Chart.js/blob/83ba8fd77689645d0f915a6c8b4881b882be5276/src/helpers/helpers.extras.ts#L38
> 
> We can't mock it easily (even if we wanted) because it's kept in a local variable: https://github.com/chartjs/Chart.js/blob/83ba8fd77689645d0f915a6c8b4881b882be5276/src/helpers/helpers.extras.ts#L13-L20

Why can't we mock window.requestAnimationFrame? It seems like it's always accessing `window.requestAnimationFrame` so replacing that function should work.

rniwa:
Hm... Can we add an inline (classic) script to override requestAnimationFrame at the beginning of the test file? That should do the trick, right?

julienw:
As discussed yesterday, I also added a step to draw using an opaque color in addition to the previous one, as this can go through different paths in browser engines.

resources/tentative/charts/chartjs.html
rniwa: Looks like it's still not using a relative path?

resources/tests.mjs
```diff
@@ -391,6 +391,30 @@ Suites.push({
     ],
 });
 
+Suites.push({
+    name: "Charts-chartjs",
+    url: "tentative/charts/dist/chartjs.html",
+    async prepare(page) {},
+    tests: [
+        new BenchmarkTestStep("Prepare", (page) => {
```
rniwa, line 399: Why isn't this done in prepare function above like other tests?

rniwa:
This workload has a very interesting perf characteristics so it's exciting for it to be a part of Speedometer 3!

resources/tests.mjs
julienw: That's because I thought there is value in having this code measured too. The prepare step isn't measured.
In this case there are some iterations, map creation (with everything related to map) and operations, and some trigonometry.
FYI I did the same thing in observable plot, where there are more looping / spreading operations than here, which I thought were valuable to measure too.
Tell me what you think!

resources/tentative/charts/chartjs.html
julienw: It is in the `dist` files :-) https://github.com/WebKit/Speedometer/pull/149/files#diff-39eaea024fee1f3009903ff9b4f74e887b7c5091a3d0d33d00398da2d440c1ce
As I explained above, `vite` configuration does it at build time.

resources/tentative/charts/dist/observable-plot.html
```diff
@@ -4,7 +4,8 @@
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>Observable Plot</title>
-      <script type="module" crossorigin src="./assets/plot-e9515604.js"></script>
+      <script type="module" crossorigin src="./assets/plot-92055dcb.js"></script>
+      <link rel="modulepreload" crossorigin href="./assets/flights-airports-9a9e6422.js">
```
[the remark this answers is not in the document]
julienw: Unfortunately I can't control the indentation generated by vite. One solution could be to run prettier after the build, but I'm not sure it's worth it. What do you think?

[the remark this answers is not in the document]
julienw: This also comes from the fact that this html file has been configured according to the prettier rule (with 4 spaces) but the chartjs one hasn't (so it's 2 spaces). I believe vite outputs 2 spaces because that's the most common configuration by developers.

I don't think this is so important because it's a generated file. But tell me what you think.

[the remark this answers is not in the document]
rniwa: oh I see. I'm fine with this if it's autogenerated.

resources/tests.mjs
rniwa: ok

rniwa:
Do we have measurements of various browsers with this workload? I'd like to see them to sanity check & also see the variance.

resources/tests.mjs
```diff
@@ -391,6 +391,30 @@ Suites.push({
...
+            page.querySelector("#prepare").click();
+        }),
+        new BenchmarkTestStep("Draw scatter", (page) => {
+            page.querySelector("#add-scatter-chart-button").click();
+        }),
+        // Note: this step needs the asynchronous runner to show up in results properly,
```
rniwa, line 405: This note is not necessary because we're working around that now, right?

rniwa (approved):

resources/tests.mjs
julienw: ah good catch!

julienw:
rebased and handled all comments.

Thanks for the review @rniwa!
What do you think @flashdesignory?

julienw:
I rebased and reformatted after the recent "reformat everything" patch from Camillo, and also added an additional commit to get the generated files in line after this reformatting.

julienw:
I tried again today, and now I still have problems of unreported work (the async work from showing the tooltip) even with the "setTimeout" hack.
I wonder if some of the recent runner patches will help on this

camillobruni:
| Chrome |  |  |  |  |
| -- | -- | -- | -- | -- |
Charts-chartjs/Prepare | 12.16 | ± | 2.50 (20.6%) | ms
Charts-chartjs/Draw scatter | 26.10 | ± | 5.57 (21.3%) | ms
Charts-chartjs/Show tooltip | 12.95 | ± | 2.66 (20.6%) | ms
Charts-chartjs/Draw opaque scatter | 20.44 | ± | 0.95 (4.6%) | ms

| Firefox |  |  |  |  |
| -- | -- | -- | -- | -- |
| Charts-chartjs/Prepare | 9.00 | ± | 2.08 (23.1%) | ms
| Charts-chartjs/Draw scatter | 43.00 | ± | 3.07 (7.1%) | ms
| Charts-chartjs/Show tooltip | 29.20 | ± | 1.50 (5.1%) | ms
| Charts-chartjs/Draw opaque scatter | 34.50 | ± | 1.32 (3.8%) | ms

| Safari |  |  |  |  |
| -- | -- | -- | -- | -- |
| Charts-chartjs/Prepare | 6.40 | ± | 0.84 (13.1%) | ms
| Charts-chartjs/Draw scatter | 69.50 | ± | 76.97 (110.8%) | ms
| Charts-chartjs/Show tooltip | 13.20 | ± | 1.93 (14.6%) | ms
| Charts-chartjs/Draw opaque scatter | 31.60 | ± | 5.36 (16.9%) | ms

rniwa:
Hm... something is not right about 110% variance on Draw scatter in Safari.

camillobruni:
- Reran Safari, numbers look better now
- Just double checked a chrome profile and there is no-non measured work.

julienw:
> 
>     * Just double checked a chrome profile and there is no-non measured work.

Have you run this patch alone, or with your new runner?

On Chrome, with this patch only, I see different results whether I profile or not profile. When I profile all work is indeed accounted for, with very low variance. But when I don't profile, then the variance is high on the "tooltip" step.

camillobruni:
I've just seen a bunch of `console.log` output on the hosted version, is this intended?

resources/tentative/charts/chartjs.js
```diff
@@ -0,0 +1,177 @@
...
+            datasets: [
+                {
+                    data: preparedData.flights,
+                    backgroundColor: opaqueCheckBox.checked ? "rgb(0, 125, 255)" : "rgba(0, 125, 255, .20)",
+                },
+            ],
+        },
+        options: {
+            onHover(...args) {
+                console.log(...args);
```
camillobruni, line 78: Is this on purpose?

camillobruni (approved):
I'm fine with landing and potentially fixing the async tooltip measurments (or skipping that if it doesn't work).

I could repro this on FireFox once, but then it went away ("raf" measurement didn't change anything though)

resources/tentative/charts/chartjs.js
bgrins: I assume not, but we should drop the log statement either way IMO

camillobruni: I guess I'm fine either way :) as folks ship too many log statements anyway :)

julienw: And this is removed :-)
This was useful while debugging the benchmark step.

PR #150, opened 2023-05-04
Tentative Workload: Embed TodoMVC-React in a big DOM with complex CSS.

lpardosixtosMs:
As discussed here is the PR for the workload embedding TodoMVC-React in a big DOM with complex CSS.

Included in this PR:

- Complex UI generator. The generated html is a mail-like UI with a big DOM embedded in the sidebar, and styles using adobe's @spectrum-css library.
- Complex CSS generator, 200 matching rules and 200 not matching rules, with every rule triggered by the addition of a new todo element.
- A copy of the TodoMVC-React benchmark with a few modifications: Added class names to list items and view items to trigger the generated CSS, moved the info footer to inside the react component.

Before diving in the review of the code itself I would like us to agree on:

- [ ] The structure of the generated DOM (`src/react-todomvc/public/index.html`).
- [ ] The structure of the generated CSS (`src/react-todomvc/public/matchingCss.css`, `src/react-todomvc/public/nonMatchingCss.css`).
- [ ] The styling tool we are using for the UI (@spectrum-css).

The CLI is a bit hacky copying files from one place to another so it can be built from end-to-end with a single command. Since it's not clear to me what would be the final organization of the project, I didn't spend more time on it.

The hosted version can be found [here](https://lpardosixtosms.github.io/complex-dom/).

Please look at the [README](https://github.com/lpardosixtosMs/Speedometer/tree/lpardosixtos/ComplexDomStaticHTML/resources/tentative/complex-static-html) in `resources/tentative/complex-static-html` for more information about the code and how it is organized.

Impact Analysis available in [Embedding TodoMVC-React in a complex DOM (impact analysis) v2](https://docs.google.com/document/d/1SFVuYC76RRruheNJWc1W6LFd5BIgXwD0ZqLS_PihQm8/edit?usp=sharing)


resources/tentative/complex-static-html/.gitignore
```diff
@@ -0,0 +1,3 @@
+/dist/*
```
bgrins, line 1: FYI - we check in the dist folder so it can be run directly from a static file server with no build step

bgrins: (and having this included in the PR would make it easier to check locally & get to a decision on the checklist items in the description)

resources/tentative/complex-static-html/src/big-dom-generator/components/tree-area.jsx
```diff
@@ -0,0 +1,69 @@
...
+  const {nodeCount, rando, maxDepth, maxBreadth, childProb, currentDepth} = props;
+  if (currentDepth >= maxDepth) {
+    return (null);
+  }
+  // Choose a random number of children.
+  const numChildren = rando.randRange(1, maxBreadth);
+  const children = [];
+  for (let i = 0; i < numChildren && nodeCount.current < TARGET_SIZE; i++) {
+    children.push(<TreeItem key={i}
+                            nodeCount={nodeCount}
```
rniwa, line 16: Nit: wrong indentation style.
It should be indented by exactly 4 spaces.

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
```diff
@@ -0,0 +1,231 @@
+import { LCG } from 'random-seedable';
+import { SEED, MAX_SELECTOR_LEN } from './params.js';
+
+const rando = new LCG(SEED);
```
rniwa, line 4: Please don't use abbreviations like rando. Spell out the full word: random.

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
```diff
@@ -0,0 +1,231 @@
...
+
+// Returns a classname for the element at depth `depth` and in position `index`.
+const getClassname = (depth, index) => {
+  switch (depth) {
+    case 7:
+      return `.view-${index}`;
+    case 6:
+      return `.li-${index}-0`;
+    case 5:
+      if (index === 0) {
```
rniwa, line 45: Nit: These should be indented by 4 spaces, not 2 spaces,  and there should be no curly braces around single line statements.

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
```diff
@@ -0,0 +1,231 @@
...
+  if (selLen === badSelector) {
+    const wrongDepth = rando.randRange(Math.min(depth+1, 7), 8);
+    const wrongSelector = getClassname(wrongDepth, 0);
+    return selector + wrongSelector + oldCombinator;
+  }
+
+  // Otherwise, recurse.
+  const combinator = chooseCombinator(depth, index);
+  const nextDepth = getNextDepth(combinator, depth);
+  const nextIndex =  getNextIndex(combinator, nextDepth, index);
```
rniwa, line 205: Nit: Two spaces after =.

resources/tentative/complex-static-html/src/big-dom-generator/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { renderToStaticMarkup } from 'react-dom/server';
+import { App } from './app';
+import { genCss } from './gen-css'
+
+import './app.css';
+
+const fs = require('fs');
+
+// Generate css matching and non-matching complex css selectors targeted at the todoMVC items.
```
rniwa, line 10: This comment simply repeats what the code conveys.
I generally find these "what" comments distracting. Remove?

resources/tests.mjs
```diff
@@ -132,6 +132,35 @@ Suites.push({
     ],
 });
 
+Suites.push({
+    name: "TodoMVC-React-Big",
```
rniwa, line 136: "Big" is a vague term. How about "Complex-DOM"?

resources/tentative/complex-static-html/src/react-todomvc/todo/components/footer.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function Footer({ todos, dispatch }) {
+    const { pathname: route } = useLocation();
+
+    const activeTodos = useMemo(() => todos.filter((todo) => !todo.completed), [todos]);
+
+    const removeCompleted = useCallback(() => dispatch({ type: "REMOVE_COMPLETED_ITEMS" }), [dispatch]);
+
+    // prettier-ignore
+    if (todos.length === 0)
```
rniwa, line 13: We typically don't compare against 0 like this. Use !todos.length instead:
https://webkit.org/code-style-guidelines/#zero-comparison

resources/tentative/complex-static-html/src/big-dom-generator/params.js
```diff
@@ -0,0 +1,12 @@
...
+export const SEED = 4212021;
+
+// Constrains for the generated DOM tree.
+export const MAX_DEPTH = 40;
+export const MAX_BREADTH = 16;
+export const CHILD_PROB = 0.7;
+export const TARGET_SIZE = 3000;
+
+// Maximum length of a generated CSS selector.
+export const MAX_SELECTOR_LEN = 7;
```
rniwa, line 11: Instead of adding a comment, give the constant more descriptive name:
e.g. MAX_SELECTOR_LENGTH_TO_GENERATE.

resources/tentative/complex-static-html/src/big-dom-generator/params.js
```diff
@@ -0,0 +1,12 @@
+// Default seed for the random number generator.
```
rniwa, line 1: Instead of adding a comment like this, simply give the constant a more descriptive name.
e.g. DEFAULT_SEED_FOR_RANDOM_NUMBER_GENERATOR.

resources/tentative/complex-static-html/IMPACT_ANALYSIS.md
```diff
@@ -0,0 +1,124 @@
...
+
+On Safari, the performance of the benchmark is impacted by every variation. The DOM size has an impact that is worsen when adding CSS.
+
+### Per-step analysis
+
+The benchmarks impact the different test steps differently for each browser, as we can see in the per-step heatmaps below.
+
+### Chrome
+<p align = "center" background-color="white">
+<img src="impact-analysis-images/step-chrome.png" alt="heatmap-chrome" width="800"/>
```
rniwa, line 103: Please don't include this kind of design doc as a part of the benchmark content.
This kind of stuff should be kept in an external document somewhere instead.

rniwa:
While the general approach seems sound, I'd like to see more realistic DOM structure. Namely, we should have more popup menus, etc... hidden in DOM tree. It's quite unrealistic for the entire DOM to be mostly visible as is currently the case.

resources/tentative/complex-static-html/src/big-dom-generator/components/tree-area.jsx
```diff
@@ -0,0 +1,69 @@
...
+                            rando={rando}
+                            numChildren={numChildren}
+                            maxBreadth={maxBreadth}
+                            maxDepth={maxDepth}
+                            childProb={childProb}
+                            currentDepth={currentDepth+1}/>);
+  }
+  nodeCount.current = nodeCount.current + 1;
+  return (
+    <ul className="ui spectrum-TreeView spectrum-TreeView--sizeM">
```
bgrins, line 26: I wonder if we should use `<details open>` instead of `<ul>` here. Potentially then we could have something like a maxVisibleDepth option where, once crossed, the details become closed. This may help with @rniwa's request for more hidden elements and also prevent the nest level from causing a large hidden overflow-x in the sidebar.

rniwa: Do people use details element in real life? I've personally never seen a real website that makes a heavy use of details elements.

bgrins: As a datapoint, I was just looking at a tree I made with `<details>` like https://bgrins.github.io/xbl-analysis/tree/2017-10-01.html and the structure is

```
<details>
  <summary>Item 1</summary>
  <ul>
    <li>
      <details>
        <summary>Item 1.1</summary>
      </details>
     </li>
    <li>
      <details>
        <summary>Item 1.2</summary>
      </details>
     </li>
   </ul>
</details>
```

So actually not too different from what the structure here, except an extra wrapper element, not having the custom chevron element, and likely removing some of the spectrum classes.

rniwa: But is that really a common pattern? I'd rather not use details element unless we know it's frequently used in wild. Websites like nytimes.com, gmail, cnn.com, etc... don't use details element.

lpardosixtosMs: FWIW the pattern I used here is supported by the [treeview component](https://opensource.adobe.com/spectrum-css/treeview.html#foldersfiles) of adobe spectrum-css, so I'd guess it is a common pattern. Also, the library itself uses the `.is-open` class to control the visibility of the children via css. So we can just remove the `.is-open` for elements after `maxVisibleDepth`.

resources/tentative/complex-static-html/.gitignore
lpardosixtosMs: Added them. Also added the path to the files I meant on the checklist.

resources/tentative/complex-static-html/src/big-dom-generator/components/tree-area.jsx
flashdesignory: @lpardosixtosMs - you can opt into our format options that we use from root, by adding exceptions into these files:
https://github.com/WebKit/Speedometer/blob/main/.eslintignore
https://github.com/WebKit/Speedometer/blob/main/.prettierignore

and running `npm run format`
https://github.com/WebKit/Speedometer/blob/main/package.json#L23

This should get you pretty close to the webkit code style.

resources/tentative/complex-static-html/src/big-dom-generator/components/tree-area.jsx
bgrins: Sure, that sounds reasonable. IMO whatever max visible depth which prevents things from being scrolled out of view of the sidebar would be good.

sulekhark:
> While the general approach seems sound, I'd like to see more realistic DOM structure. Namely, we should have more popup menus, etc... hidden in DOM tree. It's quite unrealistic for the entire DOM to be mostly visible as is currently the case.

Yes, we do plan to address this point!

bgrins:
> The structure of the generated CSS (src/react-todomvc/public/matchingCss.css, src/react-todomvc/public/nonMatchingCss.css).

It's not a strong opinion, but I think I'd prefer to generate a single `generated.css` file which puts all matching and non matching in one file since that's more likely how this gets deployed in the wild (i.e. people are more like accidentally including non matching styles in a bundle with matching ones)

bgrins:
> The styling tool we are using for the UI (@spectrum-css).

I'm OK with the choice. I don't know of a clear winner in application-like styling frameworks which work with vanilla HTML / no component library, and I think the output looks relatively good (with some minor nits with spacing / sizing / etc that seem fixable).

bgrins:
> A copy of the TodoMVC-React benchmark with a few modifications: Added class names to list items and view items to trigger the generated CSS, moved the info footer to inside the react component.

For reference, I generated a diff of the changes via `diff -r src/react-todomvc/todo/ ../../todomvc/architecture-examples/react/src/todo`

```
diff -r src/react-todomvc/todo/app.css ../../todomvc/architecture-examples/react/src/todo/app.css
24c24
< }
\ No newline at end of file
---
> }
diff -r src/react-todomvc/todo/components/item.jsx ../../todomvc/architecture-examples/react/src/todo/components/item.jsx
6c6
< export const Item = memo(function Item({ todo, dispatch, index }) {
---
> export const Item = memo(function Item({ todo, dispatch }) {
35,36c35,36
<         <li className={classnames({ completed: todo.completed }) + ` li-${index}-0`} data-testid="todo-item">
<             <div className={`view-${index}`}>
---
>         <li className={classnames({ completed: todo.completed })} data-testid="todo-item">
>             <div className="view">
diff -r src/react-todomvc/todo/components/main.jsx ../../todomvc/architecture-examples/react/src/todo/components/main.jsx
36,37c36,37
<                 {visibleTodos.map((todo, index) => (
<                     <Item todo={todo} key={todo.id} dispatch={dispatch} index={index}/>
---
>                 {visibleTodos.map((todo) => (
>                     <Item todo={todo} key={todo.id} dispatch={dispatch} />
```

I'm happy to see that's so small, as it shows a path forward to bringing the layout to all Todo workloads (and not just React)

bgrins:
Not for this PR, but the way I imagine this would ultimately be structured is:

```
resources/
  todomvc/
    big-dom-generator/
       ...
       generated.css
    ...
```

And then each workload would copy the generated HTML into their index.html, and reference to generated.css (possibly via an npm file reference in the package.json to the big-dom-generator folder similar to how we vendor in the common todomvc CSS).

sulekhark:
> Not for this PR, but the way I imagine this would ultimately be structured is:
> 
> ```
> resources/
>   todomvc/
>     big-dom-generator/
>        ...
>        generated.css
>     ...
> ```
> 
> And then each workload would copy the generated HTML into their index.html, and reference to generated.css (possibly via an npm file reference in the package.json to the big-dom-generator folder similar to how we vendor in the common todomvc CSS).

@bgrins, we would be happy to take this up and try to complete it in time for the June code freeze.

lpardosixtosMs:
> > The structure of the generated CSS (src/react-todomvc/public/matchingCss.css, src/react-todomvc/public/nonMatchingCss.css).
> 
> It's not a strong opinion, but I think I'd prefer to generate a single `generated.css` file which puts all matching and non matching in one file since that's more likely how this gets deployed in the wild (i.e. people are more like accidentally including non matching styles in a bundle with matching ones)

@bgrins I can do that. Is there any comment on the generated css rules themselves?

bgrins:
> > > The structure of the generated CSS (src/react-todomvc/public/matchingCss.css, src/react-todomvc/public/nonMatchingCss.css).
> > 
> > 
> > It's not a strong opinion, but I think I'd prefer to generate a single `generated.css` file which puts all matching and non matching in one file since that's more likely how this gets deployed in the wild (i.e. people are more like accidentally including non matching styles in a bundle with matching ones)
> 
> @bgrins I can do that. Is there any comment on the generated css rules themselves?

I spoke with some of our layout folks and one thing I think could be improved is having some variety in the CSS properties that get set in the matching rules. It currently only sets background-color which means only one of the properties actually gets applied after selector matching and iteration - in Gecko's case we'd hit this [`continue`](https://searchfox.org/mozilla-central/rev/4a5c56f4aca291802ce27320cd9a752dd5dd955e/servo/components/style/properties/cascade.rs#655-657) over and over.

I expect in practice there would be some more non-overlapping properties set by these rules, so perhaps we could find a way to simulate that. I made a quick trial patch to gen-css.js that iterates over a variety of properties and sets a couple for each rule. I don't see it obviously move scores in any browser, but I think something like this (though not necessarily this exactly implementation as there may be more interesting things to set) might be a more robust/realistic test.
```
--- a/resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
+++ b/resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
@@ -206,6 +206,18 @@ const buildNonMatchingSelector = (depth, index, oldCombinator, selLen, badSelect
     return buildNonMatchingSelector(nextDepth, nextIndex, combinator, selLen + 1, badSelector) + selector + oldCombinator;
 };
 
+const props = [
+    "accent-color",
+    "border-bottom-color",
+    "border-color",
+    "border-left-color",
+    "border-right-color",
+    "border-top-color",
+    "column-rule-color",
+    "outline-color",
+    "text-decoration-color",
+];
+
 // Returns a random 200 matching selectors and 200 non-matching selectors targeted at the todoMVC items.
 export const genCss = () => {
     const matchingSelectors = [];
@@ -220,12 +232,27 @@ export const genCss = () => {
     // Create random color styles. Same color, different opacity.
     // TODO: Choose a better color for the todoMVC theme.
     const matchingCssRules = [];
+    let currentPropIndex = 0;
     matchingSelectors.forEach((selector, i) => {
-        matchingCssRules.push(`${selector} { background-color: rgba(140,140,140,${i / 1000}) }`);
+        let propOne = props[currentPropIndex];
+        currentPropIndex = (currentPropIndex + 1) % props.length;
+        let propTwo = props[(currentPropIndex) % props.length];
+        currentPropIndex = (currentPropIndex + 1) % props.length;
+        matchingCssRules.push(`${selector} {
+            ${propOne}: rgba(140,140,140,${i / 1000});
+            ${propTwo}: rgba(140,140,140,${i / 1000});
+        }`);
     });
     const nonMatchingCssRules = [];
     nonMatchingSelectors.forEach((selector, i) => {
-        nonMatchingCssRules.push(`${selector} { background-color: rgba(140,140,140,${i / 1000}) }`);
+        let propOne = props[currentPropIndex];
+        currentPropIndex = (currentPropIndex + 1) % props.length;
+        let propTwo = props[(currentPropIndex) % props.length];
+        currentPropIndex = (currentPropIndex + 1) % props.length;
+        matchingCssRules.push(`${selector} {
+            ${propOne}: rgba(140,140,140,${i / 1000});
+            ${propTwo}: rgba(140,140,140,${i / 1000});
+        }`);
     });
     return { matchingCss: matchingCssRules.join("\n"), nonMatchingCss: nonMatchingCssRules.join("\n") };
 };
```

bgrins:
Beyond that feedback and some UI nitpicks which I sent to you earlier, I'm pretty happy with what's being generated and would like to hear from the other reviewers.

resources/tentative/complex-static-html/src/big-dom-generator/components/tree-area.jsx
lpardosixtosMs: Added the exception and ran formatter, thanks!

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
lpardosixtosMs: Done.

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
lpardosixtosMs: Done.

resources/tentative/complex-static-html/src/react-todomvc/todo/components/footer.jsx
lpardosixtosMs: Done.

resources/tentative/complex-static-html/src/big-dom-generator/params.js
lpardosixtosMs: Done.

resources/tentative/complex-static-html/src/big-dom-generator/components/tree-area.jsx
lpardosixtosMs: Reduced the spacing, font and set a limit of 15 to open objects. See the [updated version](https://lpardosixtosms.github.io/complex-dom/)

lpardosixtosMs:
> While the general approach seems sound, I'd like to see more realistic DOM structure. Namely, we should have more popup menus, etc... hidden in DOM tree. It's quite unrealistic for the entire DOM to be mostly visible as is currently the case.

@rniwa Added popovers to most of the buttons and hid them. You can see them in this [hosted version](https://lpardosixtosms.github.io/complex-dom-popovers/)

bgrins:
@lpardosixtosMs one more request on top of https://github.com/WebKit/Speedometer/pull/150#issuecomment-1540915292 - can we not set the background-color property specifically (or if we do set it make it an alpha variation on the existing color)? The effect it makes with the slightly darker todo area (followed by additional darkness for nested nodes) doesn't match the otherwise realistic looking app.

lpardosixtosMs:
> @lpardosixtosMs one more request on top of [#150 (comment)](https://github.com/WebKit/Speedometer/pull/150#issuecomment-1540915292) - can we not set the background-color property specifically (or if we do set it make it an alpha variation on the existing color)? The effect it makes with the slightly darker todo area (followed by additional darkness for nested nodes) doesn't match the otherwise realistic looking app.

@bgrins Sure, we'll make it match the existing colors, and we'll probably use the css properties that you suggested instead of bakground-color.

resources/tests.mjs
bgrins: I don't disagree (and don't oppose a name change) but my understanding of the longer term plan based on the working group discussions is that this test will be removed and so the name isn't too important. Instead, the complex DOM will ultimately get absorbed into the existing todo workloads (example how this might look in https://github.com/WebKit/Speedometer/pull/150#issuecomment-1535307111). Does that match your thinking?

There are of course sequencing questions about if that should happen now (i.e. close this PR once everyone is happy with the code and make a new one that lands the generator into resources/todomvc and modifies TodoMVC-React) or later (land this and then make a PR that moves the files around and deletes the tentative test).

bgrins (approved):
Thanks for the follow up changes. Overall, I'm happy with this and would like to see it move forward. There are some questions about "how" we land this (i.e. temporarily in a tentative workload or land directly into one or more of the existing todo workloads) - let's discuss options in our upcoming meeting.

flashdesignory:
general question: we usually have a "dev" script, would that make sense here as well?

resources/tentative/complex-static-html/src/react-todomvc/todo/components/item.jsx
```diff
@@ -0,0 +1,51 @@
...
+            else
+                updateItem(id, title);
+
+            setIsWritable(false);
+        },
+        [id, removeItem, updateItem]
+    );
+
+    return (
+        <li className={classnames({ completed: todo.completed }) + ` li-${index}-0`} data-testid="todo-item">
```
flashdesignory, line 35: nit: could we make this more "react'isch"? 

`<li className={classnames({ completed: todo.completed }, `li-${index}-0`)} data-testid="todo-item">`

flashdesignory:
nit: something weird happening with the "all" navigation highlight. 
Do the styles behave differently within this workload? 

Screenshot 2023-05-19 at 6 06 08 PM

resources/tentative/complex-static-html/src/react-todomvc/todo/components/item.jsx
lpardosixtosMs: Done

lpardosixtosMs:
> nit: something weird happening with the "all" navigation highlight. Do the styles behave differently within this workload?
> 
> Screenshot 2023-05-19 at 6 06 08 PM
> These css rules shouldn't interfere with the todo app: Screenshot 2023-05-19 at 6 10 28 PM

I agree. I added a new class "targeted" to every todoMVC list item and their direct children. These are the only elements that should trigger the generated CSS rules, now every generated CSS rules ends with `.targeted` and they should not affect any other elements.

flashdesignory:
@lpardosixtosMs - thanks for making the changes 🥳 

Once the ci/linters succeed, this is good from my side.

flashdesignory (approved):
LGTM

bgrins:
Summarizing our discussion today:

* We should go ahead and merge this PR into tentative. Any final suggestions for content changes can be handled as follow ups.
* Perform some follow-up analysis on the performance impact it has in each browser compared with the existing React test (presumably building on what was done already before the PR was opened)
* There was some back and forth (without a decision) about (a) merging the big DOM into existing Todo workloads vs preserving the small workloads and adding new-new tests and (b) which specific Todo workloads the big DOM should be incorporated into.

sulekhark:
> Summarizing our discussion today:
> 
> * We should go ahead and merge this PR into tentative. Any final suggestions for content changes can be handled as follow ups.
> * Perform some follow-up analysis on the performance impact it has in each browser compared with the existing React test (presumably building on what was done already before the PR was opened)
> * There was some back and forth (without a decision) about (a) merging the big DOM into existing Todo workloads vs preserving the small workloads and adding new-new tests and (b) which specific Todo workloads the big DOM should be incorporated into.

Thanks for summarizing, @bgrins!

rniwa:
Sigh... this payload is triggering "SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds" error in WebKit again.

PR #167, opened 2023-05-19
News Site - Navigation workload

flashdesignory:
News site built with next.js, using its static export feature.

To visualize the test - stepping through breakpoints:
output

@kara


rniwa:
This workload errors out in Safari:
```
TypeError: null is not an object (evaluating 'page.querySelector("#close-toast-link").click')
(anonymous function) — tests.mjs:343
_runTest — benchmark-runner.mjs:288
(anonymous function) — benchmark-runner.mjs:273
```

flashdesignory:
@rniwa  - could you try one more time for me?
I am waiting now for that close button, before running the test.

rniwa:
Hm... I'm still hitting this error midway through the suite after a few iterations:
```
TypeError: null is not an object (evaluating 'page.querySelector("#close-message-link").click')
```

flashdesignory:
> Hm... I'm still hitting this error in Safari midway through the suite:
> 
> ```
> TypeError: null is not an object (evaluating 'page.querySelector("#close-message-link").click')
> ```

ah - I can't repro on my machine.

resources/tests.mjs
```diff
@@ -332,6 +332,53 @@ Suites.push({
...
+    },
+    tests: [
+        new BenchmarkTestStep("NavigateToUS", (page) => {
+            page.querySelector("#close-toast-link").click();
+            page.querySelector("#navbar-dropdown-toggle").click();
+            page.querySelector("#navbar-dropdown-toggle").click();
+            page.querySelector("#navbar-navlist-us-link").click();
+        }),
+        new BenchmarkTestStep("NavigateToWorld", (page) => {
+            page.querySelector("#close-message-link").click();
```
rniwa, line 349: Removing this line makes this test consistently pass in Safari so maybe we can do that for now?
We can always revisit & debug this in more detail but it seems like landing the rest of content into the tentative directory seems quite useful.

flashdesignory: done - thanks for finding!

rniwa (approved):

resources/tests.mjs
```diff
@@ -332,6 +332,52 @@ Suites.push({
...
 
+Suites.push({
+    name: "NewsSite-Next",
+    url: "tentative/newssite/news-next/dist/index.html",
+    async prepare(page) {
+        await page.waitForElement("#close-toast-link");
+    },
+    tests: [
+        new BenchmarkTestStep("NavigateToUS", (page) => {
+            page.querySelector("#close-toast-link").click();
```
bgrins, line 343: We discussed this a bit before, but I have been considering if we should to force a layout in between each of these actions within a step. I believe this would create a more realistic workload for this type of test. While opening and immediately closing the navbar will create React and DOM work there won't be any layout required, while in practice a user discreetly clicking the toggle button twice would cause multiple layouts.

An alternative would be to make each step tiny (with a single action in sync time, then taking the forced layout in async time) but I understand there's a practical problem with noise when steps are so short. So I wonder if having a few big steps that run through a bunch of user actions would be a good compromise that allows them to be heavy enough to be more measurable but also not skipping work that would be happening if they were individual steps.

```
page.querySelector("#close-toast-link").click();
page.layout();
page.querySelector("#navbar-dropdown-toggle").click();
page.layout();
// and so on, possibly containing actions from multiple of the current BenchmarkTestSteps if there is a logical grouping
```

I'm curious for thoughts from others on the idea (cc @rniwa @camillobruni @smaug----)

rniwa: Forcing layouts after each click seems like a reasonable compromise to me since splitting work into multiple pieces can also lead to CPU getting throttled down.

flashdesignory: @bgrins - just to ensure I am making the right change - instead of multiple test steps, we'd have just one and calling the page.layout() function in between each action?

flashdesignory: added page.layout() function, but kept the individual steps for now.

bgrins: Yes, something like that is what I was thinking

bgrins:
I like the shape of this test and think it adds some good coverage for the benchmark, thanks @flashdesignory! Will give  @julienw a chance to make any comments on the code or the generated site itself (https://news-thor-next--speedometer-preview.netlify.app/resources/tentative/newssite/news-next/dist/), although I know there are a number of other items from the repo on his plate so we can discuss as a group on Wednesday how to sequence this if he doesn't have a chance to in the meantime (i.e. if we should just land in tentative and then take any issues as follow ups, or keep it as a PR until then).

resources/tentative/newssite/news-next/src/components/sitemap/sitemap.jsx
```diff
@@ -0,0 +1,40 @@
...
+import { NavLink } from "react-router-dom";
+import { HashLink } from "react-router-hash-link";
+
+export default function Sitemap() {
+    const keys = Object.keys(content);
+    const navItems = keys.reduce(
+        (result, key) => {
+            // priority 0 does not show up in nav
+            /* if (content[key].priority === 0)
+                return result; */
```
julienw, line 11: Should we remove this code?

resources/tentative/newssite/news-next/src/components/toggle/Toggle.jsx
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from "react";
+
+export default function Toggle({ label, onChange, checked }) {
+    const [isSelected, setIsSelected] = useState(false);
+
+    useEffect(() => {
+        setIsSelected(checked);
+    }, [checked]);
```
julienw, line 8: It's not clear to me why we need a local state instead of using directly the prop `checked`.

resources/tentative/newssite/news-next/src/pages/_app.js
```diff
@@ -0,0 +1,28 @@
...
+import "@/styles/message.css";
+import "@/styles/icons-group.css";
+
+import { useEffect, useState } from "react";
+
+function App({ Component, pageProps }) {
+    const [render, setRender] = useState(false);
+    useEffect(() => setRender(true), []);
+    return render ? <Component {...pageProps} /> : null;
+}
```
julienw, line 27: It's not clear to me what the intent of this is. Have you written this or is that generated?

resources/tests.mjs
```diff
@@ -332,6 +332,67 @@ Suites.push({
     ],
 });
 
+Suites.push({
+    name: "NewsSite-Next",
+    url: "tentative/newssite/news-next/dist/index.html",
+    async prepare(page) {
+        await page.waitForElement("#close-toast-link");
+    },
```
julienw, line 340: Can you look whether we can have the initial rendering (including rehydration!) in the tests step instead of in the prepare phase? This would be really good for the benchmark IMO.

julienw (commented):
This looks pretty good to me. I just left a few comments, but it's not clear to me what was generated and what was written by you, so feel free to ignore them..

Also I profiled locally and it doesn't seem to me that we're losing any work in the user timing marks. To be honest I'm a bit surprised because this uses React 18, I was expecting things to be more asynchronous :-)

resources/tentative/newssite/news-next/src/components/navbar/Navbar.jsx
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    function handleClick() {
+        setIsOpen(false);
+    }
+
+    function calculateViewportHeight() {
+        // Since the navbar is supposed to appear below the menu, we can't use position: fixed, height: 100%.
+        // Therefore we are using 100vh for the height. This function fixes the challenge on mobile, where
+        // 100vh might include the addressbar, ect.
```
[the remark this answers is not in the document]
flashdesignory: Yeah, there are newer units, but I think this still represents a larger amount of websites using this approach. 
The only downside is that it's using JavaScript to set the custom property in CSS, but that's really common. 

It's working and I think it's ok to keep and maybe follow up at a later point with a newer approach.

resources/tentative/newssite/news-next/src/components/toggle/Toggle.jsx
flashdesignory: The toggle component uses a "checked" prop, which will let us assign the initial state, or change the toggle from a parent component.

resources/tentative/newssite/news-next/src/pages/_app.js
flashdesignory: ah, yes, this tripped me off too. 
This next.js app uses static files, and this piece of code ensures that next.js doesn't try to render anything "on the server".  With the useEffect, we ensure this code can only run client-side.

resources/tentative/newssite/news-next/src/components/navlist/NavList.jsx
```diff
@@ -0,0 +1,36 @@
...
+            if (content[key].priority === 0)
+                return result;
+
+            // priority 1 shows up in nav list
+            // priority 2 shows up in nav dropdown
+            result[content[key].priority === 1 ? 0 : 1].push(key);
+            return result;
+        },
+        [[], []]
+    );
```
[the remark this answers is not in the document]
flashdesignory: pushed above snippet for review

bgrins (approved):

camillobruni:
Looking good from my side, module sync-step length.
We've talked separately about this, and I think it will be good enough to just add a for-loop with some 4 iterations to bump the sync step time closer to 10ms.

flashdesignory:
> Looking good from my side, module sync-step length.
> We've talked separately about this, and I think it will be good enough to just add a for-loop with some 4 iterations to bump the sync step time closer to 10ms.

Done in the last commit - thanks @camillobruni

Issue #163, opened 2023-05-19
Speedometer doesn't run on a non-secure HTTP server


rniwa:
The current version of Speedometer doesn't run because it relies on `crypto.randomUUID`, which is a HTTPS only feature.

bgrins:
Fixed by #168

PR #168, opened 2023-05-19
[TodoMVC]: replace crypto.randomUUID

flashdesignory:
addresses https://github.com/WebKit/Speedometer/issues/163

@kara


resources/todomvc/architecture-examples/vue/src/components/TodosComponent.vue
```diff
@@ -23,7 +23,9 @@ function uuid() {
     let uuid = "";
     for (let i = 0; i < 32; i++) {
         let random = (Math.random() * 16) | 0;
-        if (i === 8 || i === 12 || i === 16 || i === 20) uuid += "-";
+        // prettier-ignore
```
flashdesignory, line 26: I have to look into the vue built to see why this doesn't format, but for consistency I overwrote it manually.

resources/todomvc/architecture-examples/react-redux/src/reducers/todos.js
```diff
@@ -1,7 +1,17 @@
...
 const initialState = [];
-const uuid = () => crypto.randomUUID();
+function uuid() {
+    let uuid = "";
+    for (let i = 0; i < 32; i++) {
+        const random = (Math.random() * 16) | 0;
+        if (i === 8 || i === 12 || i === 16 || i === 20)
+            uuid += "-";
+
+        uuid += (i === 12 ? 4 : i === 16 ? (random & 3) | 8 : random).toString(16);
```
rniwa, line 11: This line is hard to read. Why not just do this:
```js
let currentNumber = random;
if (i === 12)
    currentNumber = 4;
else if (i === 16)
    currentNumber = 8 | (random & 3);
uuid += currentNumber.toString(16);
```

rniwa (approved):

resources/todomvc/architecture-examples/react-redux/src/reducers/todos.js
flashdesignory: @rniwa  - I can refactor - it was just copy and paste from other implementations. 
Do you want me to make that change to all of them?

rniwa: I think we can use it for ones we're adding newly. I don't think we need to modify vue's.

flashdesignory:
@bgrins @julienw - is this good to merge?

bgrins:
> @bgrins @julienw - is this good to merge?

It's not clear to me that running on a non-https server in production should be a goal - especially if it limits us from using and testing the performance of relevant and widely used APIs. I'm curious if this is meant as a way to be able to host the benchmark on a non-https webserver, or if it's creating an issue in development?

That said, I'll discuss the change with a few folks early next week and see if we're fine to land it regardless.

rniwa:
> > @bgrins @julienw - is this good to merge?
> 
> It's not clear to me that running on a non-https server in production should be a goal - especially if it limits us from using and testing the performance of relevant and widely used APIs. I'm curious if this is meant as a way to be able to host the benchmark on a non-https webserver, or if it's creating an issue in development?

This is blocking us from testing iOS devices in our lab because those machines don't have the public internet access, and therefore would have to run it from a local HTTP server. Landing this change would greatly help us at least in short term since we don't have a setup for running a local HTTPs server even though such is theoretically possible.

bgrins:
> This is blocking us from testing iOS devices in our lab because those machines don't have the public internet access, and therefore would have to run it from a local HTTP server. Landing this change would greatly help us at least in short term since we don't have a setup for running a local HTTPs server even though such is theoretically possible.

I see. OK, we can land this now to unblock you, and if needed take a look at the specific uuid generation function later on. Likewise if we find APIs that are important to test but require a secure context that may require relevant changes in CI systems.

bgrins (approved):

rniwa:
> > This is blocking us from testing iOS devices in our lab because those machines don't have the public internet access, and therefore would have to run it from a local HTTP server. Landing this change would greatly help us at least in short term since we don't have a setup for running a local HTTPs server even though such is theoretically possible.
> 
> I see. OK, we can land this now to unblock you, and if needed take a look at the specific uuid generation function later on. Likewise if we find APIs that are important to test but require a secure context that may require relevant changes in CI systems.

Thank you! Yeah, we can definitely revisit this if more APIs come up that are only available in secure contexts.

PR #173, opened 2023-05-20
Add raf based measurement

rniwa:
This PR adds a new query string parameter requestAnimationFrame, which takes the value of either "timer" or "raf". "timer" is the existing and default measurement method. When it's "raf", we use two consecutive requestAnimationFrame first of which runs the sync step and second of which schedules a 0s timer to signify the end of async step.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -163,6 +163,29 @@ const WarmupSuite = {
     ],
 };
 
+class TestInvoker {
+    constructor(syncCallback, asyncCallback, reportCallback) {
+        this._syncCallback = syncCallback;
+        this._asyncCallback = asyncCallback;
+        this._reportCallback = reportCallback;
+    }
+
+    start() {
+        return new Promise((resolve) => {
+            setTimeout(() => {
+                this._syncCallback();
+                setTimeout(() => {
+                    this._asyncCallback();
+                    requestAnimationFrame(async () => {
+                        await this._reportCallback();
+                        resolve();
+                    });
+                }, 0);
+            }, 0);
+        });
+    }
+}
+
 export class BenchmarkRunner {
     constructor(suites, client) {
         this._suites = suites;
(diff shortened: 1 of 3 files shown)


resources/benchmark-runner.mjs
```diff
@@ -265,56 +311,52 @@ export class BenchmarkRunner {
...
-
-        const syncTime = syncEndTime - syncStartTime;
-
-        performance.mark(asyncStartLabel);
-        const asyncStartTime = performance.now();
-        setTimeout(() => {
+        let syncTime;
+        let asyncStartTime;
+        let asyncTime;
+        const invokerClass = params.measurementMethod === 'raf' ? RAFTestInvoker : TimerTestInvoker;
```
camillobruni, line 326: nice!

resources/benchmark-runner.mjs
```diff
@@ -265,56 +311,52 @@ export class BenchmarkRunner {
...
+        const invoker = new invokerClass(() => {
```
camillobruni, line 327: As on the other PR, maybe named-methods + bind makes this more readable?

camillobruni (approved):

resources/benchmark-runner.mjs
rniwa: What do you mean by that? Could you show me an example?

camillobruni: Along these lines (implying moving the *time values onto the class):

```
   const invoker = new invokerClass(this.measureSync.bind(this), this.measureAsync.bind(this), this._report.bind(this));
    ...
  }
  measureSync(...) {
   ...
  }
  measureAsync(...) {
    ...
  }
  ...
```
Maybe easier: using named arrow functions would be an improvement:

```
   ...
   const syncStart = () => {...};
   const measureAsync = () => {...};
   const report = () => {...};
   const invoker = new invokerClass(measureSync, measureAsync, report);
   
```

resources/benchmark-runner.mjs
```diff
@@ -265,56 +311,52 @@ export class BenchmarkRunner {
...
             performance.measure(`${suite.name}.${test.name}-async`, asyncStartLabel, asyncEndLabel);
-            window.requestAnimationFrame(() => {
-                this._recordTestResults(suite, test, syncTime, asyncTime, height, testDoneCallback);
-            });
-        }, 0);
+        }, () => {
+            return this._recordTestResults(suite, test, syncTime, asyncTime);
+        });
+
+        return invoker.start();
```
camillobruni, line 352: Maybe `await invoker.start();`

resources/benchmark-runner.mjs
smaug----: .bind() makes it harder to read ;) 
But that latter case looks better to me.

smaug---- (approved):
r+ for the actual benchmark-runner. (I'm less familiar with test tests/* code).
Perhaps assign the callbacks to variables and pass those as params, that might improve the readability a bit.

resources/benchmark-runner.mjs
rniwa: Done that.

bgrins (approved):

PR #176, opened 2023-05-23
Stockcharts: Mock requestAnimationFrame to setTimeout so that the run…

### LINKED ISSUE #97: Change the panning and zooming behavior for React-Stockcharts tests
As @squarewave mentioned in https://github.com/WebKit/Speedometer/pull/11#issuecomment-1366851864, it's unrealistic to send i.e. 50 wheel events to paint 50 frames inside of a single request animation frame callback (https://github.com/WebKit/Speedometer/blob/f16557016d2a0bd8f512611c34cc6538afdde121/resources/tests.mjs#L385).

We could consider breaking them up into a handful of steps where we one or a few events. I'm not sure if this would then cause issues on fast desktop hardware where the steps become too short to be effectively measured, but we should take a close look at our options.

I don't know if this is directly contributing to what @camillobruni was seeing with bimodal results in Chrome, but I know there have been some concerns with that as well.

julienw:
…ner can catch asynchronous work

#97

[current main branch](https://speedometer-preview.netlify.app/?suite=React-Stockcharts-SVG)
[with this patch](https://stabilize-react-stockcharts--speedometer-preview.netlify.app/?suite=React-Stockcharts-SVG)

I'm especially interested in the SVG version of the benchmark, and looking especially at the PanTheChart steps. Indeed these steps were the most unstable in the SVG version of the benchmark.

With this change I get very stable results in Firefox, and stable results in Chrome.

What do you think @rniwa @flashdesignory @bgrins?

Firefox Before/After
image
image

Chrome Before/After
image
image


bgrins:
I haven't checked other browsers but this pretty clearly moves work into `React-Stockcharts-SVG.PanTheChart-async` in Firefox (with PR: https://share.firefox.dev/3BTd1ni, without PR: https://share.firefox.dev/42YUSAs). And given that we're using this pattern for other workloads it seems like we should do it here, in addition to other updates to make the SVG workload more realistic like in #97.

bgrins (approved):

camillobruni:
- Maybe thus us the easiest fix for now.
- Ive added an RAF metric in PR #28 maybe this would make StockCharts more reliable on it's own?

rniwa (approved):

julienw:
>     * Ive added an RAF metric in PR [BenchmarkRunner test measurement refactoring #28](https://github.com/WebKit/Speedometer/pull/28) maybe this would make StockCharts more reliable on it's own?

I tested with this oatch. This seems to change things but isn't removing the problem completely:
* When not profiling, I still see the problem in the reported metrics.
* When profiling, the problem doesn't seem to be there anymore.

So there might still be subtle racing problems that I don't fully understand, even with your patch.

julienw:
I rebased on top of the latest main branch and included the results in Firefox and Chrome in the PR description above.

PR #178, opened 2023-05-23
News-site: adjust test steps

flashdesignory:
To avoid to run into the safari limitation, I adjusted the test: 
`SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds.`

Details:
Screenshot 2023-05-23 at 4 18 54 PM

With this change we limit the amount of history api calls.

Snapshot from chrome:

before change:
Screenshot 2023-05-23 at 4 07 46 PM

after change:
Screenshot 2023-05-23 at 4 14 40 PM

### DIFF
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -338,76 +338,64 @@ Suites.push({
         new BenchmarkTestStep("NavigateToUS", (page) => {
             page.querySelector("#close-toast-link").click();
             page.layout();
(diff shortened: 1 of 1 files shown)


rniwa (approved):

bgrins (approved):
We may want to revisit the exact numbers here but let's go ahead and get this landed

Issue #179, opened 2023-05-24
Latest Speedometer 3.0 is broken on Safari


rniwa:
I'm getting this error in NewsSite-Next:

```
[Error] TypeError: null is not an object (evaluating 'page.querySelector("#navbar-dropdown-toggle").click')
	(anonymous function) (tests.mjs:372)
	_runTest (benchmark-runner.mjs:291)
	(anonymous function) (benchmark-runner.mjs:276)
```

rniwa:
Fixed in https://github.com/WebKit/Speedometer/pull/180

rniwa:
I think we need to rewrite some of these tests to not use replaceState.

flashdesignory:
that security "feature" is really not ideal for us 😄 

I think it might be the hash routers from react-router-dom that cause the issue. 
We have that module in all the react-based apps, but only need it in the news-site (since that's the only test that actually navigates). 

We don't test the filters (all, active, complete) in the todoMVC apps and another band-aid might be to remove those  filters instead. 

@rniwa since I can't replicate the bug, what happens if you omit a todomvc react based test? 
Does it pass?

@julienw - you're pretty much reviewed all the workload updates, any other ideas besides the potential culprit being the react-router-dom?

bgrins:
It'd be great if someone familiar with WebKit could log out how many times each individual test is calling replaceState to help narrow down the problem and what change would have the most impact.

rniwa:
Fixed by https://github.com/WebKit/Speedometer/pull/182

rniwa:
Tracking this overarching issue as https://bugs.webkit.org/show_bug.cgi?id=257269.

rniwa:
It's broken because of complex-static-html again :(

rniwa:
I landed a WebKit side fix in https://commits.webkit.org/264499@main.

PR #180, opened 2023-05-24
newssite - potential fix for test

flashdesignory:
this might fix this issue: https://github.com/WebKit/Speedometer/issues/179

update:
simplified test and removed toast close action on the homepage.

### DIFF
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -332,12 +332,10 @@ Suites.push({
     name: "NewsSite-Next",
     url: "tentative/newssite/news-next/dist/index.html",
     async prepare(page) {
-        await page.waitForElement("#close-toast-link");
+        await page.waitForElement("#navbar-dropdown-toggle");
     },
     tests: [
         new BenchmarkTestStep("NavigateToUS", (page) => {
-            page.querySelector("#close-toast-link").click();
-            page.layout();
             for (let i = 0; i < 25; i++) {
                 page.querySelector("#navbar-dropdown-toggle").click();
                 page.layout();
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -365,36 +365,6 @@ Suites.push({
             page.querySelector("#navbar-navlist-politics-link").click();
             page.layout();
         }),
-        new BenchmarkTestStep("NavigateToBusiness", (page) => {
-            for (let i = 0; i < 25; i++) {
-                page.querySelector("#navbar-dropdown-toggle").click();
-                page.layout();
-                page.querySelector("#navbar-dropdown-toggle").click();
-                page.layout();
-            }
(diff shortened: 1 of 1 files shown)


rniwa:
This doesn't fix the problem. As far as I can tell, this is some kind of a WebKit bug that somehow ends up loading speedometer's home page in the iframe.

flashdesignory:
> This doesn't fix the problem. As far as I can tell, this is some kind of a WebKit bug that somehow ends up loading speedometer's home page in the iframe.

Ah, that's weird. So it's not that it can't find the element on the page of the newssite, it just loads something totally different?

I can't replicate it on my machine so it's hard to debug. 
Is it on certain versions of safari? any steps to repro?
How does it actually look when running? do you get that error in the console?

rniwa:
> > This doesn't fix the problem. As far as I can tell, this is some kind of a WebKit bug that somehow ends up loading speedometer's home page in the iframe.
> 
> Ah, that's weird. So it's not that it can't find the element on the page of the newssite, it just loads something totally different?

Yeah, we're somehow loading Speedometer's home page after having ran previous steps. So somehow click() triggering a navigation to home page but src stays in the news site so it's even more bizarre.

> I can't replicate it on my machine so it's hard to debug. Is it on certain versions of safari? any steps to repro? How does it actually look when running? do you get that error in the console?

I can reproduce it on both Safari Technology Preview 169 as well as Safari 16.4. It's such a bizarre bug. Somehow we're replacing the contents of iframe with Speedometer's harness page in the middle of tests after having successfully ran previous steps.
Screenshot 2023-05-24 at 12 29 19 AM

rniwa:
Screenshot 2023-05-24 at 12 31 42 AM

rniwa:
Hm... removing failing steps won't be enough. We're now hitting this error instead at the end of TodoMVC-Angular:
```
[Error] SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds
replaceState
i — vue-router.mjs:514
(anonymous function) — vue-router.mjs:483
A — vue-router.mjs:569
841 — index.js:5
d — bootstrap:19
(anonymous function) — chunk loaded:25
(anonymous function) — startup:5
Global Code — startup:5
```

rniwa:
> Is the security error triggered by summing up the api calls from all workloads or per test?

Yeah, it seems like the issue is that across various tests, we end up exceeding 100 requests / 30s :(

flashdesignory:
pushed change - if you want to try it. 

that security error will always be an issue the more workloads we add in that case.

flashdesignory:
> This "fixes" the issue for me. I should debug Safari/WebKit though. This is such a bizarre bug!

awesome - i'll merge it in for now.

PR #181, opened 2023-05-24
Remove Stockcharts-Canvas steps

### LINKED ISSUE #97: Change the panning and zooming behavior for React-Stockcharts tests

julienw:
Fixes #97

This only removes the steps and otherwise leaves the files as-is. I thought that there's value in keeping the (small) code supporting this in the stockcharts codebase, but please tell me if you'd like me to clean this up as well.

### DIFF
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -439,69 +439,6 @@ Suites.push({
 });
 
 Suites.push({
-    name: "React-Stockcharts",
-    url: "tentative/react-stockcharts/build/index.html?type=hybrid",
-    async prepare(page) {
-        await page.waitForElement("#render");
-    },
-    tests: [
-        new BenchmarkTestStep("Render", (page) => {
-            page.getElementById("render").click();
-        }),
-        new BenchmarkTestStep("PanTheChart", (page) => {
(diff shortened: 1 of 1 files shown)


bgrins (approved):
LGTM. I don't think it's worth it at this point to refactor and remove the `type` option inside the workload, since it's still up in the air whether we'll keep the SVG variation (ongoing investigation in #97).

rniwa (approved):

flashdesignory (approved):

PR #182, opened 2023-05-24
Newssite - use home link for load test

flashdesignory:
for testing:

Screenshot 2023-05-24 at 8 14 55 AM


bgrins:
I'm not able to tell if this helps with #179 since I can't reproduce the problem locally anymore after #178. The change itself seems fine but will defer to @rniwa in terms of whether it helps contribute to a fix.

rniwa:
Nice, this change seems to fix the issue at least on my local testing!

rniwa (approved):

PR #183, opened 2023-05-24
Add a meta viewport tag so that the viewport ias always the same size…

### LINKED ISSUE #130: Consider adding a viewport meta tag to the runner
Currently the runner doesn't have a viewport meta tag (this is true for both [index.html](https://github.com/WebKit/Speedometer/blob/main/index.html) and [InteractiveRunner.html](https://github.com/WebKit/Speedometer/blob/main/InteractiveRunner.html). Because of that, the benchmarks don't work the same on browsers on mobile. Indeed the behavior of browsers isn't the same on mobile when no meta viewport tag is present.

I'd be happy to write a patch if there's a consensus.

julienw:
… on all mobile browsers

Fixes #130

The Editor tests still don't show the same thing on Firefox vs Chrome on Android, but this is already a first step.

[deploy preview](https://meta-viewport--speedometer-preview.netlify.app/)
[deploy preview for editor tests](https://meta-viewport--speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap)
[current branch](https://speedometer-preview.netlify.app/)
[current branch for editor tests](https://speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap)

### DIFF
--- a/InteractiveRunner.html
+++ b/InteractiveRunner.html
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html>
     <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+        <meta name="viewport" content="width=850" />
         <title>Speedometer 3.0 Interactive Runner</title>
         <script src="resources/interactive.mjs" type="module"></script>
         <link rel="icon" href="resources/favicon.png" />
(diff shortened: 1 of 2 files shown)


bgrins (approved):
This does seem like an improvement. Nit: typo in the commit message `ias` -> `is`.

flashdesignory (approved):

rniwa (approved):

PR #184, opened 2023-05-24
Rename "big" and "small" to "long" and "short"

julienw:
After discussing with @bgrins and other folks, we found that the term "big" was confusing, because we could think this is related to the font size. To clear up the confusion, I'm renaming "big" to "long", and "small" to "short".

This PR also moves the "long" content to separate files, relying on vite's capability to inline them to the scripts. This makes it easier to tweak them later. (Note: this is the first commit).

[deploy preview](https://big-to-long-text--speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap)


rniwa (approved):

resources/tentative/editors/assets/README.md
```diff
@@ -0,0 +1,2 @@
+longscript.js: the script comes from Apple
```
julienw, line 1: I'm not sure where this comes from actually, but because I'm changing it in a future PR I didn't spend too much time on finding the source.

bgrins (approved):

flashdesignory (approved):

Issue #153, opened 2023-05-13
Remove non-inclusive language


annevk:
I found this repository after taking a look at https://bugs.webkit.org/show_bug.cgi?id=220104.
```
%  report-non-inclusive-language --verbose
resources/todomvc/architecture-examples/react-redux/dist/app.bundle.js.map: blacklist: 1
resources/todomvc/architecture-examples/vue/dist/js/chunk-vendors.fe2ee873.js.map: whitelist: 1
resources/tentative/editors/bigtext.js: slave: 1
resources/tentative/editors/dist/assets/index.es-8aa334e1.js: slave: 1
Total: 4 (blacklist: 1, whitelist: 1, slave: 2)
```
1. Can these resources be patched or do they also come from upstream somewhere?
2. Is it possible to patch earlier versions of Speedometer with regards to this issue?

flashdesignory:
Hi @annevk - thank you for raising the issue.
We will take a look to see how we can resolve it. 

1. Let us do an audit to see what's the best way to address each case. 
2. Some workloads are new and don't need patching in previous versions of the app. Others are updated and we need to take a look if the same issue existed beforehand.

rniwa:
This has gotten better but we still have one instance of blacklist and whitelist each:
```
Tools/Scripts/report-non-inclusive-language --verbose     
resources/todomvc/architecture-examples/react-redux/dist/app.bundle.js.map: blacklist: 1
resources/todomvc/architecture-examples/vue/dist/js/chunk-vendors.fe2ee873.js.map: whitelist: 1
Total: 2 (blacklist: 1, whitelist: 1, slave: 0)
```

rniwa:
Filed https://github.com/vuejs/core/issues/8416 for Vue.

rniwa:
Hm... looks like React-Redux's usage of blacklist comes from https://github.com/mridgway/hoist-non-react-statics but the package has gotten rid of the non-inclusive language 3 years ago. It's just that React-Redux is using an old version of the package which still contains non-inclusive language. I wonder if we can force React-Redux example to use a newer version of https://github.com/mridgway/hoist-non-react-statics...

flashdesignory:
@annevk - we just merged the last pr to remove non-inclusive language (https://github.com/WebKit/Speedometer/pull/286).

Are we good to close this issue?

rniwa:
Still not quote gone:
```
% ../webkit/Tools/Scripts/report-non-inclusive-language --verbose
[...]

Issue #157, opened 2023-05-18
Investigate noise in Editor tests


bgrins:
We're looking into ways to make the editor tests less noisy in order to move them out of `tentative`. Areas to investigate:

* Squeeze more work into steps. For example - set the text to small, unhighlight, set the text back to big. Right now the actions are fairly limited because when adding an individual step for each the steps were very short.
* Make the input text bigger. In particular the "big" texts aren't actually that big. 
* Understand CodeMirror/TipTap highlighting, is it rAF based? If so, investigate rAF as setTimeout 0 like we did with the charting tests.
* @rniwa had mentioned that Chrome may be doing less layout work, and this issue can be a place to document any further details
* Adding a meta viewport tag (at least on the workload itself) to ensure that the same text gets rendered in each browser (#130)

julienw:
Today I tried to look at the font-size difference between chrome and Firefox on Android, and it looks like this happens even with normal text. Chrome is making the text bigger, presumably to make it more readable, but I don't find any reference or pointer about where or why this is implemented like this.

Only when using `width=device-width` in the viewport meta tag there's no difference, but with the current `width=850` there is one. Currently we don't use `initial-scale=1`. If we use it we get a different behavior, but Chrome and Firefox are still different.

I pushed some examples in https://julienw.github.io/font-size-mobile/ and it would be good if some Google folks (ping @flashdesignory :D) could find out where this Chrome behavior comes from.

If we can't fix this cross-browser issue with our current `width=850` value, I think that our next step would be to use `width=device-width` and resize the iframe (for example with `max-width: 100%`?) so that on one specific device we have the same behavior in all browsers. But then the devices won't be comparable with each other (maybe that's already the case?).

julienw:
Found: https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust

bgrins:
Thanks for the improvements here @julienw. Closing in lieu of #214

PR #185, opened 2023-05-24
Inclusive language cleanup - part 1

flashdesignory:
Partial update regarding issue: [Remove non-inclusive language](https://github.com/WebKit/Speedometer/issues/153).

this update replaces the word `"slave"` with `"[original text removed]"`

@kara


bgrins:
Actually, @julienw is working on updating the text to be longer as part of #157 and I believe is using different source material. Julien, can you confirm?

julienw:
> Actually, @julienw is working on updating the text to be longer as part of #157 and I believe is using different source material. Julien, can you confirm?

yep! a PR is on its way.
It's a french 🇫🇷 text so I doubt it will trigger the inclusive check, however I didn't check manually about french terms that could be non-inclusive.

flashdesignory:
thanks @bgrins @julienw - i'll close this one. 

@julienw - we should maybe still check at some point though 😄

PR #186, opened 2023-05-24
[editor] Use longer text in the benchmarks

julienw:
[current version](https://speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap)
[deploy preview](https://editor-bigger-text--speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap)

CodeMirror is still not super great but TipTap improved greatly.

Firefox v115 nightly Before/After:
image
image

Chrome v115 dev Before/After:
image
image

I think Safari doesn't have a great performance for Tiptap with this longer text, hopefully that will still be OK to include in this benchmark.


bgrins:
I think the current Safari performance with the TipTap test is known & seems to be capturing a real world issue (https://github.com/WebKit/Speedometer/pull/81#discussion_r1117989341)

resources/tentative/editors/assets/longtext.html
```diff
@@ -1,88 +1,3060 @@
...
-    Chief among these motives was the overwhelming idea of the great whale himself. Such a portentous and mysterious monster roused all my curiosity. Then the wild and distant seas where he rolled his island bulk; the undeliverable, nameless perils
-    of the whale; these, with all the attending marvels of a thousand Patagonian sights and sounds, helped to sway me to my wish. With other men, perhaps, such things would not have been inducements; but as for me, I am tormented with an everlasting
-    itch for things remote. I love to sail forbidden seas, and land on barbarous coasts. Not ignoring what is good, I am quick to perceive a horror, and could still be social with it—would they let me—since it is but well to be on friendly terms with
-    all the inmates of the place one lodges in.
-</p>
-<p>
-    By reason of these things, then, the whaling voyage was welcome; the great flood-gates of the wonder-world swung open, and in the wild conceits that swayed me to my purpose, two and two there floated into my inmost soul, endless processions of
-    the whale, and, mid most of them all, one grand hooded phantom, like a snow hill in the air.
-</p>
+<!DOCTYPE html>
```
bgrins, line 1: It's interesting that this works - I guess the editor is stripping HTML passed in which isn't relevant to the doc? I doubt this is typical usage though, so unless if you have a reason to think otherwise I would replace this only with the contents of the <body> element and drop all the HTML boilerplate.

bgrins:
This does look like much more stable numbers in TipTap and potentially an improvement in CodeMirror (certainly using a more typical "big" JS file). Hopefully combined with some of the other improvements we can continue to get the noise down for CM. It looks like it's mostly the create step causing the noise - I wonder if we should include a bit more work in there (perhaps setting to small text after creation)

rniwa (approved):

PR #188, opened 2023-05-24
Additional developerMode controls

bgrins:
This also makes non-default params not frozen so they can be changed without a reload. Demo at https://devmode-2--speedometer-preview.netlify.app/?developerMode.

### DIFF
--- a/resources/developer-mode.mjs
+++ b/resources/developer-mode.mjs
@@ -1,4 +1,5 @@
 import { Suites } from "./tests.mjs";
+import { params, defaultParams } from "./params.mjs";
 
 export function createDeveloperModeContainer() {
     const container = document.createElement("div");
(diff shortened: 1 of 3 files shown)


bgrins:
@camillobruni would be curious especially for thoughts on how to handle these mutable params. Perhaps you can think of a better way to do this.

flashdesignory:
oooh this looks nice ✨ !

- could we add a display of the iterations count? I don't know the count when I use the slider

resources/developer-mode.mjs
```diff
@@ -12,12 +13,71 @@ export function createDeveloperModeContainer() {
...
+    range.max = 20;
+    range.value = params.iterationCount;
+
+    range.onchange = () => {
+        params.iterationCount = parseInt(range.value);
+        fixupURL();
+    };
+
+    let label = document.createElement("label");
+    label.append("iterations ", range);
```
bgrins, line 76: Todo - @flashdesignory mentioned adding the actual count alongside the slider

rniwa: Yeah, can we add the iteration count? Right now, I can't tell what number of iterations are picked.

flashdesignory (approved):

resources/developer-mode.mjs
```diff
@@ -83,6 +146,8 @@ export function createUIForSuites() {
 }
 
 function fixupURL() {
```
camillobruni, line 148: nit: `updateURL` ?

resources/params.mjs
```diff
@@ -73,9 +71,11 @@ class Params {
 }
 
 export const defaultParams = new Params();
+Object.freeze(defaultParams.viewport);
```
camillobruni, line 74: I mostly wanted to guard against accidental modifications of the params objects.
Maybe you can freeze the params object if there are no searchParams (aka the default mode?):

```
    ...
    constructor(searchParams = undefined) {
        if (searchParams)
            this._copyFromParams(searchParams);
        if (!searchParams || searchParams.size == 0) {
            Object.freeze(this.viewport);
            Object.freeze(this);
       }
    }
    ...
```

resources/developer-mode.mjs
```diff
@@ -96,15 +161,28 @@ function fixupURL() {
...
+
+    if (params.iterationCount !== defaultParams.iterationCount)
+        url.searchParams.set("iterationCount", params.iterationCount);
+    else
+        url.searchParams.delete("iterationCount");
+
+    if (params.useWarmupSuite !== defaultParams.useWarmupSuite)
+        url.searchParams.set("useWarmupSuite", params.useWarmupSuite);
+    else
+        url.searchParams.delete("useWarmupSuite");
```
camillobruni, line 182: For a separate PR maybe: This encoding should probably go on the `Params` object directly.

Something like: `url.search = params.encodeSearchParams().toString()`

camillobruni (approved):

resources/params.mjs
bgrins: I thought this wouldn't work initially (i.e. if there are no searchParams at load and then you start wanting to configure). Though with how it's currently built there must by at least one (`developerMode`) to get into this configuration mode. It seems a little coincidental but I don't have a strong preference here so we can update that with a comment.

bgrins: Hm, this change was leading to readonly errors when toggling controls locally. I think I will go ahead with the current change where we explicitly freeze only the defaultParams object and then revisit this along with the URL encoding moving into params object.

PR #190, opened 2023-05-25
Add more markers

camillobruni:
Adding more markers to simplify checking if work is measured
- Add iteration-X for each runner iteration
- Add runner-prepare and runner-finalize to annotate time it takes for adding/removing the iframe
- Use suite-X-prepare


bgrins (approved):
Makes sense to me

rniwa (approved):

flashdesignory (approved):

PR #194, opened 2023-05-25
Developer Mode UI improvements

bgrins:
* Update iterationCount slider label on input instead of change
* Allow ctrl/cmd click on a suite label in addition an input

Preview - https://devmode-3--speedometer-preview.netlify.app/?developerMode

### DIFF
--- a/resources/developer-mode.mjs
+++ b/resources/developer-mode.mjs
@@ -100,8 +100,16 @@ export function createUIForSuites() {
         checkbox.id = suite.name;
         checkbox.type = "checkbox";
         checkbox.checked = !suite.disabled;
-        checkbox.onclick = (event) => {
+        checkbox.onchange = () => {
             suite.disabled = !checkbox.checked;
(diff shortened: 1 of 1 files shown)


flashdesignory:
super nit: instead of having two buttons "select all" "unselect all", would it be possible to use a simple toggle?

bgrins:
> super nit: instead of having two buttons "select all" "unselect all", would it be possible to use a simple toggle?

I did consider something like that when porting from the interactive UI, but it seemed ambiguous what should happen if you press the toggle button when some subset of suites are checked (should it select all or select none?). I'm sure that's solvable but I think the main use cases for this UI are (a) seeing all the suites listed in one place (b) selecting a single suite and (c) a helper UI for people who don't know the don't know the other params / names, which are all covered by the naive thing.

rniwa (approved):

flashdesignory (approved):

PR #196, opened 2023-05-26
Add warmupBeforeSync parameter, which triggers 100ms of infinite loop before each sync step.

rniwa:
This will allow us to evaluate the impact of using rAF vs. timer with regards to CPU throttling.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -327,6 +327,10 @@ export class BenchmarkRunner {
         let asyncStartTime;
         let asyncTime;
         const runSync = () => {
+            if (params.warmupBeforeSync) {
+                const startTime = performance.now();
+                while (performance.now() - startTime < params.warmupBeforeSync); // Infinite loop for the specified ms.
+            }
             performance.mark(startLabel);
             const syncStartTime = performance.now();
             test.run(this._page);
--- a/resources/params.mjs
+++ b/resources/params.mjs
@@ -51,6 +51,13 @@ class Params {
             searchParams.delete("useWarmupSuite");
         }
 
+        if (searchParams.has("warmupBeforeSync")) {
+            this.warmupBeforeSync = parseInt(searchParams.get("warmupBeforeSync"));
+            if (this.warmupBeforeSync < 1)
+                throw new Error(`Invalid warmupBeforeSync param: ${this.warmupBeforeSync}`);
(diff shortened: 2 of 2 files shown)


bgrins:
Interesting idea. Do you have any initial findings with running this locally?

rniwa:
Chrome seems to benefit from this warmup:

Chrome 113.0.5672.126

Timer-based
35.6 ± 1.1
35.6 ± 0.78
36.1 ± 0.82

rAF-based
34.3 ± 0.67
33.9 ± 0.70
34.6 ± 0.76

rAF-based with warmupBeforeSync
35.6 ± 0.47
35.2 ± 0.53
35.8 ± 0.45

Firefox 113.0.2

Timer-based
26.1 ± 0.32
25.9 ± 0.46
25.8 ± 0.32

rAF-based
24.6 ± 0.29
24.7 ± 0.33
24.7 ± 0.64

rAF-based with warmupBeforeSync
24.2 ± 0.82
24.6 ± 0.60
24.5 ± 0.64

Safari (264557@main-1685063151_9e002edc0603)

Timer-based
28.3 ± 1.2
27.9 ± 0.95
28.1 ± 1.1

rAF-based
25.6 ± 0.95
25.2 ± 0.84
25.3 ± 0.89

rAF-based with warmupBeforeSync
25.8 ± 0.91
26.0 ± 0.81
25.6 ± 0.82

bgrins (approved):
This seems worth having for the analysis (not sure if we'll want to ship this by default until we understand it better and figure out what we are doing with the other timing options). If we want to add a developer UI for this I think we can extract the [logic used for the warmup suite](https://github.com/WebKit/Speedometer/blob/07a4618a83cc6164794e73f97b3a83d63c1c1881/resources/developer-mode.mjs#LL45C20-L45C20) (another boolean option) to make it less code to add more like this.

resources/benchmark-runner.mjs
```diff
@@ -327,6 +327,10 @@ export class BenchmarkRunner {
         let asyncStartTime;
         let asyncTime;
         const runSync = () => {
+            if (params.warmupBeforeSync) {
+                const startTime = performance.now();
+                while (performance.now() - startTime < 100); // Infinite loop for 100ms.
```
camillobruni, line 332: Could we make this configurable and use an int for `params.warmupBeforeSync` instead?

PR #198, opened 2023-05-30
[Editors] Set the font-size and text-size-adjust to alleviate differences betwe…

julienw:
…en browsers


bgrins:
Good find, this makes the experience more consistent across engines

camillobruni (approved):

rniwa (approved):

PR #205, opened 2023-06-01
[codemirror] Mock requestAnimationFrame to setTimeout so that we don'…

julienw:
…t lose any work in the async time

I noticed that we were sometimes missing work in the async part of the "long" step for codemirror, both in Chrome and Firefox (both with and without the raf-based runner).
With this hack that also used elsewhere, this doesn't seem to happen anymore.

/cc @flashdesignory

### DIFF
--- a/resources/tentative/editors/codemirror.html
+++ b/resources/tentative/editors/codemirror.html
@@ -126,6 +126,12 @@
             </div>
             <div id="editor"></div>
         </div>
+        <script>
+            // This hack allows to capture the work normally happening in a rAF. We
+            // may be able to remove it if the runner improves.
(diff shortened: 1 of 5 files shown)


bgrins (approved):

flashdesignory (approved):

camillobruni (approved):

julienw:
Feel free to merge as I don't have the write rights yet!
Thanks

bgrins:
> Feel free to merge as I don't have the write rights yet!

@rniwa do you want to review before this is merged?

rniwa (approved):

Issue #201, opened 2023-05-31
Investigate CPU throttling effects


camillobruni:
We should measure the impact on various runner configurations on the final score on various platforms:

- Android (low-end and high-end devices)
- macOS (Apple Silicon and intel)
- linux (probably mostly intel desktop)
- windows (intel)

These are the axis we want to test:
- measurementMode: rAF vs. timer
- warmup Time:  0ms and 50ms
- step delay: 0ms, 50ms, 200ms

TODO:
- [x] Add params for step delay
- [x] Compile complete list of platforms we want to test
- [x] Compile list of test urls with parameters want to check

camillobruni:
Once #202 lands we can use it directly until, I've hosted my branch:

Delayed steps without warmup, `delay in [0, 25, 50, 100, 200]`
- [default](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?startAutomatically)
- [25ms](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=25&startAutomatically)
- [50ms](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=50&startAutomatically)
- [100ms](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=100&startAutomatically)
- [200ms](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=200&startAutomatically)

Delayed steps with `warmup = 50ms` and `delay in [0, 25, 50, 100, 200]`

- [default + 50ms warmup](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?warmupBeforeSync=50&startAutomatically)
- [25ms + 50ms warmup](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=25&warmupBeforeSync=50&startAutomatically)
- [50ms + 50ms warmup](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=50&warmupBeforeSync=50&startAutomatically)
- [100ms + 50ms warmup](https://2023-05-31-delay-params--charming-squirrel-41ddae.netlify.app/?waitBeforeSync=100&warmupBeforeSync=50&startAutomatically)
[...]

PR #206, opened 2023-06-01
News site next updates

flashdesignory:
* Addressed a11y related issues I found
* Used css modules, pulling from https://www.npmjs.com/package/news-site-css?activeTab=versions
* Added raf mock from Julien

The change to use css modules resulted in a lot of changed files, but changes are similar in most of them.

standalone preview:
https://flashdesignory.github.io/news-site-next-static/

@kara


resources/tentative/newssite/news-next/package.json
```diff
@@ -10,9 +10,11 @@
     "serve": "http-server ./dist -p 7002 -c-1 --cors"
   },
   "dependencies": {
+    "classnames": "^2.3.2",
     "eslint": "8.39.0",
     "eslint-config-next": "13.3.4",
     "http-server": "^14.1.1",
+    "news-site-css": "^1.6.5",
```
bgrins, line 17: I wonder if this should be a `file:` reference to a local directory in Speedometer as opposed to distributing through npm? I guess we do this already with todo CSS but I think it could be easier to audit and update styles over time if they live in this project. What do you think?

flashdesignory: I tinkered with that idea and it's an option.

Would the same approach be desirable, if we introduce global css changes to the todomvc apps? 
I'd like to find a consistent approach moving forward for other workloads as well. This seems like a step before moving towards something like a real monorepo, where we'd have better ways of managing dependencies and sharing packages.

The only downside (or change) in process I see with using locally linked packages, is that changes to these than would run through the same approval process as everything else in this repo... but maybe that's a good thing.

flashdesignory: > (i.e. if you intend to use this template as a tool for other projects, more like TodoMVC).

That was my initial thought, but I also don't want to hold up this pr. 
Before I make the change to a local package, @rniwa - do you have any preference?
Long term goal we could also discuss in an issue, where it might be more fitting.

camillobruni:
Mostly for documentation: with this update we  have the timer vs rAF difference in Chrome.

With timers some microtasks are not measured
Screenshot 2023-06-06 at 01 03 38

With rAF we're good:
Screenshot 2023-06-06 at 01 03 51

resources/tentative/newssite/news-next/package.json
julienw: If we do intend to keep it as a npm package, it would be good to use a scoped package @speedometer.

julienw:
> Mostly for documentation: with this update we have the timer vs rAF difference in Chrome.
> 
> With timers some microtasks are not measured

I definitely see this in Firefox too.

resources/tentative/newssite/news-next/package.json
flashdesignory: I like that idea, in that case if it's "official" who should be the owner? Is there a webkit npm account? 
@rniwa  - in general how would you like to manage css packages?

rniwa: Per today's discussion, we prefer having a local package.

camillobruni:
Should we wait until we have the intel investigation on [issue #201](https://github.com/WebKit/Speedometer/issues/201) and switch over to rAF measurements by default?

camillobruni (approved):

bgrins (approved):

resources/tentative/newssite/news-next/package.json
flashdesignory: @rniwa - last push did just that. 
Please let me know if there's anything else to adjust or if we're good to merge 😄

flashdesignory:
> Should we wait until we have the intel investigation on [issue #201](https://github.com/WebKit/Speedometer/issues/201) and switch over to rAF measurements by default?

I'd love to get these changes merged and rather do a follow up pr for this if needed.

rniwa:
Do we have before/after numbers for this change?

flashdesignory:
snapshots:
chrome before:
chrome_before

chrome after:
chrome_after

safari before:
safari_before

safari after:
safari_after

firefox before:
firefox_before

firefox after:
firefox_after

edge before:
edge_before

edge after:
edge_after

rniwa (approved):

PR #208, opened 2023-06-02
Reduce the size of the long text loaded in TipTap to reduce its runni…

julienw:
…ng time

At a first glance this doesn't seem to change the performance characteristics, but feel free to double-check on your respective engines.
The test is now approximately half as long as before, and the variance still look acceptable to me (less than 10%).

[current version](https://speedometer-preview.netlify.app/?suite=Editor-Tiptap)
[this patch](https://reduce-tiptap-long-text--speedometer-preview.netlify.app/?suite=Editor-Tiptap)

Tell me what you think!


resources/tentative/editors/assets/longtext.html
```diff
@@ -1,200 +1,3 @@
-<!DOCTYPE html>
-<html lang="fr">
-    <head>
-        <meta charset="utf-8" />
```
rniwa, line 4: Why are we removing these HTML / CSS?

julienw: This was a comment from @bgrins in the patch that added this workload: these are not displayed at all so it's not useful to add them. But feel free to disagree!

julienw: See https://github.com/WebKit/Speedometer/pull/186#discussion_r1204671489

But the PR was merged before I could fix it.

rniwa (approved):

resources/tentative/editors/assets/longtext.html
rniwa: Ok, thanks for the explanation.

flashdesignory (approved):
LGTM - didn't translate the text and have no idea what it says 🤣 
J'ai appris le français au lycée mais je ne me souviens de rien !

julienw:
> LGTM - didn't translate the text and have no idea what it says rofl J'ai appris le français au lycée mais je ne me souviens de rien !

hehe :) This is the beginning of https://en.wikipedia.org/wiki/In_Search_of_Lost_Time

PR #209, opened 2023-06-05
Freeze params again in non-developerMode 

camillobruni:
This prevents accidental modification in normal runs.

Drive-by-fix:
- Fix comment and run `npm run format`

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -353,7 +353,8 @@ export class BenchmarkRunner {
             if (params.warmupBeforeSync) {
                 const startTime = performance.now();
                 // Infinite loop for the specified ms.
-                while (performance.now() - startTime < params.warmupBeforeSync);
+                while (performance.now() - startTime < params.warmupBeforeSync)
+                    continue;
             }
             performance.mark(startLabel);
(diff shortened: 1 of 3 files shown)


resources/benchmark-runner.mjs
```diff
@@ -352,7 +352,9 @@ export class BenchmarkRunner {
         const runSync = () => {
             if (params.warmupBeforeSync) {
                 const startTime = performance.now();
-                while (performance.now() - startTime < params.warmupBeforeSync); // Infinite loop for the specified ms.
+                // Infinite loop for the specified ms.
+                while (performance.now() - startTime < params.warmupBeforeSync)
+                    continue;
```
rniwa, line 357: Why are we making this change?

rniwa (approved):

bgrins (approved):

resources/benchmark-runner.mjs
camillobruni: Auto-formatting changed it to:

```
while (performance.now() - startTime < params.warmupBeforeSync) // Infinite loop for the specified ms.
    ;
```

So I preferred moving the comment + adding a `continnue` instead of having this weird dangling semicolon.

Issue #204, opened 2023-06-01
Consider using nanoid's non-secure version of id generation


julienw:
I noticed that [nanoid is used inside redux toolkit's `createAsyncThunk`](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/nanoid.ts), so I had a look at its usage, and it looks like [nanoid](https://github.com/ai/nanoid) is used a lot (28M downloads on npm, ported to other languages as well). It would make sense to include this as our id generator in at least one of our workload:

https://github.com/ai/nanoid/blob/950dbdaefef465cc6925c6ec596b3478233a1c2d/non-secure/index.js#L23-L32

Here is our current version that I believe we use in a lot of places: https://github.com/WebKit/Speedometer/blob/e1957e39199e0a6ee361ddd1e61a95b485460394/resources/todomvc/architecture-examples/react/src/todo/reducer.js#L1-L16

What do you think @flashdesignory ?

flashdesignory:
@julienw - nice, totally open to change a workload to use nanoid. 
Did you have a specific workload in mind? React or React-Redux TodoMVC app?

julienw:
We could pick one short-running (preact or svelte), one "normal" running (React, Angular) and possibly one long-running (jQuery)?

flashdesignory:
Let's do Preact and React, if no-one has different opinions. 
Svelte already uses a different version that I'd like to keep.

PR #211, opened 2023-06-05
use nanoid for preact / react todomvc

flashdesignory:
This pr resolves: https://github.com/WebKit/Speedometer/issues/204

For comparison - function to remove:

```
function uuid() {
  let uuid = "";
  for (let i = 0; i < 32; i++) {
      const random = (Math.random() * 16) | 0;
      if (i === 8 || i === 12 || i === 16 || i === 20)
          uuid += "-";

      let currentNumber = random;
      if (i === 12)
          currentNumber = 4;
      else if (i === 16)
          currentNumber = 8 | (random & 3);
      uuid += currentNumber.toString(16);
  }
  return uuid;
}
```

function to replace with:
```
// Borrowed from https://github.com/ai/nanoid/blob/3.0.2/non-secure/index.js
// This alphabet uses `A-Za-z0-9_-` symbols. A genetic algorithm helped
// optimize the gzip compression for this alphabet.
const urlAlphabet =
  'ModuleSymbhasOwnPr-0123456789ABCDEFGHNRVfgctiUvz_KqYTJkLxpZXIjQW'

/**
 *
 * @public
 */
function nanoid(size = 21){
  let id = ''
  // A compact alternative for `for (var i = 0; i < step; i++)`.
  let i = size
  while (i--) {
    // `| 0` is more compact and faster than `Math.floor()`.
    id += urlAlphabet[(Math.random() * 64) | 0]
  }
  return id
}
```
@kara


julienw (approved):
Looks good to me, thanks for the change!

resources/todomvc/architecture-examples/preact/src/app/model.js
```diff
@@ -1,13 +1,21 @@
-function uuid() {
-    let uuid = "";
-    for (let i = 0; i < 32; i++) {
-        let random = (Math.random() * 16) | 0;
-        if (i === 8 || i === 12 || i === 16 || i === 20)
-            uuid += "-";
+// Borrowed from https://github.com/ai/nanoid/blob/3.0.2/non-secure/index.js
```
rniwa, line 1: We should probably include the LICENSE file with this.

resources/todomvc/architecture-examples/preact/src/app/model.js
```diff
@@ -1,13 +1,21 @@
...
+// This alphabet uses `A-Za-z0-9_-` symbols. A genetic algorithm helped
+// optimize the gzip compression for this alphabet.
+const urlAlphabet = "ModuleSymbhasOwnPr-0123456789ABCDEFGHNRVfgctiUvz_KqYTJkLxpZXIjQW";
```
camillobruni, line 4: The alphabet here probably doesn't matter all that much for us, but any reason we don't use the latest version 3.X ?
https://github.com/ai/nanoid/blob/3.3.5/non-secure/index.js

resources/todomvc/architecture-examples/preact/src/app/model.js
flashdesignory: added on top of the file.

resources/todomvc/architecture-examples/preact/src/app/model.js
flashdesignory: updated

rniwa (approved):

PR #214, opened 2023-06-07
Move the editor workload out of tentative

julienw:
I put up this PR to start the discussion about moving the "editor" workload out the tentative directory.
* The variance is fairly low (< 10ms) for both Firefox and Chrome
* The total running time is ~50ms for codemirror and ~150ms for Tiptap
* I think there is some missing work in Chrome, but that it's accounted properly with the `raf` measurement method.

So what do you all think? What else do we need to move this out of tentative?

[Link for editor workloads](https://speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap)
[Link for editor workloads with raf-based measurement](https://speedometer-preview.netlify.app/?suite=Editor-Codemirror,Editor-Tiptap&measurementMethod=raf)

Firefox:
image

Chrome:
image


bgrins (approved):

rniwa:
Hm... I'm still seeing ~10% variance on CodeMirror test case in Firefox & Chrome even with measurementMethod=raf. Is there any way we can reduce the noise level further?

camillobruni:
I did 10 total runs with 10 iterations each and the noise levels from run to run are rather ok:

```
label                    default                default                default
browser                  Safari                 Google Chrome          Firefox
version                  16.5                   114.0.5735.106         114.0
os                       macos 13.4 arm64       macos 13.4 arm64       macos 13.4 arm64
device                   MacBookPro18,2         MacBookPro18,2         MacBookPro18,2
cpu                      Apple M1 Max 10 cores  Apple M1 Max 10 cores  Apple M1 Max 10 cores

timer
Editor-CodeMirror/total  28.87 ± 2.8%           24.45 ± 1.1%           30.30 ± 1.6%
Editor-TipTap/total      724 ± 3.2%             100.82 ± 0.25%         110.64 ± 0.36%
Score                    8.31 ± 2.6%            24.17 ± 0.55%          20.73 ± 0.83%

rAF
Editor-CodeMirror/total  37.20 ± 1.4%           26.98 ± 2.7%           32.40 ± 1.7%
Editor-TipTap/total      767 ± 1.4%             99.48 ± 0.52%          110.67 ± 0.60%
Score                    7.108 ± 0.90%          23.17 ± 1.5%           20.04 ± 0.88%
```

julienw:
On my machine I occasionally see some variance in Chrome, but never in Firefox.
Do you see it a lot @rniwa ?

bgrins:
> I get ~10% noise half the time on timer based measurement but maybe the average error rate might be lower in raf based measurement.

I think we should focus on raf-based measurement at this point, right? Or is there a reason to think it might not ship as the default in 3?

rniwa:
> What do your absolute numbers look like @rniwa? I try to understand if the variance might be huge because the workloads are too fast.

Looks like most of noise comes from the initial iteration:

● | Editor-CodeMirror/Create | 5.37 | ± | 2.15 (40.1%) | ms
-- | -- | -- | -- | -- | --
● | Editor-CodeMirror/Long | 9.47 | ± | 1.55 (16.3%) | ms
● | Editor-CodeMirror/Highlight | 14.10 | ± | 2.44 (17.3%) | ms


Iteration | /Create [ms] | /Long [ms] | /Highlight [ms]
-- | -- | -- | --
0 | 10.20 | 15.30 | 23.70
1 | 2.90 | 9.60 | 12.60
2 | 3.50 | 8.60 | 13.40
3 | 8.80 | 8.10 | 13.00
4 | 2.90 | 7.90 | 12.40
5 | 8.20 | 8.30 | 13.00
6 | 3.00 | 9.80 | 13.40
7 | 3.10 | 9.00 | 14.10
8 | 3.10 | 9.70 | 12.50
9 | 8.00 | 8.40 | 12.90

bgrins:
> Looks like most of noise comes from the initial iteration:

I've seen this with other tests as well. I think Chrome is caching some kind of performance hints across runs. It does seem particularly large here but I don't know there's anything to change with the test in particular. As @camillobruni noticed the time across runs is quite stable, so even if there's a lot of variance across iterations it appears to be getting consistent results.

rniwa:
> > Looks like most of noise comes from the initial iteration:
> 
> I've seen this with other tests as well. I think Chrome is caching some kind of performance hints across runs. It does seem particularly large here but I don't know there's anything to change with the test in particular. As @camillobruni noticed the time across runs is quite stable, so even if there's a lot of variance across iterations it appears to be getting consistent results.

One thing we could consider is to vary the text content in each iteration so that such a cache won't be useful. I'd argue that would be more realistic of real life scenario.

camillobruni:
See my comment on #223. I think for chrome it's expected that the first run is a bit slower since Chrome lazily compiles many functions on first execution only (as opposed to compiling all of them when the script arrives). Once compiled we cache functions and subsequent runs reuse those (relevant real-world use-case when navigating on a page).

This is a trade-off that makes less sense for speedometer, but has positive memory and startup impact on most websites where only a subset of functions are initially used.

rniwa:
> I presume Firefox and Safari compile all functions in the non-measured prepare step?

Safari lazily compiles functions as well. Nevertheless, the analysis makes sense. Thanks for looking into this.

camillobruni:
This is how a typical first execution in Chrome looks like (aka not very nice):

Screenshot 2023-06-09 at 00 29 03

bgrins:
> > > Looks like most of noise comes from the initial iteration:
> > 
> > 
> > I've seen this with other tests as well. I think Chrome is caching some kind of performance hints across runs. It does seem particularly large here but I don't know there's anything to change with the test in particular. As @camillobruni noticed the time across runs is quite stable, so even if there's a lot of variance across iterations it appears to be getting consistent results.
> 
> One thing we could consider is to vary the text content in each iteration so that such a cache won't be useful. I'd argue that would be more realistic of real life scenario.

That's an interesting idea. Right now the long text is just exported as a string from a js file, although we could consider appending a new line of code for each iteration or something. I'm actually not sure how we'd do that though since each iteration re-runs the same steps - have we done anything like this before?

And of course, we can always test and make changes to this if we find something after moving this out of tentative (just like we have been with the todo workloads).

camillobruni (approved):

julienw:
> > > Looks like most of noise comes from the initial iteration:
> > 
> > 
> > I've seen this with other tests as well. I think Chrome is caching some kind of performance hints across runs. It does seem particularly large here but I don't know there's anything to change with the test in particular. As @camillobruni noticed the time across runs is quite stable, so even if there's a lot of variance across iterations it appears to be getting consistent results.
> 
> One thing we could consider is to vary the text content in each iteration so that such a cache won't be useful. I'd argue that would be more realistic of real life scenario.

I tried locally (removed a random amount of lines (between 0 and 15) at the end at each time) but this didn't change much. That was also my expectation because I believe the cache isn't about the actual text content but about the functions.

FWIW I did a full SP3 run on Chrome on my machine, and I see the same thing happening for nearly all workloads. The first run is visibly slower. I think that the workloads where this doesn't happen are the ones where the function compilation is out of the measured time (Vue is an example, I'd argue that all functions are already visited at the initial render time, which isn't measured).

camillobruni:
Yes, so I think from our side this is acceptable. As long as the first run is predictable slower (as it is in this case).
So still thumbs up from our side.

rniwa (approved):

PR #215, opened 2023-06-07
Add Suite tags for complex filtering

camillobruni:
- Each suite has a set of tags
  - "all" applies to all suites
  - "default" applies to suites that are enabled by default (and not tentative)
  - "tentative" applies to suites that are in the tentative directory
- `&tags=` can be used for filtering Suites by tag

### DIFF
--- a/resources/interactive.mjs
+++ b/resources/interactive.mjs
@@ -196,9 +196,14 @@ function createUIForSuites(suites, onStep, onRunSuites) {
 }
 
 function startTest() {
+    let message;
     if (params.suites.length > 0) {
-        if (!Suites.enable(params.suites)) {
-            const message = `Suite "${params.suites}" does not exist. No tests to run.`;
+        if (!Suites.enable(params.suites))
+            message = `Suite "${params.suites}" does not exist. No tests to run.`;
+    } else if (params.tags.length > 0) {
+
+    }
+    if (message) {
             alert(message);
             console.error(
                 message,
@@ -206,7 +211,6 @@ function startTest() {
                 "\nValid values:",
                 Suites.map((each) => each.name)
             );
-        }
     }
 
     const interactiveRunner = new window.BenchmarkRunner(Suites, params.iterationCount);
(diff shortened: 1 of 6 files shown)


resources/tests.mjs
```diff
@@ -3,19 +3,56 @@ import { BenchmarkTestStep } from "./benchmark-runner.mjs";
...
+            suite.disabled = !suite.tags.some((tag) => tags.has(tag));
+        });
+    } else {
+        console.warn("Neither names nor tags provided. Enabling all default suites.");
+        this.forEach((suite) => {
+            suite.disabled = !("default" in suite.tags);
+        });
+        return;
+    }
+    if (Suites.some((suite) => !suite.disabled))
```
rniwa, line 31: Can we flip this condition sot that error case is in an early return if?
That'll make it easier to follow the code path.

rniwa (approved):

resources/interactive.mjs
```diff
@@ -196,18 +196,8 @@ function createUIForSuites(suites, onStep, onRunSuites) {
 }
 
 function startTest() {
-    if (params.suites.length > 0) {
-        if (!Suites.enable(params.suites)) {
```
bgrins, line 200: Eventually I guess we'll want to call `enable` with only the "default" tag if nothing is passed, right?

I actually prefer to keep the behavior how you have it here for now while we're still pre-release and working to get tests out of tentative.

bgrins:
It'll be nice to have developerMode UI to list and select tags but that can be follow up material. And actually getting the UX right for exclusive selection across suites & tags might be non trivial.

bgrins (approved):
Neat

resources/interactive.mjs
camillobruni: yes 👍

resources/tests.mjs
camillobruni: Not sure I fullly follow, did you mean?

```
if (this.every((suite) => suite.disabled) {
  ... throw errors...
 }
```

camillobruni:
I'll add a list of tags to the developerUI in a later PR.

resources/tests.mjs
rniwa: Yes.

camillobruni: DONE

PR #217, opened 2023-06-08
Developer Mode: Support filtering by tags

camillobruni:
- Add Tags filtering buttons
- Update url with tag if a single tag can be used to represent the current selection
- Make sure auto-added tags are first in Tags and suite.tags
 
Screenshot 2023-06-08 at 11 37 40

### DIFF
--- a/resources/developer-mode.mjs
+++ b/resources/developer-mode.mjs
@@ -1,4 +1,4 @@
-import { Suites } from "./tests.mjs";
+import { Suites, Tags } from "./tests.mjs";
 import { params, defaultParams } from "./params.mjs";
 
 export function createDeveloperModeContainer() {
@@ -88,7 +88,6 @@ export function createUIForSuites() {
     const control = document.createElement("nav");
     const ol = document.createElement("ol");
     const checkboxes = [];
-
     const setSuiteEnabled = (suiteIndex, enabled) => {
         Suites[suiteIndex].disabled = !enabled;
         checkboxes[suiteIndex].checked = enabled;
@@ -122,13 +121,8 @@ export function createUIForSuites() {
 
         ol.appendChild(li);
     }
-
-    this.createFilterButtons();
-}
-
-function createFilterButtons() {
     control.appendChild(ol);
-    const buttons = control.appendChild(document.createElement("div"));
+    let buttons = control.appendChild(document.createElement("div"));
     buttons.className = "button-bar";
 
     let button = document.createElement("button");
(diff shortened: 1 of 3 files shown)


resources/developer-mode.mjs
```diff
@@ -147,6 +145,37 @@ export function createUIForSuites() {
     };
     buttons.appendChild(button);
 
+    let i = 0;
+    const kTagsPerLine = 3;
+    for (const tag of Tags) {
+        if (tag === "all")
+            continue;
+        if (i % kTagsPerLine === 0) {
```
rniwa, line 153: I think WebKit's coding style is to use !(i % kTagsPerLine) in these cases:
https://webkit.org/code-style-guidelines/#null-false-and-zero

resources/developer-mode.mjs
```diff
@@ -168,7 +193,24 @@ function updateURL() {
     } else {
         url.searchParams.delete("tags");
         url.searchParams.delete("suite");
-        url.searchParams.set("suites", selectedSuites.join(","));
+        // Try finding common tags that would result in the current suite selection.
+        let commonTags = new Set(selectedSuites[0].tags);
+        for (const suite of Suites) {
+            if (suite.disabled) {
+                // commonTags = commonTags.difference(suite.tags);
+                suite.tags.forEach((tag) => commonTags.delete(tag));
```
rniwa, line 201: Just because this suite is disabled doesn't mean all the tags associated with this suite need to be removed / hidden, right? Those tags may have been associated with other suits that are enabled & appeared earlier in this loop.

camillobruni: Currently we have to do it this way, as the `Suites.enable` method enables any suite that contains any of the provided flags (as opposed to only enabling it if all tags are present). (Up for changing that as well)

rniwa: Hm... I see. That's rather convoluted but I suppose we can address that in a follow up.

resources/developer-mode.mjs
```diff
@@ -147,6 +145,37 @@ export function createUIForSuites() {
...
+        if (tag === "all")
+            continue;
+        if (i % kTagsPerLine === 0) {
+            buttons = control.appendChild(document.createElement("div"));
+            buttons.className = "button-bar";
+        }
+        i++;
+        button = document.createElement("button");
+        button.className = "tag";
+        button.textContent = `#${tag}`;
```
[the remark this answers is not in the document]
bgrins: I prefer it as-is, but I don't care much either way

resources/developer-mode.mjs
```diff
@@ -168,7 +193,24 @@ function updateURL() {
...
         url.searchParams.delete("suite");
-        url.searchParams.set("suites", selectedSuites.join(","));
+        // Try finding common tags that would result in the current suite selection.
+        let commonTags = new Set(selectedSuites[0].tags);
+        for (const suite of Suites) {
+            if (suite.disabled) {
+                // commonTags = commonTags.difference(suite.tags);
+                suite.tags.forEach((tag) => commonTags.delete(tag));
+            } else {
+                // commonTags = commonTags.intersection(suite.tags);
```
bgrins, line 203: Nit: commented line can be removed

camillobruni: I mostly added them as comments since the code is maybe not fully clear (but happy to remove as well).

resources/developer-mode.mjs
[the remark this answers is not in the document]
camillobruni: I'd prefer the `+` here as well, but it get's auto-formatted to the template string 🤷 .

resources/developer-mode.mjs
```diff
@@ -155,11 +184,7 @@ function updateURL() {
...
     // to comma separate only the selected
-    const selectedSuites = [];
-    for (let suiteIndex = 0; suiteIndex < Suites.length; suiteIndex++) {
-        if (!Suites[suiteIndex].disabled)
-            selectedSuites.push(Suites[suiteIndex].name);
-    }
+    const selectedSuites = Suites.filter((suite) => !suite.disabled);
 
     if (!selectedSuites.length || selectedSuites.length === Suites.length) {
         url.searchParams.delete("tags");
```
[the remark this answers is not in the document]
camillobruni: Good idea. Let me do that in a follow-up PR.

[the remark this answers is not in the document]
camillobruni: Added https://github.com/WebKit/Speedometer/pull/233 to use a progress element.

PR #218, opened 2023-06-08
Fix broken tests - add tags to nuxt app

flashdesignory:
overlooked the tags in the tests...


rniwa (approved):

bgrins:
@camillobruni I think we should make the tags field optional (just default to an empty array if not specified).

camillobruni:
SG, will prepare a PR later today

Brian Grinstead ***@***.***> schrieb am Do., 8. Juni 2023,
18:43:

Issue #221, opened 2023-06-08
Speedometer is broken when ran inside another directory


rniwa:
The latest Speedometer is broken in NewsSite-Next because it specifies resources by absolute paths. We should be using relative path using ./dist/ instead.

rniwa:
@flashdesignory : I tried to fix it by replacing the base URL for production and running `npm run build` but that seemed to remove a whole bunch of files instead. Could you look into that?

flashdesignory:
Nuxt specific - we need to use `npm run generate` instead of the usual build command. 

Is this broken inside the tentative folder or in the open pr?

rniwa:
It's broken in the tentative folder.

flashdesignory:
this is using a relative path, but the only way I get it to work is if I pass in the path like: 

`"/resources/tentative/newssite/news-nuxt/dist/"`

rniwa:
I guess we may need to modify the file in dist directly...

rniwa:
Run HTTD on a directory above Speedometer such that you need to access Speedometer with Speedometer/ directory patch like https://browserbench.org/Speedometer2.1/

flashdesignory:
ah

rniwa:
Hm... specifying a relative path causes it to error in 404.

rniwa:
Hm... some Googling seems to indicate Nuxt doesn't support a relative path. That would be a pretty big show stopper.

flashdesignory:
this should work when running speedometer locally, but I assume it breaks when running from another parent directory

rniwa:
I think we need to disable this workload until we figure out how to solve this problem.

PR #225, opened 2023-06-08
Set base url dynamically for Nuxt site

danielroe:
This PR enables the Nuxt site to be accessed from multiple URLs.

For example:

* http://127.0.0.1:7000/resources/tentative/newssite/news-nuxt/dist/
* https://browserbench.org/Speedometer2.1/resources/tentative/newssite/news-nuxt/dist/

Before initialising the Nuxt router, we basically reset the base URL up until the `/dist/` path. We also set `cdnURL` to a relative path so we render the relative preload links in the HTML. (This will not be compatible with deep linking to other pages if in the future that becomes relevant.)

You can test locally by moving `resources/tentative/newssite/news-nuxt` after build to another folder (e.g. `bob`) and accessing `http://localhost:7000/bob/dist`.

Resolves https://github.com/WebKit/Speedometer/issues/221.


rniwa (approved):
Nice! Thank you!!

flashdesignory (approved):

PR #228, opened 2023-06-08
Display iteration marker in progress bar

camillobruni:
With the default 10 iterations:
Screenshot 2023-06-08 at 23 25 25

With 20:
Screenshot 2023-06-08 at 23 23 26

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -42,6 +42,15 @@ class MainBenchmarkClient {
         }
 
         this._developerModeContainer?.remove();
+        this._progressCompleted = document.getElementById("progress-completed");
+        if (params.iterationCount < 50) {
+            const progressNode = document.getElementById("progress");
+            for (let i = 1; i < params.iterationCount; i++) {
+                const iterationMarker = progressNode.appendChild(document.createElement("div"));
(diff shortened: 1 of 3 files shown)


flashdesignory (approved):

rniwa (approved):

resources/main.mjs
```diff
@@ -42,6 +42,15 @@ class MainBenchmarkClient {
         }
 
         this._developerModeContainer?.remove();
+        this._progressCompleted = document.getElementById("progress-completed");
+        if (params.iterationCount < 50) {
+            const progressNode = document.getElementById("progress");
+            for (let i = 1; i < params.iterationCount; i++) {
+                const iterationMarker = progressNode.appendChild(document.createElement("div"));
```
julienw, line 49: Wouldn't this be a great use case for https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress ? :-)

julienw (approved):

resources/main.mjs
camillobruni: Ah didn't know about that one. I'll try to add that next week then :).

PR #230, opened 2023-06-08
run format on all

flashdesignory:
tiny cleanup

@kara


camillobruni (approved):

bgrins (approved):

rniwa (approved):

rniwa:
Can we get before/after numbers for this change?

rniwa:
Can we get before/after numbers for this change?

flashdesignory:
@rniwa - I missed your comment about numbers. 
It's already merged and was just formatting code. 
There shouldn't be any difference in how the app behaves.

PR #233, opened 2023-06-12
Use progress element

camillobruni:


index.html
```diff
@@ -41,7 +41,7 @@
                 </a>
                 <div id="testContainer"></div>
                 <div id="progress">
-                    <div id="progress-completed"></div>
+                    <progress id="progress-completed"></progress>
```
bgrins, line 44: I think this should have an `aria-label` or be wrapped inside a `<label>` element as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#labelling

resources/main.css
```diff
@@ -310,7 +309,17 @@ button.show-about {
...
+    appearance: none;
+    border: none;
+    background-color: var(--inactive-color);
+}
+
+#progress-completed::-webkit-progress-value {
+    background-color: var(--foreground);
+}
+
+#progress-completed::-moz-progress-bar {
```
flashdesignory, line 322: any way to make the progress smoother? 
maybe something simple like:
`transition: all 0.3s linear;` 🤷

julienw: why not but it's not 100% needed as usually the steps are short in width, so it's not hugely visible anyway...

index.html
camillobruni: Added `aria-label`, thanks for the review.

resources/main.css
camillobruni: I think it want to have as little work as possible happening between the benchmark steps.
I'd prefer not having animations (except in workloads maybe).

PR #238, opened 2023-06-14
[Optional] : News site Nuxt - use hash router

flashdesignory:
This change uses a hash router - comparable to the next.js version and consistent with all other routers we use. 

* ability to point to an index.html in our tests
* consistency

Chrome - Before:
before

Chrome - After:
after

@kara


rniwa:
Can we have before/after numbers for each browser?

flashdesignory:
> Can we have before/after numbers for each browser?

Yeah, will add shortly

flashdesignory:
Firefox Before:
Firefox_Before

Firefox After:
Firefox_After

Safari Before:
Safari_Before

Safari After:
Safari_After_

Edge Before:
Edge_Before

Edge After:
Edge_After

bgrins:
FWIW the problem we were having in CI with the non index.html path has been resolved, but thank you for putting this together for comparison either way. If the non-hash router is more typically used with Nuxt (is it?) then it may make sense to stick with that (even if it's inconsistent with other tests internally)

rniwa:
Why is Chrome's total time going down by 16% whereas it stays same for other browsers?

flashdesignory:
> Why is Chrome's total time going down by 16% whereas it stays same for other browsers?

Maybe just weirdness in a single test.. I did 20 runs each before / after in chrome and here are some better comparisons:

before:
before

after:
after

flashdesignory:
>If the non-hash router is more typically used with Nuxt (is it?) then it may make sense to stick with that (even if it's inconsistent with other tests internally)

TLDR; Hash router is less popular, but the current router implementation is also not used exactly how it's done here (we are using it slightly differently to support Speedometer).

rniwa (approved):

bgrins:
> TLDR; Hash router is less popular, but the current router implementation is also not used exactly how it's done here (we are using it slightly differently to support Speedometer).

I'm OK either way and will leave it up to you whether you think the different-from-normal-usage of the more popular thing or the standard usage of the less popular thing is most representative for our purposes.

PR #246, opened 2023-06-19
Refactor Complex DOM CSS generator

issackjohn:
Refactored the CSS generator to use JSDOM and html markups to create the targeted CSS selectors.

With this PR we can generate CSS for the Complex DOM version of Angular TodoMVC.


lpardosixtosMs:
@bgrins can you take a look. @issackjohn refactored the Complex CSS generator to be more flexible. Now we can generate targeted css for the Angular TodoMVC.

bgrins:
Maybe naive question, but why can't we update the Angular markup to match all the others so it doesn't need special handling?

bgrins:
There are some updates in here that look good regardless, but it would be nice if we didn't have to work around a case for Angular specifically.

One other question - I see changes between the current CSS and the new CSS after this refactor. For example in the deleted matching.css there were some rules with `.top-bar` https://github.com/WebKit/Speedometer/pull/246/files#diff-2d83897a8d4871731e922269b9fa2c8eea145c17801e3e7c807748f0b9614bd7L1 but then in the new `generated.css` I don't see it https://github.com/WebKit/Speedometer/pull/246/files#diff-c8db8ce599a2ff7895e5a612f92938ffb6db09f4339e7928e0e300a93f4d1d47R1. I don't have an opinion on which CSS is better but would like to understand why that's happened. I.e. is that a consequence of the specified HTML changing (and then the seeded random number generator selects a different element instead)?

lpardosixtosMs:
> Maybe naive question, but why can't we update the Angular markup to match all the others so it doesn't need special handling?

I'm not familiar with angular, we assumed that those custom tags are the usual way to code and went with it. @flashdesignory would it make sense to change those tags to match the others?

lpardosixtosMs:
> There are some updates in here that look good regardless, but it would be nice if we didn't have to work around a case for Angular specifically.
> 
> One other question - I see changes between the current CSS and the new CSS after this refactor. For example in the deleted matching.css there were some rules with `.top-bar` https://github.com/WebKit/Speedometer/pull/246/files#diff-2d83897a8d4871731e922269b9fa2c8eea145c17801e3e7c807748f0b9614bd7L1 but then in the new `generated.css` I don't see it https://github.com/WebKit/Speedometer/pull/246/files#diff-c8db8ce599a2ff7895e5a612f92938ffb6db09f4339e7928e0e300a93f4d1d47R1. I don't have an opinion on which CSS is better but would like to understand why that's happened. I.e. is that a consequence of the specified HTML changing (and then the seeded random number generator selects a different element instead)?

Yes, even though we are using the same seed, the steps of the generator changed a little bit so now it picks different elements.

lpardosixtosMs:
> > Maybe naive question, but why can't we update the Angular markup to match all the others so it doesn't need special handling?
> 
> I'm not familiar with angular, we assumed that those custom tags are the usual way to code and went with it. @flashdesignory would it make sense to change those tags to match the others?

I imagine we'll have a similar issue if we introduce something using web components, so we might as well have flexible generator from the beginning.

flashdesignory:
> > Maybe naive question, but why can't we update the Angular markup to match all the others so it doesn't need special handling?
> 
> I'm not familiar with angular, we assumed that those custom tags are the usual way to code and went with it. @flashdesignory would it make sense to change those tags to match the others?

what specific markup are we referring to? It's been a while that I looked at this workload...

lpardosixtosMs:
> > > Maybe naive question, but why can't we update the Angular markup to match all the others so it doesn't need special handling?
> > 
> > 
> > I'm not familiar with angular, we assumed that those custom tags are the usual way to code and went with it. @flashdesignory would it make sense to change those tags to match the others?
> 
> what specific markup are we referring to? It's been a while that I looked at this workload...

The angular version have some extra `<app-todo-header>`, `<app-todo-list>`, and `<app-todo-item>` tags surrounding the header, list and items. 
image

flashdesignory:
Got it... the updated Angular version uses components (todo-list, todo-item, ect), instead of the previous version that was just one file (in speedometer 2.1). That's where these tags are coming from. 

This is the most common way I saw Angular devs using components in their apps and therefore the most representing way of a typical Angular app. 

There are probably several different ways to do this differently, but we usually tend to go with the most common approach.

resources/tentative/complex-static-html/src/big-dom-generator/html-markup.js
```diff
@@ -0,0 +1,50 @@
+export const ANGULAR_TODO_MVC_HTML_MARKUP = `
+<div class="main-ui" dir="ltr">
```
bgrins, line 2: This doesn't match the markup I see on https://speedometer-preview.netlify.app/resources/todomvc/architecture-examples/angular/dist/. 

I was pasting this in the web console to simulate 100 items, although it's not necessary to see the discrepency:

```
var numberOfItemsToAdd = 100;
var newTodo = document.querySelector(".new-todo");
  newTodo.value = `Something to do ${i}`;
  newTodo.dispatchEvent(new Event("input"))
  enter(newTodo)
}


function enter(node, options = undefined) {
  let keyCode = 13;
  let eventOptions = { bubbles: true, cancelable: true, keyCode, key: "Enter", which: keyCode };
  const event = new KeyboardEvent("keyup", eventOptions);
  node.dispatchEvent(event);
}
```

bgrins: We might consider generating the necessary HTML with 100 todo items from the live test contents (starting with something like the above) and pasting it here, compared with needing to keep a duplicate shell and todo generation logic in the script here.

I think that would be an improvement, as it would be easier to keep in sync with changes and I think have less room for error in the codegen script, but I don't feel strongly that it must happen right now.

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
```diff
@@ -1,238 +1,197 @@
...
 const cssProperties = ["accent-color", "border-bottom-color", "border-color", "border-left-color", "border-right-color", "border-top-color", "column-rule-color", "outline-color", "text-decoration-color"];
 
-// Returns a random 200 matching selectors and 200 non-matching selectors targeted at the todoMVC items.
-export const genCss = () => {
+/**
+ * Returns a random 200 matching selectors and 200 non-matching selectors targeted at the todoMVC items.
+ * @param {string} isAngular whether to generate angular or react markup
+ * @returns {string} The css rules for the matching and non-matching selectors.
+ */
+export const genCss = (isAngular = false) => {
```
bgrins, line 174: I guess we can cross this bridge if/when it happens but if there are other special cases to handle then we'd want to pass some other type besides a boolean around.

resources/tentative/complex-static-html/src/big-dom-generator/gen-css.js
```diff
@@ -1,238 +1,197 @@
...
-
-    // If we've reached the target length, pick a random classname from its children.
-    const getSelector = randomWeighted([getClassname, getType, () => "*"], [0.6, 0.3, 0.1]);
-    const selector = getSelector(depth, index);
-    if (selLen === badSelector) {
-        const wrongDepth = random.randRange(Math.min(depth + 1, 7), 8);
-        const wrongSelector = getClassname(wrongDepth, 0);
-        return selector + wrongSelector + oldCombinator;
-    }
+const ANGULAR_VIEW_DEPTH = 8;
```
bgrins, line 144: Since we now have JSDOM running in the script, is it feasible / worthwhile to derive these numbers from the HTML instead of having constants (i.e. querySelector for the relevant element and count the number of parents, or something). It looks like the only consumer is called from a function that has access to the element (`getInitialDepth`) so perhaps these numbers could be derived directly from the element (and it may not even need to know if we're running in Angular or not)

lpardosixtosMs: Agreed.

lpardosixtosMs:
> > There are some updates in here that look good regardless, but it would be nice if we didn't have to work around a case for Angular specifically.
> > One other question - I see changes between the current CSS and the new CSS after this refactor. For example in the deleted matching.css there were some rules with `.top-bar` https://github.com/WebKit/Speedometer/pull/246/files#diff-2d83897a8d4871731e922269b9fa2c8eea145c17801e3e7c807748f0b9614bd7L1 but then in the new `generated.css` I don't see it https://github.com/WebKit/Speedometer/pull/246/files#diff-c8db8ce599a2ff7895e5a612f92938ffb6db09f4339e7928e0e300a93f4d1d47R1. I don't have an opinion on which CSS is better but would like to understand why that's happened. I.e. is that a consequence of the specified HTML changing (and then the seeded random number generator selects a different element instead)?
> 
> Yes, even though we are using the same seed, the steps of the generator changed a little bit so now it picks different elements.

It is not only that `.top-bar` is not appearing, no classname on that level of the DOM is used, which seems unlikely. It is probably an error with the depths, we are investigating.

resources/tentative/complex-static-html/src/big-dom-generator/html-markup.js
bgrins: Makes sense. I'd like to get #239 landed to make this easier to compare with

flashdesignory:
> It is not only that `.top-bar` is not appearing, no classname on that level of the DOM is used, which seems unlikely. It is probably an error with the depths, we are investigating

@lpardosixtosMs - was this solved?

lpardosixtosMs:
> > It is not only that `.top-bar` is not appearing, no classname on that level of the DOM is used, which seems unlikely. It is probably an error with the depths, we are investigating
> 
> @lpardosixtosMs - was this solved?

@flashdesignory Yes, the latest patch addresses all the comments.

flashdesignory (approved):

bgrins (approved):

flashdesignory:
@rniwa - do you want to take a look at this one too, or is it good to merge?

issackjohn:
Just resolved a new conflict with .gitignore.

rniwa (approved):

rniwa:
Administrative: @issackjohn you should have a write access to the repository now. Let me know if you're not. Also, are you on webkit.slack.com?

issackjohn:
> Administrative: @issackjohn you should have write access to the repository now. Let me know if you do not. Also, are you on webkit.slack.com?
I still don't have write access. and no, I'm not on there, could you invite me please?

rniwa:
How about now?

issackjohn:
> How about now?

Yeah I have it now.

rniwa:
> > How about now?
> 
> Yeah I have it now.

In that case, you can go ahead & squash-merge this.

PR #249, opened 2023-06-20
[OPTIONAL]: TodoMVC built with Web Components

flashdesignory:
This pr is a simple implementation of the todo app using native web-components (with shadow dom).. 
This is a "backup" solution in case we want something in the workloads that uses web components extensively. 

Notes: 
- this uses the proposed updated styles as well and therefore the dom structure itself is somewhat bigger. 
- Due to the usage of the shadow doms, I had to find a different way to target the elements in the tests
- DOM structure is different compared to other todo apps 
- No build steps used to minify or bundle the app (I could minify the files at some point if need be)

Scores:
Chrome:
Screenshot 2023-06-20 at 3 08 56 PM

Firefox:
Screenshot 2023-06-20 at 3 09 10 PM

Safari:
Screenshot 2023-06-20 at 3 09 22 PM

Edge:
Screenshot 2023-06-20 at 3 09 32 PM

@kara


bgrins:
I suspect the Lit workload will cover similar technical ground, but we can always compare and contrast profiles.

flashdesignory:
> I suspect the Lit workload will cover similar technical ground, but we can always compare and contrast profiles.

Yup, this is an alternate version to cover web components, in case we don't get the pr in time 😄

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,164 @@
...
+
+    get isReady() {
+        return this._isReady;
+    }
+
+    getInstance() {
+        return this;
+    }
+
+    addItem = (e) => {
```
rniwa, line 36: Why are these instance variables??

flashdesignory: are you referring to `get isReady` & `getInstance` ?

I am currently not using `get isReady`, but that could be used to tell the runner when this app is ready to go do stuff. 

The `getInstance` method I added to be able to get a reference to the instance from the runner. 
Reason is that the usage of shadow doms makes it harder to just use `querySelector` from within the test steps

Happy to change that, if there are better options.

rniwa: I'm referring to addItem, toggleItem, etc...

flashdesignory: those are event handlers: https://github.com/WebKit/Speedometer/blob/a02536f4d1dc8c412ba351825628952c151b4405/resources/tentative/todomvc/src/components/todo-app/todo-app.component.js#L101

rniwa: I get that these are event handlers. Why are they instance variables instead of regular member functions?

flashdesignory: The reason I am using arrow functions instead of binding the functions in the constructor, is for readability. 
I can always refactor.... but I assume it would be mainly the todo-items where there might be a potential performance boost, since there's only one instance of the other components.

rniwa: @flashdesignory : using arrow functions is fine but using arrow functions which are defined as instance variables seems quite odd. I've never seen code that does that in real life.

flashdesignory: I can refactor to keep the handlers on the prototypes 👍 . 
I'll make that change and push updates.

rniwa: Or we could define those inside constructors. That's fine too. It's just very unusual for arrow functions to appear as class instance variables.

flashdesignory: changes done

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+
+    get isReady() {
+        return this._isReady;
+    }
+
+    getInstance() {
+        return this;
+    }
+
+    addItem(e) {
```
rniwa, line 43: Please spell out event.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+        const { detail: item } = e;
+
+        this._data.push(item);
+        this.list.addItem(item);
+
+        this.update("add-item", item.id);
+    }
+
+    toggleItem(e) {
```
rniwa, line 52: Ditto

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+        this._data.forEach((entry) => {
+            if (entry.id === e.detail.id)
+                entry.completed = e.detail.completed;
+
+        });
+
+        this.update("toggle-item", e.detail.id);
+    }
+
+    removeItem(e) {
```
rniwa, line 62: Ditto.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+        this._data.forEach((entry, index) => {
+            if (entry.id === e.detail.id)
+                this._data.splice(index, 1);
+
+        });
+
+        this.update("remove-item", e.detail.id);
+    }
+
+    updateItem(e) {
```
rniwa, line 72: Ditto.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+        this._data.forEach((entry) => {
+            if (entry.id === e.detail.id)
+                entry.title = e.detail.title;
+
+        });
+
+        this.update("update-item", e.detail.id);
+    }
+
+    toggleItems(e) {
```
rniwa, line 82: Ditto.

resources/tentative/todomvc/src/components/todo-app/todo-app.template.js
```diff
@@ -0,0 +1,20 @@
...
+const styles = "src/styles/app.css";
+const mainStyles = "src/styles/main.css";
+
+template.id = "todo-app-template";
+template.innerHTML = `
+    <link rel="stylesheet" href="${globals}" />
+    <link rel="stylesheet" href="${styles}" />
+    <link rel="stylesheet" href="${mainStyles}" />
+    <section class="app">
+        <slot name="topbar"></slot>
```
rniwa, line 12: I think more idiomatic thing to do is to have <todo-topbar></todo-topbar> instead and rely on custom elements to instantiate the component.

resources/tentative/todomvc/src/components/todo-bottombar/todo-bottombar.component.js
```diff
@@ -0,0 +1,96 @@
...
+        // state
+        this.connected = false;
+
+        this.clearCompletedItems = this.clearCompletedItems.bind(this);
+
+        this.shadow = this.attachShadow({ mode: "open" });
+    }
+
+    updateDisplay() {
+        if (this["total-items"] !== "0")
```
rniwa, line 25: It's somewhat odd to do string comparison like this.
Maybe `parseInt(this["total-items"])` or `+this["total-items"]` first?

resources/tentative/todomvc/src/components/todo-bottombar/todo-bottombar.component.js
```diff
@@ -0,0 +1,96 @@
...
+    constructor() {
+        super();
+
+        // elements
+        this.element = undefined;
+        this.clearCompletedButton = undefined;
+        this.todoStatus = undefined;
+        this.filterLinks = undefined;
+        // state
+        this.connected = false;
```
rniwa, line 17: Why do we need this? We can just use [Node.prototype.isConnected](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected)?

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
+import template from "./todo-app.template.js";
+import TodoTopbar from "../todo-topbar/todo-topbar.component.js";
+import TodoList from "../todo-list/todo-list.component.js";
+import TodoBottombar from "../todo-bottombar/todo-bottombar.component.js";
+import { useRouter } from "../../hooks/useRouter.js";
+
+// import { data } from "../../data.js";
```
rniwa, line 7: Maybe delete this commented out line?

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+class TodoApp extends HTMLElement {
+    constructor() {
+        super();
+
+        // elements
+        this.topbar = undefined;
+        this.list = undefined;
+        this.bottombar = undefined;
+        // state
+        this.connected = false;
```
rniwa, line 18: This boolean seems unnecessary given we have https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+
+        node.querySelector("[name=\"topbar\"]").append(this.topbar);
+        node.querySelector("[name=\"list\"]").append(this.list);
+        node.querySelector("[name=\"bottombar\"]").append(this.bottombar);
+
+        this.list.addItems(this._data);
+        this.addListeners();
+        this.update("connected");
+
+        this.shadow.append(node);
```
rniwa, line 157: Doesn't this mean we'd keep adding new clone of template each time the element becomes connected?
We should probably exit early when we've already done this once.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,171 @@
...
+
+        this.router.initRouter(this.routeChange);
+        this._isReady = true;
+    }
+
+    disconnectedCallback() {
+        this.connected = false;
+        this.removeListeners();
```
rniwa, line 165: What's the point of removing event listener here?

flashdesignory: In case the app ever gets removed from the DOM, which we don't need here. 
I can remove

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
flashdesignory: yup - i'll adjust. This app doesn't re-connect at any point as far as I can tell, but it should be done better.

flashdesignory: changed approach and appending the template in the constructors. 
connectedCallback will now only update the template and add listeners

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
flashdesignory: disconnectedCallback in all components will now be responsible to remove event listeners that are added in the connectedCallback.

rniwa:
So interesting. With Lit workload (https://github.com/WebKit/Speedometer/pull/251), browsers in the order of high score are: Chrome, Firefox, then Safari. With this workload, on the other hand, browsers in the order of high score are Safari, Chrome then Firefox with very little difference between Chrome & Firefox. This might be interesting / different enough workload to merit the inclusion based on that observation.

resources/benchmark-runner.mjs
```diff
@@ -105,6 +105,10 @@ class PageElement {
         return new PageElement(this.#node[name]());
     }
 
+    callElementMethod(name, ...args) {
+        return this.#node[name](args);
```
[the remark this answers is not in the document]
flashdesignory: yup, typo 😞 

@rictic - I am working on an alternate approach that might work for your pr as well. 
Hoping to have something up eod or tomorrow.

resources/tests.mjs
```diff
@@ -137,6 +137,60 @@ Suites.push({
...
+                elements[i].toggleInput.click();
+        }),
+        new BenchmarkTestStep("DeletingAllItems", (page) => {
+            const app = page.querySelector("todo-app").callElementMethod("getInstance");
+            const elements = app.list._elements;
+            for (let i = 0; i < numberOfItemsToAdd; i++)
+                elements[i].todoButton.click();
+        }), */
+        new BenchmarkTestStep(`Adding${numberOfItemsToAdd}Items`, (page) => {
+            const input = page.querySelector(".new-todo-input", ["todo-app", "todo-topbar"]);
```
flashdesignory, line 170: page.querySelector now accepts a 2nd optional parameter with an array to go through to target the selector. 

I kept previous implementation commented out for reference.

resources/tests.mjs
```diff
@@ -137,6 +137,60 @@ Suites.push({
...
+            for (let i = 0; i < numberOfItemsToAdd; i++) {
+                input.setValue(`Something to do ${i}`);
+                input.dispatchEvent("input");
+                input.enter("keyup");
+            }
+        }),
+        new BenchmarkTestStep("CompletingAllItems", (page) => {
+            const items = page.querySelectorAll("todo-item", ["todo-app", "todo-list"]);
+            for (let i = 0; i < numberOfItemsToAdd; i++){
+                const item = items[i].querySelector(".toggle-todo-input");
```
flashdesignory, line 180: PageElement now also has a querySelector method that I used here. This method looks at the shadow dom if it exists, otherwise a typical selector.

rniwa:
> > So interesting. With Lit workload (#251), browsers in the order of high score are: Chrome, Firefox, then Safari. With this workload, on the other hand, browsers in the order of high score are Safari, Chrome then Firefox with very little difference between Chrome & Firefox. This might be interesting / different enough workload to merit the inclusion based on that observation.
> 
> There are several questions:
> 
> * which workload(s) are more typical of real-world usage?

It's hard to tell.

> * what are the performance characteristics of this workloads compared to the lit-based one? Especially do you know why is Safari faring better here than in the other one? This could help us make a decision.

As far as I can tell, there isn't much difference between the two. In fact, Safari's time is very similar between two workloads: 11.2ms for WC vs. 12.3ms for Lit for adding items and 3.6ms (WC) vs. 5.7ms (Lit) for completing items. Nothing jumps out as unusual in the profiling either although these times are so fast that getting reliable profiling data is rather hard.

> `lit` is clearly gaining a lot of traction lately (x3 more downloads on npm in a year) I also know about Stencil which is more stable, but quite used otherwise. Do you have evidence of developers using web components without a library?

With that line of reasoning, we should not be testing vanilla JS at all since virtually every website in existence uses a major library/framework now.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,161 @@
...
+        this.list.listNode.removeEventListener("remove-item", this.removeItem);
+        this.list.listNode.removeEventListener("update-item", this.updateItem);
+
+        this.bottombar.removeEventListener(
+            "clear-completed-items",
+            this.clearCompletedItems
+        );
+    }
+
+    routeChange = (route) => {
```
rniwa, line 140: Make this to a regular method?

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
```diff
@@ -0,0 +1,194 @@
...
+    stopEdit() {
+        this.item.classList.remove("editing");
+    }
+
+    cancelEdit() {
+        this.editInput.blur();
+    }
+
+    toggleItem() {
+        // update item first, before dispatch
```
rniwa, line 91: I see that from the code but why is that necessary?

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
```diff
@@ -0,0 +1,194 @@
...
+        this.dispatchEvent(
+            new CustomEvent("toggle-item", {
+                detail: { id: this.id, completed: this.toggleInput.checked },
+                bubbles: true,
+            })
+        );
+    }
+
+    removeItem() {
+        // dispatch first, before updating item
```
rniwa, line 103: Ditto.

resources/tentative/todomvc/src/components/todo-item/todo-item.template.js
```diff
@@ -0,0 +1,23 @@
...
+const globals = "src/styles/global.css";
+const styles = "src/styles/todo-item.css";
+
+template.id = "todo-item-template";
+template.innerHTML = `
+    <link rel="stylesheet" href="${globals}" />
+    <link rel="stylesheet" href="${styles}" />
+    <li class="todo-item">
+        <div class="display-todo">
+            <label class="toggle-todo-label visually-hidden">Toggle Todo</label>
```
rniwa, line 11: Shouldn't this specify "for"?

resources/tentative/todomvc/src/components/todo-list/todo-list.component.js
```diff
@@ -0,0 +1,123 @@
...
+        else
+            this.listNode.style.display = "none";
+
+    }
+
+    updateView(element) {
+        switch (this._route) {
+            case "completed":
+                element.style.display
+                    = element.completed === "true" ? "block" : "none";
```
rniwa, line 69: It seems like this can fit in a single line??

resources/tentative/todomvc/src/components/todo-list/todo-list.component.js
```diff
@@ -0,0 +1,123 @@
...
+                break;
+            case "active":
+                element.style.display
+                    = element.completed === "true" ? "none" : "block";
```
rniwa, line 73: Ditto.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
flashdesignory: done - overlooked previously

resources/tentative/todomvc/src/components/todo-item/todo-item.template.js
flashdesignory: once the "id" attribute of the component is assigned, node ids and "for" attributes are set. 
https://github.com/flashdesignory/Speedometer/blob/todo-wc/resources/tentative/todomvc/src/components/todo-item/todo-item.component.js#L46

resources/tentative/todomvc/src/components/todo-list/todo-list.component.js
flashdesignory: fixed

resources/tentative/todomvc/src/components/todo-list/todo-list.component.js
flashdesignory: fixed

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
flashdesignory: 1. The app listens for the "remove-item" event and updates the attributes on topbar, list, bottombar.
2. The list removes the reference from the _elements array.
3. The item removes itself.

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
flashdesignory: 1.  The item updates its "completed" state   
2. The app listens for the "toggle-item" event and updates the attributes on topbar, list, bottombar.
3. The list checks the "completed" attribute to filter based on route (therefore I need the completed state to already be updated before the check)

resources/tentative/todomvc/src/components/todo-item/todo-item.template.js
flashdesignory: Example of rendered output:
Screenshot 2023-06-23 at 12 55 45 PM

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,162 @@
...
+import TodoTopbar from "../todo-topbar/todo-topbar.component.js";
+import TodoList from "../todo-list/todo-list.component.js";
+import TodoBottombar from "../todo-bottombar/todo-bottombar.component.js";
+import { useRouter } from "../../hooks/useRouter.js";
+class TodoApp extends HTMLElement {
+    constructor() {
+        super();
+        // state
+        this._isReady = false;
+        this._data = [];
```
rniwa, line 11: Should we use private variables instead?

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,162 @@
...
+        // elements
+        const node = document.importNode(template.content, true);
+        this.topbar = new TodoTopbar();
+        this.list = new TodoList();
+        this.bottombar = new TodoBottombar();
+        node.querySelector("[name=\"topbar\"]").append(this.topbar);
```
rniwa, line 17: I think it's rather unusual for component to be instantiating each subcomponent manually like this instead of relying on custom elements to create them automatically.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,162 @@
...
+        node.querySelector("[name=\"list\"]").append(this.list);
+        node.querySelector("[name=\"bottombar\"]").append(this.bottombar);
+        // add items, if data
```
rniwa, line 20: These comments unnecessary? They just repeat what the code already tells us.

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
rniwa: So I think it would be more useful for this comment to say 'The todo-list checks the "completed" attribute to filter based on route (therefore the completed state needs to already be updated before the check)'

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
```diff
@@ -0,0 +1,194 @@
...
+                detail: { id: this.id },
+                bubbles: true,
+            })
+        );
+        this.remove();
+    }
+
+    updateItem(event) {
+        if (event.target.value !== this.title) {
+            if (event.target.value.length === 0) {
```
rniwa, line 115: WebKit's coding style is to do: !event.target.value.length instead.

resources/tentative/todomvc/src/components/todo-item/todo-item.template.js
rniwa: That seems rather unusual way for shadow DOM to operate. IDs in each shadow tree lives in a different namespace so we should be able to just say `id="toggle-todo"` and `for="toggle-todo"` and be done with it. Same goes for edit-todo and remove-todo as well. These can all be "edit-todo" and "remove-todo".

resources/tentative/todomvc/src/components/todo-list/todo-list.component.js
```diff
@@ -0,0 +1,121 @@
...
+
+class TodoList extends HTMLElement {
+    static get observedAttributes() {
+        return ["total-items"];
+    }
+
+    constructor() {
+        super();
+        // state
+        this._elements = [];
```
rniwa, line 12: Should we consider using private variables?

resources/tentative/todomvc/src/components/todo-topbar/todo-topbar.component.js
```diff
@@ -0,0 +1,139 @@
...
+                    completed: false,
+                },
+            })
+        );
+
+        event.target.value = "";
+    }
+
+    updateDisplay() {
+        if (parseInt(this["total-items"]) === 0) {
```
rniwa, line 67: WebKit's coding style is to do `!parseInt(~)` instead.

resources/tentative/todomvc/src/hooks/useDoubleClick.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * A simple function to normalize a double-click and a double-tab action.
+ * There is currently no comparable tab action to dblclick.
+ *
+ * @param {Function} fn
+ * @param {number} delay
+ * @returns
+ */
+export function useDoubleClick(fn, delay){
```
rniwa, line 9: Space between ) and {?

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
```diff
@@ -0,0 +1,194 @@
...
+
+    connectedCallback() {
+        this.update("id", "title", "completed");
+
+        this.keysListeners.push(
+            useKeyListener({
+                target: this.editInput,
+                event: "keyup",
+                callbacks: {
+                    ["Enter"]: this.updateItem,
```
rniwa, line 170: Not sure how this works. Did you mean to say `this.updateItem.bind(this)` or `(event) => this.updateItem(event)`?

flashdesignory: the event object is passed along when `this.updateItem` is called, which is already bound in the constructor.

resources/tentative/todomvc/src/components/todo-item/todo-item.template.js
flashdesignory: done!

resources/tentative/todomvc/src/components/todo-item/todo-item.component.js
rniwa: Oh I see.

rniwa (approved):

resources/tentative/todomvc/src/components/todo-list/todo-list.template.js
```diff
@@ -0,0 +1,12 @@
+const template = document.createElement("template");
+const globals = "src/styles/global.css";
+const styles = "src/styles/todo-list.css";
+
+template.id = "todo-list-template";
+template.innerHTML = `
+    <link rel="stylesheet" href="${globals}" />
```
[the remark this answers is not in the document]
rniwa: > I believe using `<link rel=stylesheet>` in an element's template may cause a fetch of the stylesheet for each instance of the element. With proper caching this may still be reasonably fast but as I understand it the platform's intended primitive for this use case is constructible stylesheets.

Browser would cache the content referenced by link element just fine.

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
flashdesignory: @rniwa @rictic - seemed like an easy change, but threw me off for a little bit. 
I made the change, adding the custom elements directly in the template file. 

The order of the scripts in the head of the html also matter, since I need to ensure those used in the template file are added first, so they can register properly before the app tries to interact with them.

Let me know what you think.

resources/tentative/todomvc/src/components/todo-app/todo-app.template.js
flashdesignory: Hopefully addressed in https://github.com/WebKit/Speedometer/pull/249#discussion_r1240598703 below

resources/tentative/todomvc/src/components/todo-app/todo-app.component.js
rniwa: Looks good to me.

flashdesignory:
@rniwa @rictic - using constructable styles makes a difference. 
The latest push is using them.. 

Chrome before:
chrome_before

Chrome after:
chrome_after

Safari before:
safari_before

Safari after: 
safar_after

Firefox before:
firefox_before

Firefox_after:
firefox_after

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/app.constructable.js
```diff
@@ -0,0 +1,5 @@
+const sheet = new CSSStyleSheet();
+sheet.replaceSync(
+    ":host {\n    display: block;\n\tbox-shadow: none !important;\n\tmin-height: 68px;\n}\n\n.app {\n\tbackground: #fff;\n\tmargin: 24px 16px 40px 16px;\n\tposition: relative;\n\tbox-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.2),\n\t            0 25px 50px 0 rgba(0, 0, 0, 0.1);\n}\n"
```
julienw, line 3: I wonder why you didn't use template strings with "\`" so that it could be much more readable than with these `\n` everywhere? Is there a good reason?

resources/todomvc/vanilla-examples/javascript-web-components/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,147 @@
...
+    constructor() {
+        super();
+
+        const node = document.importNode(template.content, true);
+        this.topbar = node.querySelector("todo-topbar");
+        this.list = node.querySelector("todo-list");
+        this.bottombar = node.querySelector("todo-bottombar");
+
+        this.shadow = this.attachShadow({ mode: "open" });
+        this.htmlDirection = document.querySelector("html").getAttribute("dir") || "ltr";
```
julienw, line 19: nit: I believe that `document.dir` should give you this value, you don't need to use getAttribute
```suggestion
        this.htmlDirection = document.dir || "ltr";
```
or even
```suggestion
        this.htmlDirection = document.dir;
```

resources/todomvc/vanilla-examples/javascript-web-components/src/components/todo-app/todo-app.component.js
```diff
@@ -0,0 +1,147 @@
...
+        this.shadow.host.setAttribute("dir", this.htmlDirection);
```
julienw, line 20: I believe `this.shadow.host` is exactly the same as `this`.
```suggestion
        this.setAttribute("dir", this.htmlDirection);
```

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/global.constructable.js
```diff
@@ -0,0 +1,5 @@
+const sheet = new CSSStyleSheet();
```
julienw, line 1: I have a concern that this might diverge from global.css eventually. Would it be possible to fetch the content of global.css there, maybe at build time?

A simpler solution could be that you add a comment both here and there to make sure to update one when the other changes. It should also be easier if you use "`" template strings.

resources/todomvc/vanilla-examples/javascript-web-components/src/components/todo-item/todo-item.component.js
```diff
@@ -0,0 +1,181 @@
...
+        this.updateItem = this.updateItem.bind(this);
+        this.toggleItem = this.toggleItem.bind(this);
+        this.removeItem = this.removeItem.bind(this);
+        this.startEdit = this.startEdit.bind(this);
+        this.stopEdit = this.stopEdit.bind(this);
+        this.cancelEdit = this.cancelEdit.bind(this);
+    }
+
+    update(...args) {
+        [...args].forEach((argument) => {
```
julienw, line 46: It's not clear to me why you're spreading `args`, it's already an array as far as I understand.

resources/todomvc/vanilla-examples/javascript-web-components/src/components/todo-item/todo-item.component.js
```diff
@@ -0,0 +1,181 @@
...
+                        this.item.id = `todo-item-${this.id}`;
+                    break;
+                case "title":
+                    if (this.title !== undefined) {
+                        this.todoText.textContent = this.title;
+                        this.editInput.value = this.title;
+                    }
+                    break;
+                case "completed":
+                    this.toggleInput.checked = this.getAttribute("completed") === "true" ? true : false;
```
julienw, line 59: My understanding is that attributes are mirrored as local properties, so we can just do that:
```suggestion
                    this.toggleInput.checked = this.completed === "true" ? true : false;
```

resources/todomvc/vanilla-examples/javascript-web-components/src/data.js
```diff
@@ -0,0 +1,17 @@
+export const data = [
```
julienw, line 1: Is this data used anywhere? If you want to keep it for testing or development purpose, can you please add a comment explaining how to use it?

resources/todomvc/vanilla-examples/javascript-web-components/src/index.js
```diff
@@ -0,0 +1 @@
+// placeholder
```
julienw, line 1: Is it useful for something? I believe you can remove this file and from `index.html`, right?

julienw (commented):
Thanks for the work, this looks pretty solid! Also it's pretty nice to finally have an example that uses esm without a build step. IMO this may be the biggest interest for this workload.

I added a few comments, I didn't add them everywhere they apply by simplicity but of course they apply in all similar code.

I'm not approving just yet because I'd like to see the code with these changes again. Please add new commits for the changes so that the new review will be easy to do :-)

resources/todomvc/vanilla-examples/javascript-web-components/src/data.js
flashdesignory: removed

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/global.constructable.js
flashdesignory: the constructable css files (or any styles) will be updated once this pr lands: https://github.com/WebKit/Speedometer/pull/254

adjustments on how to generate them, should ideally happen there.

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/app.constructable.js
flashdesignory: that's how my rollup plugin from https://github.com/WebKit/Speedometer/pull/254 generated the output. 
Can we modify there and update once that pr lands?

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/global.constructable.js
flashdesignory: re: diverge of global.css - the rollup plugin generates the constructable js files from the css files.
Once that pr is merged, we can install the styles as a local packages, similar to the news sites and that "should" keep things synced.

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/app.constructable.js
flashdesignory: updated in both prs - thanks @julienw !

resources/todomvc/vanilla-examples/javascript-web-components/src/styles/global.constructable.js
julienw: sounds good to me

resources/todomvc/vanilla-examples/javascript-web-components/src/components/todo-bottombar/todo-bottombar.component.js
```diff
@@ -0,0 +1,82 @@
...
+
+        const node = document.importNode(template.content, true);
+        this.element = node.querySelector(".bottombar");
+        this.clearCompletedButton = node.querySelector(".clear-completed-button");
+        this.todoStatus = node.querySelector(".todo-status");
+        this.filterLinks = node.querySelectorAll(".filter-link");
+
+        this.shadow = this.attachShadow({ mode: "open" });
+        this.htmlDirection = document.querySelector("html").getAttribute("dir") || "ltr";
+        this.shadow.host.setAttribute("dir", this.htmlDirection);
```
julienw, line 24: I wrote in my previous comments that the same comments were applying in other files :-) can you fix those too?

1. `document.querySelector("html").getAttribute("dir")` => `document.dir`
2. `this.shadow.host` => `this`

In all components

Thanks :)

Issue #52, opened 2023-01-26
Add a lit-html todomvc demo


bgrins:
As discussed in #3. Unfortunately there's not an example at https://github.com/tastejs/todomvc/tree/master/examples so we may have to write one from scratch or look around elsewhere

rniwa:
@justinfagnani : Any appetite from lit-html community to add this as a benchmark content for Speedometer v3?

Everyone: should this be v3 blocker or is it deferrable to v4?

justinfagnani:
Definitely, lots of appetite!

I'm about to take some vacation, but let me see if the team can jump on this. What's the timeframe for v3 look like?

bgrins:
> There's a PR for Lit here: https://github.com/tastejs/todomvc/pull/2178 but it also includes Redux. I can make a version that doesn't have Redux.

Yes, something like that without Redux sounds nice. Quickly scanning that I believe we've been using a shared CSS file instead of CSS-in-JS but I might be misremembering (cc @flashdesignory who's been driving the version updates and cleanup with the workloads - for example in https://docs.google.com/document/d/1MXxk9DryJmvAmPMt2uMneaS3J5Y-dMFGt4EGvMzhhJE/edit)

> What's the timeframe for v3 look like?

We are aiming for a code freeze on workloads before the end of half.

rniwa:
>
> I'm about to take some vacation, but let me see if the team can jump on this. What's the timeframe for v3 look like?

We're planning to finalize the benchmark content by the end of June.

justinfagnani:
@bgrins 

> Quickly scanning that I believe we've been using a shared CSS file instead of CSS-in-JS

Any library using shadow DOM will need to consume CSS in a way that let's them add it to the shadow roots, and not add it globally. CSS-in-JS is how Lit does this. We also support CSS modules (with import attributes), but that's only available in Chromium browsers right now.

bgrins:
> How that's done, should be framework specific in my opinion.

Sounds good, thanks @justinfagnani!

justinfagnani:
For Speedometer purposes I would much rather do things the idiomatic Lit way - which is to have inline CSS-in-JS scoped to that one shadow root, than try to import a global stylesheet into every shadow root. That will test the characteristics of real-world Lit projects.

[...]

Issue #165, opened 2023-05-19
Review the "About" text on the homepage


bgrins:
We should audit and likely update the About text on the homepage before we launch 3

bgrins:
The actual content is here https://github.com/WebKit/Speedometer/blob/64798bc754db2d4020bc1e2adae205af02305ee9/index.html#L113

bgrins:
This was done in #293

PR #251, opened 2023-06-21
Add a Lit TodoMVC app

rictic:
This aims to fix https://github.com/WebKit/Speedometer/issues/52


resources/todomvc/architecture-examples/lit/src/lib/todos.ts
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @fires a `change` event when the todo list changes.
+ */
+export class Todos extends EventTarget {
+	#nextId;
+	#todos: Array<Todo>;
+	#filter: TodoFilter = this.#filterFromUrl();
+	constructor() {
+		super();
+		const stored = window.localStorage.getItem("todos");
```
julienw, line 24: I'm worried about the use of `localStorage`.

1. Is this use typical in real-world applications? This involves a lot of JSON.parse/stringify as well.
2. This may be a problem with the repeatibility of the benchmark, we need to make sure that the data in the storage is empty at the start of each run.

What do other folks think? @rniwa @bgrins @camillobruni @flashdesignory

resources/todomvc/architecture-examples/lit/package-lock.json
julienw, line 1: When running `npm i` locally, my npm removes a few entries in this file. Can you make sure the checked in version is the correct one? (I use npm v9.5.1)

.prettierignore
```diff
@@ -9,6 +9,7 @@ package-lock.json
 /resources/todomvc/architecture-examples/*
 !/resources/todomvc/architecture-examples/angular
 !/resources/todomvc/architecture-examples/backbone
+!/resources/todomvc/architecture-examples/lit
```
julienw, line 12: can you please make sure that it's also enabled in .eslintignore?

resources/tests.mjs
```diff
@@ -404,6 +404,32 @@ Suites.push({
...
+    tags: ["todomvc"],
+    async prepare(page) {
+        await page.wait
+        await page.waitForElement("todo-app");
+    },
+    tests: [
+        new BenchmarkTestStep(`Adding${numberOfItemsToAdd}Items`, (page) => {
+            const todoApp = page.querySelector('todo-app');
+            for (let i = 0; i < numberOfItemsToAdd; i++) {
+                todoApp.dispatchEvent('speedometer-addtodo', {detail: `Something to do ${i}`}, CustomEvent);
```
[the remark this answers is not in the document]
julienw: That said I believe the result is similar, because then this dispatches the same events as before under the hood.

resources/todomvc/architecture-examples/lit/src/lib/todos.ts
flashdesignory: some of the original todoMVC apps (todomvc.com) use local storage. 
We don't use that in the speedometer specific todo apps though.

bgrins: There have been concerns raised in the past about including `localStorage` or other APIs that hit disc in terms of I/O dominating measurements - so even if they are widely used APIs the current guidance is to not use them. I think we could revisit that decision if there was a strong functional need for them in a workload, but that's not the case here so I would drop it.

resources/tests.mjs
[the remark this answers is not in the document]
bgrins: Ah, I see the argument for how it's done in the readme:

> The speedometer TodoMVC tests typically interact directly with DOM elements, emulating typing into the `<input>` and clicking on buttons and checkboxes, etc. Because Lit uses the Shadow DOM, this code lives in this app and the benchmark code simply dispatches `speedometer-*` events on the `<todo-app>` element, which then calls the same APIs that speedometer usually would to emulate the requested user input.

Thoughts, @rniwa @flashdesignory ?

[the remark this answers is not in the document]
flashdesignory: if it's easy to get a reference to the actual input from within the test file, that's preferred maybe. 
I didn't look at the workload yet, but maybe the reason was that it's not that easy to get that reference directly.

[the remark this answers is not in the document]
bgrins: Discussed this with @flashdesignory who's going to make a suggestion on the best way to handle this (as it's also come up in #249). So this can remain unchanged in the meantime, but the other comments on the PR look ready to address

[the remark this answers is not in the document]
rictic: I copied the callElementMethod from your PR @flashdesignory to expose the elements directly. This way the Lit benchmark can otherwise be the same as the other TodoMVC ones.

resources/todomvc/architecture-examples/lit/rollup/index.html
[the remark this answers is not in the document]
rictic: +1, done

flashdesignory:
@rictic - lint check is failing. A good way to catch things is by running `npm run format` from Speedometer's root locally before pushing.  
Speedometer has it's own prettier / eslint rules for consistency between workloads and to get as close as possible to Webkits style guide.

rictic:
Huh, I think something is wrong with my prettier setup, as `npm run format` is making many changes to unrelated directories. Investigating.

rictic:
I was definitely doing something wrong. This should now be properly formatted and passing eslint

resources/todomvc/architecture-examples/lit/src/lib/todos.ts
```diff
@@ -0,0 +1,141 @@
...
+                return this.completed;
+            case "all":
+            case undefined:
+            // fall through
+        }
+        return this.all;
+    }
+
+    #notifyChange() {
+        window.localStorage.setItem("todos", JSON.stringify({ todos: this.#todos, nextId: this.#nextId }));
```
bgrins, line 63: One more localStorage setter that can be removed here

rictic: +1, done

resources/tests.mjs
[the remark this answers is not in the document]
flashdesignory: @rictic - if you have time, do you want to try this and let me know if that would work for the lit workload?
https://github.com/WebKit/Speedometer/pull/253

flashdesignory:
@rictic - I noticed that you are defining the styles inside js, which is more idiomatic to lit, but doesn't let us manage consistency in the styles across todomvc apps. This workload also looks slightly different compared to the other ones.

I added an alternate approach with css-modules here (if approved):
https://github.com/WebKit/Speedometer/pull/254

these could be installed as a local package and that could potentially work?
Note: these styles expect a slightly different dom structure. 

Just posting here for thoughts.

resources/tests.mjs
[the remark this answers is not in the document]
flashdesignory: @rictic - the pr mentioned above is merged into main

flashdesignory:
This build looks great and it's pretty easy to follow along and parsing the code 🚀 .

The only thing that threw me off was that the Todos class also handles route changes. I'd expect the app to listen for those and notify or set the filter in the Todos class., but that's maybe subjective. 

Besides that, the only other difference I noticed was here the todos id is a simple int that increments, vs other implementations where we use uuid / nanoid.

resources/tests.mjs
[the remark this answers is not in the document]
rictic: That works, but it required some changes to page.querySelectorAll because the DOM is set up like:

* todo-app
  * todo-list
    * todo-item
      - .toggle
      - .destroy
    * todo-item
      - .toggle
      - .destroy
    * ...

The current implementation does the "selectAll" logic only at the final lowest level, so it would only see the first `.toggle` element

rictic:
There's a bunch of distinct things that people commonly call CSS modules, would we be able to import them and get a constructable stylesheet? If so, it'd be easy to integrate into a Lit app or just about any web component, and I'd be happy to update this PR with it.

I was split on putting the routing in the Todos class, an earlier draft actually did put it into `<todo-app>`'s code. Either seems fairly reasonable to me tbh, happy to move it if you'd like.

Good catch on the ids, I'll check out nanoid.

rniwa:
> CSS modules

It's probably best to refrain from using features that are not universally available. Otherwise, we'd be comparing apples to oranges across browsers.

resources/tests.mjs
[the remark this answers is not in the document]
flashdesignory: Can you try this?

Page.querySelectorAll -> todo-items (returns array of PageElements)

PageElement.querySelectorInShadowRoot -> .toggle
PageElement.querySelectorInShadowRoot -> .destroy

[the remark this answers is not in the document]
flashdesignory: Here is an example: https://github.com/WebKit/Speedometer/blob/e428accf51b228c1e1c207e0d8edabc518fcd20c/resources/tests.mjs#L178

flashdesignory:
> > CSS modules
> 
> It's probably best to refrain from using features that are not universally available. Otherwise, we'd be comparing apples to oranges across browsers.

I should have clarified - I assumed there's support for css loaders that would allow us to import css modules.

flashdesignory:
I don't know lit well enough, but for example something like: https://www.npmjs.com/package/lit-css-loader

rictic:
@rniwa yeah, not proposing anything that isn't shipping cross browser, but it's easy to do build time work to get a JS module that exports a constructible stylesheet.

https://github.com/Alorel/rollup-plugin-constructable-css looks like a nice API

flashdesignory:
> @rniwa yeah, not proposing anything that isn't shipping cross browser, but it's easy to do build time work to get a JS module that exports a constructible stylesheet.
> 
> https://github.com/Alorel/rollup-plugin-constructable-css looks like a nice API

thanks @rictic - i looked at the plugin and it's pretty straight forward. 
I'll add it

flashdesignory:
@rictic - turns out that I couldn't install it.. 
I did write a super simplified plugin to support your use-case though (at least I hope that's what you need).
https://github.com/WebKit/Speedometer/pull/254

the dist output folder now has constructable files (app.constructable.js, ect...).

lmk - happy to keep modifying the plugin until it gives you what you need.

resources/tests.mjs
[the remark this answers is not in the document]
rictic: Sure, that works. IMO it's less elegant but it works fine. Updated.

rictic:
The .constructable.js files look like they'll work great. We can replace the styles in the `static styles` arrays with those.

rictic:
Let me know if there's anything I can do here to help this to land, I'm very much interested in getting this into Speedometer 3!

flashdesignory:
> Let me know if there's anything I can do here to help this to land, I'm very much interested in getting this into Speedometer 3!

Sounds good - will review today

resources/todomvc/architecture-examples/lit/src/lib/todos.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    filtered() {
+        switch (this.#filter) {
+            case "active":
+                return this.active;
+            case "completed":
+                return this.completed;
+            case "all":
+            case undefined:
```
flashdesignory, line 97: we we need these two lines or is it always falling through and returning this.all?

resources/todomvc/architecture-examples/lit/package-lock.json
flashdesignory: local install still modifies the package-lock file 🤷

flashdesignory:
@bgrins @rniwa - Do we want to land this pr and deal with external css / constructable css mod later? 
Don't want to block this pr and it feels like it's almost there.

resources/todomvc/architecture-examples/lit/src/lib/todos.ts
rictic: Didn't think about that! Done

flashdesignory (approved):
LGTM - potential alignment with a11y related updates could happen in a follow up pr.

julienw:
Because this wasn't super clear for everybody, the lit code already uses socalled "constructable css" with its `css` template tag, and uses `adoptedStyleSheets` under the hood.
The related code is: https://github.com/lit/lit/blob/3711e6650a59966e5be8d92dd0abf053d9a50d32/packages/reactive-element/src/css-tag.ts

resources/todomvc/architecture-examples/lit/package-lock.json
julienw: Yeah I see this too, but when I run `git diff -w` (so without the whitespace changes) then I see no changes. I wonder where this might come from.

rictic:
Yeah, expanding on that, the `css` template tag returns a `CSSResult`. We'd like it to just return a `CSSStyleSheet`, but we support older browser versions that didn't yet have constructible stylesheets. I could update this example to only use `CSSStyleSheet`s, a LitElement's `styles` can take a `CSSStyleSheet`. I went with this approach because it's the most common way that Lit code is written in the wild.

resources/todomvc/architecture-examples/lit/src/lib/todos.ts
rictic: Good point! Done

resources/todomvc/architecture-examples/lit/package-lock.json
rictic: strange! ran `npm i` and `npm run format` again and committed the result, hopefully that solves it!

resources/todomvc/architecture-examples/lit/src/lib/todo-app.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    override render() {
+        return html`<section>
+            <header class="header">
+                <h1>todos</h1>
+                <todo-form .todoList=${this.todoList}></todo-form>
+            </header>
+            <section class="main">
+                <todo-list .todoList=${this.todoList}></todo-list>
+            </section>
```
flashdesignory, line 86: I saw that we're replacing `<section class="main">` with `<main class="main">` in other workloads...

```suggestion
            <main class="main">
                <todo-list .todoList=${this.todoList}></todo-list>
            </main>
```

resources/todomvc/architecture-examples/lit/package-lock.json
flashdesignory: that did it for me - no changes to the lock file with npm install.

resources/todomvc/architecture-examples/lit/src/lib/todo-app.ts
rictic: Done!

julienw:
> Yeah, expanding on that, the `css` template tag returns a `CSSResult`. We'd like it to just return a `CSSStyleSheet`, but we support older browser versions that didn't yet have constructible stylesheets. I could update this example to only use `CSSStyleSheet`s, a LitElement's `styles` can take a `CSSStyleSheet`. I went with this approach because it's the most common way that Lit code is written in the wild.

Yes, this sounds good to me! I just wanted other folks know that this uses `adoptedStyleSheets` under the hood (for our current browser versions).

resources/todomvc/architecture-examples/lit/package.json
```diff
@@ -1,6 +1,10 @@
 {
     "private": true,
     "type": "module",
+    "engines": {
+        "node": ">=18.13.0",
+        "npm": ">=8.19.3"
+    },
```
julienw, line 7: you need to run `npm i` and commit package-lock.json again after this change :-)

.prettierignore
```diff
@@ -9,6 +9,9 @@ package-lock.json
...
+/resources/todomvc/architecture-examples/lit/dist
+/resources/todomvc/architecture-examples/lit/.wireit
```
julienw, line 14: do we need those in .eslintignore as well?

resources/benchmark-runner.mjs
```diff
@@ -147,6 +147,10 @@ class PageElement {
         return new PageElement(this.#node[name]());
     }
 
+    callElementMethod(name, ...args) {
+        return this.#node[name](...args);
+    }
```
julienw, line 152: Is this still needed?

resources/todomvc/architecture-examples/lit/src/lib/todo-footer.ts
```diff
@@ -0,0 +1,138 @@
...
+                      <strong>${this.todoList?.active.length}</strong>
+                      items left
+                  </span>
+                  <ul class="filters">
+                      <li>
+                          ${filterLink({
+        text: "All",
+        filter: "all",
+        selectedFilter: this.todoList?.filter,
+    })}
```
julienw, line 103: Prettier seems to fail on this, you might want to report an issue at them. But nothing to do here for you!

julienw (approved):
Thanks, this looks good to me!
Just a few comments (including the one I left earlier about running `npm i`) and then we're good to go!

.prettierignore
rictic: +1, done

resources/benchmark-runner.mjs
rictic: Ah, it isn't! Reverted

resources/todomvc/architecture-examples/lit/package.json
rictic: Argh, I always forget! Thanks for catching that

resources/todomvc/architecture-examples/lit/src/lib/todo-footer.ts
rictic: Good point, though we can also make things a bit easier on prettier too.

Reformatted a bit

index.html
```diff
@@ -115,9 +115,9 @@ <h1 class="section-header">About Speedometer 2.1</h1>
...
-                            call DOM APIs directly from ECMAScript 5 (ES5), ECMASCript 2015 (ES6), ES6 transpiled to ES5, and Elm transpiled to ES5. Others use one of eleven popular JavaScript frameworks: React, React with Redux, Ember.js,
-                            Backbone.js, AngularJS, (new) Angular, Vue.js, jQuery, Preact, Inferno, and Flight. Many of these frameworks are used on the most popular websites in the world, such as Facebook and Twitter. The performance of these types
-                            of operations depends on the speed of the DOM APIs, the JavaScript engine, CSS style resolution, layout, and other technologies.
+                            call DOM APIs directly from ECMAScript 5 (ES5), ECMASCript 2015 (ES6), ES6 transpiled to ES5, and Elm transpiled to ES5. Others use one of twelve popular JavaScript frameworks: React, React with Redux, Ember.js,
+                            Backbone.js, AngularJS, (new) Angular, Vue.js, jQuery, Preact, Inferno, Flight, and Lit. Many of these frameworks are used on the most popular websites in the world, such as Facebook and Twitter. The performance of these
```
rniwa, line 119: We got rid of old angular, inferno, and flight. If we're updating this list, we might as delete those as well.

flashdesignory: There's more text that should change, since we have different type of workloads now. 
Maybe we should revert this change and tackle an update separately?

rniwa: Yeah, that would make sense.

rictic: Reverted.

flashdesignory: Thanks - changes regarding the about text can be tracked in this issue: https://github.com/WebKit/Speedometer/issues/165

resources/todomvc/architecture-examples/lit/dist/index.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
+<html>
+	<head>
```
rniwa, line 3: Can we use spaces instead of tabs like all other files?

rniwa (approved):

resources/todomvc/architecture-examples/lit/dist/index.html
rictic: +1, done

PR #220, opened 2023-06-08
[charts] Group substeps to make them more stable

julienw:
This is for preparing the move of the charts workload out of tentative.

[this pull request](https://make-charts-more-stable--speedometer-preview.netlify.app/?suite=Charts-Observable-plot,Charts-Chartjs)
[this pull request with raf measurements](https://make-charts-more-stable--speedometer-preview.netlify.app/?suite=Charts-Observable-plot,Charts-Chartjs&measurementMethod=raf)
[current main](https://speedometer-preview.netlify.app/?suite=Charts-Observable-plot,Charts-Chartjs)
[current main with raf measurements](https://speedometer-preview.netlify.app/?suite=Charts-Observable-plot,Charts-Chartjs&measurementMethod=raf)

On my machine, this seems to make the workloads more stable, both in Chrome and Firefox. Tell me what you think @rniwa @flashdesignory !

PR #252, opened 2023-06-21
Move the charts workload out of the tentative directory

julienw:
Here is a link to the [deploy preview](https://move-charts-out-of-tentative2--speedometer-preview.netlify.app/?suite=Charts-Observable-plot,Charts-Chartjs#details)

I looked at Firefox and Chrome. Chrome still has the "bad" first iteration, but we agreed that this was OK. I also see some GCMinor happening sometimes in Chrome.

What do you think @rniwa @camillobruni?

I still intend to pursue work such as #220 to group the substeps and try to make some longer.


rniwa:
Should we land https://github.com/WebKit/Speedometer/pull/220 first?

bgrins:
> Should we land https://github.com/WebKit/Speedometer/pull/220 first?

Sure, looks like it's approved now so that should be unblocked

julienw:
The order isn't important to me, what's important is keeping some pace.

rniwa (approved):

flashdesignory (approved):

PR #253, opened 2023-06-21
Adjusting Page.querySelector / Page.querySelectorAll

flashdesignory:
With the addition of web components, the previous implementations of `Page.querySelector` and `Page.querySelectorAll `didn't let us target elements inside a shadow dom.

This pr augments those methods with an additional optional path array, that allows us to parse through nested web components.

`Page.getParent` is an internal helper function to get a reference to the parent of the requested element. This helper function is used in both: `Page.querySelector` and `Page.querySelectorAll`.

`PageElement.querySelector` is a new method that lets us retrieve an element from a wrapped node. 

Example - get input to enter a new todo:
```
const input = page.querySelector(".new-todo-input", ["todo-app", "todo-topbar"]);
    input.setValue(`Something to do ${i}`);
    input.dispatchEvent("input");
    input.enter("keyup");
}
```

html structure:
```
.
└── body
    └── <todo-app>
        └── shadow-root
            └── <todo-topbar>
                └── shadow-root
                    └── <input class="new-todo-input"/>
```

Example: get all inputs to toggle completed status:
```
const items = page.querySelectorAll("todo-item", ["todo-app", "todo-list"]);
for (let i = 0; i < numberOfItemsToAdd; i++){
    const item = items[i].querySelector(".toggle-todo-input");
    item.click();
}
```

html structure:
```
.
└── body/
    └── <todo-app>/
        └── shadow-root/
            └── <todo-list>/
                └── shadow-root/
                    ├── <todo-item>/
                    │   └── shadow-root/
                    │       └── <input class="toggle-todo-input">
                    ├── <todo-item>/
                    │   └── shadow-root/
                    │       └── <input class="toggle-todo-input">
                    └── <todo-item>/
                        └── shadow-root/
                            └── <input class="toggle-todo-input">
```

usage example: https://github.com/WebKit/Speedometer/pull/249

@kara

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -36,7 +36,7 @@ class Page {
         });
     }
 
-    getParent(path) {
+    _getParent(path) {
         const parent = path.reduce((root, selector) => {
             const node = root.querySelector(selector);
             return node.shadowRoot ?? node;
@@ -45,16 +45,46 @@ class Page {
         return parent;
     }
 
+    /**
+     * Returns the first element within the document that matches the specified selector, or group of selectors.
+     * If no matches are found, null is returned.
+     *
+     * An optional path param is added to be able to target elements within a shadow DOM or nested shadow DOMs.
+     *
+     * @example
+     * // DOM structure: <todo-app> -> #shadow-root -> <todo-list> -> #shadow-root -> <todo-item>
+     * // return PageElement(<todo-item>)
+     * querySelector("todo-item", ["todo-app", "todo-list"]);
+     *
+     * @param {string} selector A string containing one or more selectors to match.
+     * @param {string[]} [path] An array containing a path to the parent element.
+     * @returns PageElement | null
+     */
     querySelector(selector, path = []) {
-        const element = this.getParent(path).querySelector(selector);
+        const element = this._getParent(path).querySelector(selector);
 
         if (element === null)
             return null;
         return this._wrapElement(element);
     }
 
+    /**
+     * Returns all elements within the document that matches the specified selector, or group of selectors.
+     * If no matches are found, null is returned.
+     *
+     * An optional path param is added to be able to target elements within a shadow DOM or nested shadow DOMs.
+     *
+     * @example
+     * // DOM structure: <todo-app> -> #shadow-root -> <todo-list> -> #shadow-root -> <todo-item>
+     * // return [PageElement(<todo-item>), PageElement(<todo-item>)]
+     * querySelectorAll("todo-item", ["todo-app", "todo-list"]);
+     *
+     * @param {string} selector A string containing one or more selectors to match.
+     * @param {string[]} [path] An array containing a path to the parent element.
+     * @returns array
+     */
     querySelectorAll(selector, path = []) {
-        const elements = Array.from(this.getParent(path).querySelectorAll(selector));
+        const elements = Array.from(this._getParent(path).querySelectorAll(selector));
         for (let i = 0; i < elements.length; i++)
             elements[i] = this._wrapElement(elements[i]);
         return elements;
@@ -156,6 +186,13 @@ class PageElement {
         this.#node.dispatchEvent(event);
     }
 
+    /**
+     * Returns the first element found in a node of a PageElement that matches the specified selector, or group of selectors. If a shadow DOM is present in the node, the shadow DOM is used to query.
+     * If no matches are found, null is returned.
+     *
+     * @param {string} selector A string containing one or more selectors to match.
+     * @returns PageElement | null
+     */
     querySelector(selector) {
         const root = this.#node.shadowRoot ?? this.#node;
         const element = root.querySelector(selector);
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -186,16 +186,25 @@ class PageElement {
         this.#node.dispatchEvent(event);
     }
 
+    _getParent(path) {
+        const parent = path.reduce((root, selector) => {
+            const node = root.querySelector(selector);
+            return node.shadowRoot ?? node;
+        }, this.#node);
+
+        return parent;
+    }
+
     /**
      * Returns the first element found in a node of a PageElement that matches the specified selector, or group of selectors. If a shadow DOM is present in the node, the shadow DOM is used to query.
      * If no matches are found, null is returned.
      *
      * @param {string} selector A string containing one or more selectors to match.
+     * @param {string[]} [path] An array containing a path to the parent element.
      * @returns PageElement | null
      */
-    querySelector(selector) {
-        const root = this.#node.shadowRoot ?? this.#node;
-        const element = root.querySelector(selector);
+    querySelector(selector, path = []) {
+        const element = this._getParent(path).querySelector(selector);
 
         if (element === null)
             return null;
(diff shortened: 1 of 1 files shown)


resources/benchmark-runner.mjs
```diff
@@ -36,15 +36,28 @@ class Page {
...
+        const element = this.getParent(path).querySelector(selector);
+
         if (element === null)
             return null;
         return this._wrapElement(element);
     }
 
-    querySelectorAll(selector) {
-        const elements = Array.from(this._frame.contentDocument.querySelectorAll(selector));
+    querySelectorAll(selector, path) {
```
rniwa, line 59: Maybe specify the default value of []?

flashdesignory: I'm checking in the `getParent` function: 

```
if (!path)
            return this._frame.contentDocument;
````

Comment (rniwa) [reply resources/benchmark-runner.mjs:59]: Yeah, I know that but I think specifying [] here and not adding that special case would be better.

Comment (flashdesignory) [reply resources/benchmark-runner.mjs]: ok - added defaults!

**Comment by bgrins** [comment]
This seems like an improvement over the previous need to expose the element inside Shadow DOM onto the global and call into it (i.e. https://github.com/WebKit/Speedometer/pull/249/commits/01fa32d55f6953e158cf2ff55a56dc0593cc0526#diff-2c1c126bd9bb884254da07a36c11788febf37ba69b0afcb3b7a735ca4d03b287R148-R167). @julienw do you have any thoughts on this?

File: resources/benchmark-runner.mjs [hunk 2]
```diff
@@ -36,15 +36,25 @@ class Page {
         });
     }
 
-    querySelector(selector) {
-        const element = this._frame.contentDocument.querySelector(selector);
+    getParent(path) {
```
camillobruni, line 39: `getParent` => `_getParent`
Since this does not wrap the node it should probably not be used externally.

resources/benchmark-runner.mjs
```diff
@@ -145,6 +155,15 @@ class PageElement {
         const event = new contentWindow.MouseEvent(type, eventOptions);
         this.#node.dispatchEvent(event);
     }
+
+    querySelector(selector) {
```
camillobruni, line 159: I guess this is accidental left-over?  (duplicate method name :) )

julienw:
> This seems like an improvement over the previous need to expose the element inside Shadow DOM onto the global and call into it (i.e. [01fa32d#diff-2c1c126bd9bb884254da07a36c11788febf37ba69b0afcb3b7a735ca4d03b287R148-R167](https://github.com/WebKit/Speedometer/commit/01fa32d55f6953e158cf2ff55a56dc0593cc0526#diff-2c1c126bd9bb884254da07a36c11788febf37ba69b0afcb3b7a735ca4d03b287R148-R167)). @julienw do you have any thoughts on this?

Yeah this looks reasonable to me too!

resources/benchmark-runner.mjs
flashdesignory: done

resources/benchmark-runner.mjs
```diff
@@ -36,15 +36,25 @@ class Page {
...
+        const element = this.getParent(path).querySelector(selector);
+
         if (element === null)
             return null;
         return this._wrapElement(element);
     }
 
-    querySelectorAll(selector) {
-        const elements = Array.from(this._frame.contentDocument.querySelectorAll(selector));
+    querySelectorAll(selector, path = []) {
```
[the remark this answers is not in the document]
flashdesignory: done

resources/benchmark-runner.mjs
```diff
@@ -145,6 +155,15 @@ class PageElement {
         const event = new contentWindow.MouseEvent(type, eventOptions);
         this.#node.dispatchEvent(event);
     }
+
+    querySelector(selector) {
+        const root = this.#node.shadowRoot ?? this.#node;
```
julienw, line 197: if this node has a shadow root, how would you access children for this node then? I'd use an optional boolean parameter "useShadowRoot" to decide. This is a bit like `getComputedStyle` taking a parameter for pseudo elements.

julienw: Or maybe Page's `querySelector` and `querySelectorAll` (and `getParent`) should be moved here, with Page's version of them defering to PageElement.

flashdesignory: @julienw - I made some updates - let me know if that's more what you had in mind

resources/benchmark-runner.mjs
```diff
@@ -145,6 +185,31 @@ class PageElement {
...
     }
+
+    _getParent(path) {
+        const parent = path.reduce((root, selector) => {
+            const node = root.querySelector(selector);
+            return node.shadowRoot ?? node;
+        }, this.#node.shadowRoot ?? this.#node);
+
+        return parent;
+    }
```
camillobruni, line 196: nit: Could you move this to a separate shared helper function that also takes a `lookupStartNode` or so?

camillobruni (approved):

resources/benchmark-runner.mjs
flashdesignory: done

resources/benchmark-runner.mjs
julienw: My question was more: what happens if this element has both a shadowRoot and children? How do we decide if we go down the shadowRoot path or the in-the-main-document path? Maybe we really want 2 different functions here: querySelector vs querySelectorInShadowRoot?

But this is theoretical so we might as well decide when/if we need it later. What do the others think?

flashdesignory: I see what you mean. I am open to rename and keeping the original Page.querySelector / Page.querySelectorAll methods if that's preferred.

bgrins: I see what you are saying - at this point it's not possible to querySelector for a direct descendant if the node has a shadow root. I agree we could punt on this and figure it out when it's needed, but I also wonder if the existence of a non-empty path could be used as semantics to imply that you want to use shadow root.

So if you wanted to jump between shadow and non shadow you could first get to the element who's direct descendant you care about using a querySelector on the document with the path (which will prefer the shadow root). Then take a reference to that element and querySelector on that without a path (which will use the element and not the root). Then if you need to go further into a shadow tree you can querySelector again with a path. Or we could just create new methods.

flashdesignory: As soon as you want to query a wrapped node, you'd need to use PageElement.querySelector, which still prefers the shadowRoot. How about this for the PageElement methods?

```
querySelectorInShadowRoot(selector, path = []) {
        const lookupStartNode = this.#node.shadowRoot ?? this.#node;
        const element = getParent(lookupStartNode, path).querySelector(selector);

        if (element === null)
            return null;
        return new PageElement(element);
    }

    querySelector(selector) {
        const element = this.#node.querySelector(selector);

        if (element === null)
            return null;
        return new PageElement(element);
    }

```

bgrins: I was suggesting we use the existence of a non-empty path to in the same manner as for the document methods (prefer shadow root if there is >= 1 element in the array, and the element if not). But I really don't have a strong opinion on the API shape until if/when we need it

bgrins (approved):
This seems like a fine way to unblock progress on two PRs. I suggest we land it and we can track improvements in follow ups. @rniwa WDYT?

rniwa (approved):

Issue #112, opened 2023-03-16
TodoMVC workloads - A11y audit


flashdesignory:
While updating the TodoMVC applications, we noticed a few a11y related issues. 
For example inconsistency in using inputs with and without labels, or using inputs without a form and therefore having to manage keyboard events more extensively. 

These inconsistencies don't interfere with the test results, nor do they speed up the workloads once addressed most likely.
 
That said, I believe that since our workloads are supposed to represent apps on the web, we should also advocate for best practices where it makes sense. This will hopefully also encourage other engineers to follow similar patterns when developing their own applications.

rniwa:
I think this is a blocker to release v3.

PR #254, opened 2023-06-22
[OPTIONAL] - updated todomvc styles

flashdesignory:
This pr adds updated styles for todo apps as discussed in the a11y audit: https://docs.google.com/document/d/1oQiN56elDw21EejEY7H7aT6CSJBMDP-9XVjSb6d6B-I/edit

This is also shareable and can be consumed by the lit / web component prs for testing if we want (web component pr basically uses identical styles / structure already). 

Folder structure matches the newssite-css setup and the package can be locally installed the same way.

@kara


bgrins:
Since this shows up as net-new code in GitHub it's hard to differentiate what is part of #112 and what's just part of bringing over the existing CSS. If it's not too much trouble can you either have two separate commits or two separate PRs where first we bring the whole set over unchanged, and then second it applies the a11y improvements?

flashdesignory:
That is gonna take some time.. Since the dom structure also changed, even with two prs it'll look pretty messy still.

flashdesignory:
What makes it hard to compare against the currently used npm package is that all styles are separated into separate files in this pr (to allow us to build the css modules)

bgrins:
> Since the dom structure also changed

Where did this happen, I don't see it in this PR?

flashdesignory:
This pr is just styles, but there are some screenshots in the document mentioned above.

bgrins:
Clarification after chatting with @flashdesignory - this CSS is targeting the new markup structure https://github.com/WebKit/Speedometer/pull/254/files#diff-c9c4e4297e9dc33905dec9420bb822198a2bd333a48d6231cc23e0df35a5eec7 which currently only the proposed Lit and Web Components workloads use. The idea would be that as other workloads migrate their markup to the new structure they would start using this local package instead of the current CSS.

flashdesignory:
Additional clarification: only the web-component version uses the new structure, but it's something that lit could potentially implement to take advantage of the css modules here.

.eslintignore
```diff
@@ -20,6 +20,8 @@ package-lock.json
 !/resources/todomvc/vanilla-examples/javascript-es6
 !/resources/todomvc/vanilla-examples/javascript-es6-webpack
 
+!resources/todomvc/todomvc-css
```
julienw, line 23: Are these lines necessary for the code to be checked by eslint? I think they're useful if there's another line that adds an ignore that matches it, but I don't think there's any for this one.

(same comment for prettierignore below)

resources/todomvc/todomvc-css/CHANGES.md
```diff
@@ -0,0 +1,109 @@
+# Changes
```
julienw, line 1: It's not 100% clear to me if you'll land this file and the images with this PR?

julienw:
This looks reasonable to me, but I didn't look too closely.

bgrins:
After some more discussion - my main note here is that it assumes a different DOM structure for the general styling so it's not a straightforward migration for all the workloads to use it - which would be a good outcome compared to having two copies of the CSS and needing to reason about when each one should be used.

So I'd like to double check if we have any blockers or issues to creating a path forward where all workloads are using it (presumably many using just the normal CSS files which are either built into a bundle or using @import). This could be follow up work and IMO can continue into the next half since it's not new workloads or major changes to existing (and AIUI the structure changes are meant as part of the a11y improvements).

One example I thought about is if we introduce a new DOM structure there may be some interdependencies with the complex DOM and codegen work. We confirmed that the refactoring in https://github.com/WebKit/Speedometer/pull/246 should make it easy to adapt to a new structure so long as we remember to re-run it.

flashdesignory:
No blockers from my side and I'd be down to make the changes. Being able to work on these in the next quarter would be ideal.

bgrins (approved):

rniwa (approved):

PR #256, opened 2023-06-23
[OPTIONAL] Avoid using generators in TimeSeries data model.

rniwa:
This avoids pathological perf hit in Firefox.


rniwa:
Hm... this doesn't seem to affect the score in Firefox at all so not sure if it's worth merging.

bgrins:
> Hm... this doesn't seem to affect the score in Firefox at all so not sure if it's worth merging.

Interesting, we'll take a look at it. Thanks for pushing it up

mstange:
This PR does not re-generate `resources/tentative/perf.webkit.org/public/v3/bundled-scripts.js` , so would we expect to be seeing a perf difference with it?

mstange:
Thanks! I still get errors in the console though, there's one remaining instance at the end of `_renderTimeSeries`.

rniwa:
> Thanks! I still get errors in the console though, there's one remaining instance at the end of `_renderTimeSeries`.

Fixed that. Hm... now Firefox is back to being slow again.

mstange:
Looks good! On my machine this improves the Firefox numbers by around 3%, which is the amount we expected.

rniwa:
> Looks good! On my machine this improves the Firefox numbers by around 3%, which is the amount we expected.

Hm... Is that difference big enough that we want this change? On my 16" M2 MacBookPro, Chrome gets a score around 30 and Safari gets score around 25 whereas Firefox gets around 10 for this subtest. That seems to indicate to me whatever primary perf issue Firefox is encountering here isn't really generators.

All things being equal, I'd rather keep the original code to use generators. I'm open to take this change if it was the primary reason Firefox is lagging behind other browsers because testing generators isn't the main point of this test, but generators is what I originally wrote and runs on production site: https://perf.webkit.org/ so there is some validity in keeping generators.

mstange:
The 3% was on macOS, where Firefox's canvas implementation is the main bottleneck.
On Windows, this PR improves Firefox's score by 10%.

bgrins:
> Hm... Is that difference big enough that we want this change? On my 16" M2 MacBookPro, Chrome gets a score around 30 and Safari gets score around 25 whereas Firefox gets around 10 for this subtest. That seems to indicate to me whatever primary perf issue Firefox is encountering here isn't really generators.

> All things being equal, I'd rather keep the original code to use generators. I'm open to take this change if it was the primary reason Firefox is lagging behind other browsers because testing generators isn't the main point of this test, but generators is what I originally wrote and runs on production site: https://perf.webkit.org/ so there is some validity in keeping generators.

The key question isn't the relative improvement in Firefox, it's still https://github.com/WebKit/Speedometer/pull/244#issuecomment-1599248127. Is it appropriate for engines to be driving resources to optimize generators instead of other features (keeping in mind the bottom line goal `When a browser improves its score on the benchmark, actual users should benefit`)? Do we have data to indicate that they are in wide use and/or that performance is currently a limiting factor towards broader adoption of the feature?

julienw:
The most common and interesting places where I've seen generators is with coroutines. 2 quite different examples:
With requestAnimationFrame: http://js-coroutines.com/docs/coroutines.js.html#line199
-> This calls the coroutine once for every frame.

With requestIdleCallback: http://js-coroutines.com/docs/coroutines.js.html#line94
-> This calls the coroutine repeatidly until the idle budget expires.

More explanations about them: https://dev.to/miketalbot/60fps-javascript-while-stringfying-and-parsing-100mbs-of-json-84l

For both these cases I don't think there's an incentive in making generators fast in a hot loop. Especially the fact that yield and next machinery is fast isn't interesting. But there's a point in having the generator functions themselves fast, that is making sure that the engines can actually optimize the work that happens in these functions, for example the JSON-parsing or gz-encoding work in a coroutine. To ensure this it would probably be better to use directly js-coroutines in a separate benchmark and make use of its already-existing functions, and call these ones in a loop.

Hope it makes sense.

bgrins:
We discussed this in a meeting - there was no strong data or argument for keeping it in so we'll plan to merge this and move the workload itself out of tentative. We can always add it later on if that changes.

flashdesignory (approved):

PR #258, opened 2023-06-23
Use more varied set of todo text rather than `Something to do ${i}`

flashdesignory:
chrome before:
before_chrome

chrome after:
after_chrome

safari before:
before_apple

safari after:
after_safari

firefox before:
before_firefox

firefox after:
after_firefox

@kara


resources/translations.mjs
```diff
@@ -0,0 +1,108 @@
...
+        " Intelligent Fresh Mouse",
+        " Tasty Granite Gloves",
+        " Small Steel Shirt",
+        " Refined Concrete Computer",
+        " Sleek Frozen Shirt",
+        " Intelligent Concrete Shoes",
+        " Handmade Rubber Car",
+        " Sleek Cotton Towels",
+        " Unbranded Concrete Hat",
+        " Incredible Plastic Fish",
```
bgrins, line 57: Is the space in front of each string intentional?

rniwa: Yeah, these leading whitespace seems unnecessary.

flashdesignory: removed!

bgrins (approved):
I think this PR can be renamed to something like "Use more varied set of todo text rather than `Something to do ${i}`" (i.e. not refer to this as translating as-is), but I think the idea makes sense

rniwa (approved):

PR #266, opened 2023-07-01
[Optional] - News Site support Forced Colors / High Contrast Mode

flashdesignory:
News site css updates, to support forced colors / high contrast mode. 

Default styling:
Screenshot 2023-07-01 at 12 05 21 PM

High Contrast Mode:
Screenshot 2023-07-01 at 12 05 32 PM

@lpardosixtosMs - I think that's what you were looking for.
There's more I can address, but this feels like a good starting point. 

@kara


resources/newssite/news-site-css/src/css/a11y.css
```diff
@@ -41,6 +43,53 @@ but useful for people who use screen readers */
     }
 }
 
+/* High Contrast / Forced-Colors:
+  Apply forced-colors class to parent element
+  to manually envoke force-colors on child elements.
```
julienw, line 48: I'm not sure I understand this comment...

julienw:
Can you please explain what the goal for this patch is in the context of speedometer 3? What are we trying to measure?

resources/newssite/news-site-css/src/css/a11y.css
flashdesignory: removed comment, which was outdated and confusing.

flashdesignory:
> Can you please explain what the goal for this patch is in the context of speedometer 3? What are we trying to measure?

The goal of this pr is to support a high contrast setting / mode, which allows a user to change the color scheme system wide. In that case, I am re-applying styles with the `@media (forced-colors: active)` media query to keep the news sites usable. 

My main motivation was to use a styling system that supports as many use-cases as possible. 
This also originated from a request from @lpardosixtosMs to use some css custom properties with fallback values. 

Scores will come shortly

flashdesignory:
Chrome before:
before_chrome

Chrome after:
after_chrome

Safari before:
before_safari

Safari after:
after_safari

Firefox before
before_firefox

Firefox after:
after_firefox

edge before:
before_edge

edge after:
after_edge

julienw:
I'm sorry but it's still not clear to me what this brings in the context of speedometer 3. What do we want to measure?

flashdesignory:
> I'm sorry but it's still not clear to me what this brings in the context of speedometer 3. What do we want to measure?

It's an optional update that doesn't affect the scores as far as I can tell. 
we are being good citizens to supply styles that break "less" and similar to the todo a11y audit, we are trying to create workloads that are as close as possible to real world applications. A typical news site would (I hope) use styles that support high contrast modes for readers that wouldn't be able to use the site otherwise.

I assume there's something else @lpardosixtosMs is looking for to measure maybe, but I can't speak for him.

lpardosixtosMs:
@julienw I initially brought this up as a use-case to squeeze some css variable fallbacks into Speedometer, like they do [here](https://github.com/adobe/spectrum-css/blob/main/components/progressbar/index.css) with high-contrast. This implementation has some CSS variable fallbacks, but they won't be resolved on Speedometer runtime. Not exactly what I wanted to measure, but I'm happy to see the accessibility issue addressed, given that this should be a realistic page.

flashdesignory:
closing this for now, will open new pr at a later point

PR #272, opened 2023-07-06
News site - add translation data

flashdesignory:
This pr adds data files for Japanese and Arabic to both the next.js and nuxt version. 
The data files are made available, but are currently NOT used in the test suites.

To preview locally, run` npm run dev` and navigate to these urls:

http://127.0.0.1:7000/resources/newssite/news-next/dist/index.html
http://127.0.0.1:7000/resources/newssite/news-next/dist/index.html?lang=jp
http://127.0.0.1:7000/resources/newssite/news-next/dist/index.html?lang=ar

http://127.0.0.1:7000/resources/newssite/news-nuxt/dist/index.html
http://127.0.0.1:7000/resources/newssite/news-nuxt/dist/index.html?lang=jp
http://127.0.0.1:7000/resources/newssite/news-nuxt/dist/index.html?lang=ar

Note: The number of files is higher, since both apps needed to get built with the data files included. 

@kara


resources/newssite/news-next/src/context/data-context.js
```diff
@@ -1,9 +1,24 @@
...
+import { footer as footerJp } from "@/data/jp/footer";
+import { footer as footerAr } from "@/data/ar/footer";
+
 import * as buttonsEn from "@/data/en/buttons";
+import * as buttonsJp from "@/data/jp/buttons";
+import * as buttonsAr from "@/data/ar/buttons";
+
 import * as linksEn from "@/data/en/links";
+import * as linksJp from "@/data/jp/links";
+import * as linksAr from "@/data/ar/links";
```
julienw, line 21: It is probably not scalable to import all the localized content like this because then everything ends up in the bundle, instead we probably want to use something like async imports or `fetch`. But maybe this is good enogh for us...

resources/newssite/news-next/src/context/data-context.js
```diff
@@ -13,6 +28,20 @@ const strings = {
...
         links: linksEn,
     },
+    jp: {
+        content: contentJp,
+        settings: settingsJp,
+        footer: footerJp,
+        buttons: buttonsJp,
+        links: linksJp,
+    },
+    ar: {
```
julienw, line 38: can we do what I was suggesting earlier? Automatically using `rtl` if `ar` is specified?
As a reminder we're using this list in my project: https://github.com/firefox-devtools/profiler/blob/main/src/app-logic/l10n.js#L57
but I'd be happy enough if you shortlist that to just `ar` currently.

(both for next and nuxt)

julienw (approved):
Thanks, this looks mostly good to me! Just a request to automatically set `rtl` from the locale please :-)

resources/newssite/news-next/src/context/data-context.js
flashdesignory: yes, it's not for production usage 😄 
I wanted to avoid any fetch of content and for our purpose it works.

resources/newssite/news-next/src/context/data-context.js
flashdesignory: Happy to add

julienw:
Looks good to me now!

flashdesignory:
@rniwa - does this pr work for you?

rniwa (approved):
This seems reasonable. I didn't check the translations for any offensive words / phrases, however

PR #289, opened 2023-07-21
Simplify the score calculation by replacing 60 / 50 by 1.

rniwa:
This makes the score runs/s, which is pretty neat.


camillobruni (approved):

bgrins (approved):
Approving as discussed in the meeting. If we want to re-add a correction factor later we can.

PR #293, opened 2023-08-09
Update about text

bgrins:


bgrins:
We've discussed also linkifying the name of each workload to the corresponding readme (happy to do that in a follow up once we're aligned on the text)

flashdesignory (approved):

rniwa (approved):

PR #296, opened 2023-08-18
Fix :last-child selector in Web Components and LIT

lpardosixtosMs:
Currently the web components CSS uses  the `.todo-item:last-child` selector, which matches all the `.todo-item` elements since each is the only child of its host. Change by `host(:last-child) > .todo-item`. Similarly, replace `.editing:last-child` with `host(:last-child) .editing` in the LIT todoMVC CSS.


lpardosixtosMs:
@flashdesignory Can you take a look? Thanks!

flashdesignory (approved):
approved, assuming it was also tested in the standalone versions.

lpardosixtosMs:
@flashdesignory would you mind merging it? I don't have write access.

> approved, assuming it was also tested in the standalone versions.

Yes, this is a fix mainly for standalone versions:

Before:

image

After:

image

PR #297, opened 2023-08-18
Remove benchmark-reporter.mjs

camillobruni:
I think we can remove this file (I'm not sure if anybody still depends on it).
From what it looks it is unused unless `?webkit` is passed, but even then, it seems to remove the `main` element which contains the runner window.


camillobruni:
@rniwa  is the plan to reactive this? (I'm not sure how you folks run this on your perf infra)

rniwa (approved):
It's fine to remove this in this repo. We'd likely need to add it back on WebKit side but that's okay.

bgrins (approved):
AFAICT we don't use this in CI

PR #310, opened 2023-09-06
Add complex DOM workloads to tests.mjs disabled by default

lpardosixtosMs:
This PR:

- Adds all the complex DOM workloads to `tests.mjs` under the `complex` tag.
- Sets the disabled flag to `true` for all complex-DOM variations except React-Complex-DOM.
- Adds a new tag `complex-default` for the workloads that the group agreed should eventually be enabled by default.
- Makes a small logic change to not add the `default` tag for the disabled tests.

[Hosted version](https://lpardosixtosms.github.io/Speedometer)

### DIFF
--- a/resources/developer-mode.mjs
+++ b/resources/developer-mode.mjs
@@ -186,7 +186,7 @@ function updateURL() {
     // to comma separate only the selected
     const selectedSuites = Suites.filter((suite) => !suite.disabled);
 
-    if (!selectedSuites.length || selectedSuites.length === Suites.length) {
+    if (!selectedSuites.length) {
         url.searchParams.delete("tags");
         url.searchParams.delete("suites");
         url.searchParams.delete("suite");
@@ -202,7 +202,11 @@ function updateURL() {
                 commonTags = new Set(suite.tags.filter((tag) => commonTags.has(tag)));
         }
         if (commonTags.size) {
-            url.searchParams.set("tags", [...commonTags][0]);
+            const tags = [...commonTags][0];
+            if (tags === "default")
+                url.searchParams.delete("tags");
+            else
+                url.searchParams.set("tags", tags);
             url.searchParams.delete("suites");
         } else {
             url.searchParams.delete("tags");
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -449,6 +449,337 @@ Suites.push({
 });
 
 Suites.push({
+    name: "TodoMVC-JavaScript-ES5-Complex-DOM",
+    url: "todomvc/vanilla-examples/javascript-es5-complex/dist/index.html",
+    tags: ["todomvc"],
+    async prepare(page) {
+        (await page.waitForElement(".new-todo")).focus();
+    },
+    tests: [
+        new BenchmarkTestStep(`Adding${numberOfItemsToAdd}Items`, (page) => {
+            const newTodo = page.querySelector(".new-todo");
+            for (let i = 0; i < numberOfItemsToAdd; i++) {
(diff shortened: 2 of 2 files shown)


rniwa:
There seems to be a slight bug in the developer menu. You probably want to apply a change like this:
```diff
diff --git a/resources/developer-mode.mjs b/resources/developer-mode.mjs
index d2daa31..65a800d 100644
--- a/resources/developer-mode.mjs
+++ b/resources/developer-mode.mjs
@@ -186,7 +186,7 @@ function updateURL() {
     // to comma separate only the selected
     const selectedSuites = Suites.filter((suite) => !suite.disabled);
 
-    if (!selectedSuites.length || selectedSuites.length === Suites.length) {
+    if (!selectedSuites.length) {
         url.searchParams.delete("tags");
         url.searchParams.delete("suites");
         url.searchParams.delete("suite");
@@ -202,7 +202,11 @@ function updateURL() {
                 commonTags = new Set(suite.tags.filter((tag) => commonTags.has(tag)));
         }
         if (commonTags.size) {
-            url.searchParams.set("tags", [...commonTags][0]);
+          const tags = [...commonTags][0];
+            if (tags == "default")
+              url.searchParams.delete("tags");
+            else
+              url.searchParams.set("tags", tags);
             url.searchParams.delete("suites");
         } else {
             url.searchParams.delete("tags");

```

rniwa (approved):
r=me with the fix to `developer-mode.mjs`.

lpardosixtosMs:
> There seems to be a slight bug in the developer menu. You probably want to apply a change like this:
> 
> ```diff
> diff --git a/resources/developer-mode.mjs b/resources/developer-mode.mjs
> index d2daa31..65a800d 100644
> --- a/resources/developer-mode.mjs
> +++ b/resources/developer-mode.mjs
> @@ -186,7 +186,7 @@ function updateURL() {
>      // to comma separate only the selected
>      const selectedSuites = Suites.filter((suite) => !suite.disabled);
>  
> -    if (!selectedSuites.length || selectedSuites.length === Suites.length) {
> +    if (!selectedSuites.length) {
>          url.searchParams.delete("tags");
>          url.searchParams.delete("suites");
>          url.searchParams.delete("suite");
> @@ -202,7 +202,11 @@ function updateURL() {
>                  commonTags = new Set(suite.tags.filter((tag) => commonTags.has(tag)));
>          }
>          if (commonTags.size) {
> -            url.searchParams.set("tags", [...commonTags][0]);
> +          const tags = [...commonTags][0];
> +            if (tags == "default")
> +              url.searchParams.delete("tags");
> +            else
> +              url.searchParams.set("tags", tags);
>              url.searchParams.delete("suites");
>          } else {
>              url.searchParams.delete("tags");
> ```

@rniwa Happy to do the changes, but could you explain what is the bug? I'm not seeing anything unexpected.

rniwa:
> @rniwa Happy to do the changes, but could you explain what is the bug? I'm not seeing anything unexpected.

If you select all tests, it clears the param so when you reload the page, it goes back to the default configuration. We should be using tags=all instead in that scenario. Also, when the default set of tests are selected, we add tags=default instead of removing tags.

lpardosixtosMs:
> > @rniwa Happy to do the changes, but could you explain what is the bug? I'm not seeing anything unexpected.
> 
> If you select all tests, it clears the param so when you reload the page, it goes back to the default configuration. We should be using tags=all instead in that scenario. Also, when the default set of tests are selected, we add tags=default instead of removing tags.

Got it, thanks!

PR #311, opened 2023-09-11
Set proper unit for "Score" Metric

camillobruni:
- Use "score" as unit
- Format benchmark-runner.mjs


julienw:
The formatting brings a conflict, it might be easier to reformat in a separate patch.

julienw (approved):
Is it true to say that it's used in the json only? I don't see where it's used otherwise.
If it's in the json only this works for me, otherwise I think I'd use the empty string instead.

rniwa (approved):

camillobruni:
yes this is only affecting the json

Issue #318, opened 2023-10-02
In news site, `.truncate-multiline` uses deprecated `-webkit-box`


rniwa:
Right now, both news site tests use the following CSS rule:

```css
.truncate-multiline {
    display: -webkit-box !important;
    -webkit-line-clamp: 2;
    -webkit-box-orient: vertical;
    width: 100%;
    overflow:hidden
}
```

This would result in the deprecated flexbox layout to kick in. I don't think that's appropriate in a supposedly cross-browser benchmark.

rniwa:
@ther

rniwa:
cc @flashdesignory

flashdesignory:
looking at other workloads, there are many uses of -webkit specific css rules. 
Are we removing all of those too? 

https://github.com/search?q=repo%3AWebKit%2FSpeedometer%20-webkit&type=code

rniwa:
Some of those are fine because they also specify the non-prefixed versions but ones that only have effect in WebKit/Blink should probably be removed.

rniwa:
This deprecated flexbox thing is particularly bad because this stuff is known to be buggy & slow.

flashdesignory:
It is used in the wild... here are two examples I found quickly:

<img width="1695" alt="Screenshot 2023-10-02 at 7 00 31 PM" src="https://github.com/WebKit/Speedometer/assets/372973/78493683-3d2a-4993-86e2-25c67d155cca">

<img width="1724" alt="Screenshot 2023-10-02 at 7 00 42 PM" src="https://github.com/WebKit/Speedometer/assets/372973/14abb3bf-cc28-4c65-9c40-8f7470bb05d7">

rniwa:
> It is used in the wild... here are two examples I found quickly:

I'm sure it's used in wild. That doesn't mean we should be using it in Speedometer content.

bgrins:
> This would result in the deprecated flexbox layout to kick in. I don't think that's appropriate in a supposedly cross-browser benchmark.

...

> but ones that only have effect in WebKit/Blink should probably be removed.

This technique works in Gecko as well: we support -webkit-box and associated properties, so if you open the page in Firefox it has the same multiline clamp + ellipsis UX.

There's a separate question about UX itself and the particular techniques for doing it, but I'm not persuaded this should be removed as a cross browser concern.

rniwa:
[...]

PR #319, opened 2023-10-02
Removed .truncate-multiline rule which used deprecated flexbox in WebKit/Blink from news sites.

rniwa:
Fixes the issue #318.


rniwa:
No substantive change to the scores before/after the change:

Chrome
Before: 13.4 ± 0.19
After: 13.4 ± 0.49

Firefox
Before: 12.4 ± 0.41
After: 12.6 ± 0.47

Safari
Before: 10.5 ± 0.38
After: 10.4 ± 0.28

[remove-truncate-multiline-rule.zip](https://github.com/WebKit/Speedometer/files/12786771/remove-truncate-multiline-rule.zip)

flashdesignory:
the css classes do have a purpose of truncating text that's too long. 
Simply removing the rule makes the concern mention in the issue go away, but alters the workload. 

What's a cross browser css rule to truncate text (one line and multi-line) from your perspective?

rniwa:
Probably https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

flashdesignory:
> Probably https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

that works for a single line, but not for multi-line... it's used in "truncate-singleline" already.

flashdesignory:
Just to make sure - I am totally fine with replacing the css rule with a more appropriate rule for multi-line truncation. 
I was not able to find a CSS solution that results in the same display as the one in question here.

rniwa:
Yeah, I don't think there is a cross-browser way to do it, in which case, we should just get rid of the rule.

flashdesignory:
Curious what other folks think... 
I definitely don't want to remove the CSS rules, since that changes how the site behaves. 

A minimum would be something like below at least. 
We'd loose the elipsis, but at least the height of the elements stays consistent. 

```
.truncate-multiline {
    max-height: calc(2 * var(--font-size-xlarge));
    display: flex;
    overflow: hidden;
}

.truncate-multiline-3 {
    max-height: calc(3 * var(--font-size-xlarge));
}
```

rniwa:
> > Yeah, I don't think there is a cross-browser way to do it, in which case, we should just get rid of the rule.
> 
> I commented in [#318 (comment)](https://github.com/WebKit/Speedometer/issues/318#issuecomment-1745754725) as well, but this has been supported in Firefox for ~5 years https://bugzilla.mozilla.org/show_bug.cgi?id=WebKit-line-clamp

Oh, interesting. I wasn't aware of that. I guess we can keep it in this case.

PR #323, opened 2023-10-04
Add warmup and step delay to developer UI

camillobruni:
Drive-by-fix: 
- Move the more frequently used iteration count to the top
- Move warmupSuite and warmup-delay closer together


resources/developer-mode.mjs
```diff
@@ -42,7 +46,33 @@ export function createUIForMeasurementMethod() {
     return label;
 }
 
-export function createUIForWarmupSuite() {
+
+function createUIForIterationCount() {
```
rniwa, line 50: Can we not move this function around? The diff is harder to read because of it.

julienw: or at least do this in separate commits

camillobruni:
Please have a look again.

camillobruni:
Also updated the slider UI:
Screenshot 2023-11-24 at 18 26 06

resources/developer-mode.mjs
```diff
@@ -15,11 +15,16 @@ export function createDeveloperModeContainer() {
     content.append(createUIForSuites());
 
     content.append(document.createElement("hr"));
+    content.append(createUIForIterationCount());
+    content.append(document.createElement("br"));
```
julienw, line 19: (optional for this patch)
Instead of these br elements, can't we just use some CSS margin?

Actually I looked with the devtools, they don't seem useful at all. Am I missing something?

resources/developer-mode.mjs
```diff
@@ -63,33 +68,56 @@ export function createUIForWarmupSuite() {
...
 
+function createTimeRangeUI(labelText, initialValue, unit = "ms", min = 0, max = 1000) {
+    const range = document.createElement("input");
+    range.type = "range";
+    range.min = min;
+    range.max = max;
+    range.value = initialValue;
+
+    const label = document.createElement("label");
+    const rangeLabel = document.createElement("span");
```
julienw, line 106: nit: the name "label" can be confusing in this context, I believe it should be renamed, for example to `rangeValueText`?

resources/developer-mode.mjs
```diff
@@ -246,6 +274,16 @@ function updateURL() {
...
+    if (params.warmupBeforeSync !== defaultParams.warmupBeforeSync)
+        url.searchParams.set("warmupBeforeSync", params.warmupBeforeSync);
+    else
+        url.searchParams.delete("warmupBeforeSync");
+
+    if (params.waitBeforeSync !== defaultParams.waitBeforeSync)
+        url.searchParams.set("waitBeforeSync", params.waitBeforeSync);
+    else
+        url.searchParams.delete("waitBeforeSync");
+
```
julienw, line 286: That's 4 times the same code, what do you think of replacing these lines with something such as:
```js
for (const paramKey of [
  'iterationCount',
  'useWarmupSuite',
  'warmupBeforeSync',
  'waitBeforeSync',
]) {
  if (params[paramKey] !== defaultParams[paramKey]) {
    url.searchParams.set(paramKey, params[paramKey]);
  } else {
    url.searchParams.delete(paramKey);
  }
}
```
This would also avoid typos when repeating the param keys.

resources/developer-mode.mjs
```diff
@@ -63,33 +68,56 @@ export function createUIForWarmupSuite() {
...
-    range.oninput = () => {
-        countLabel.textContent = range.value;
+function createUIForIterationCount() {
+    const { range, label } = createTimeRangeUI("Iterations: ", params.iterationCount, "#", 1, 100);
+    range.onchange = () => {
+        params.iterationCount = parseInt(range.value);
+        updateURL();
     };
+    return label;
+}
```
julienw, line 78: I feel like we could factorize a bit more these UI creations, but I'm not sure this would actually be easier to read. So I'm fine with keeping them this way.

resources/developer-mode.mjs
julienw: I also look at using `display: flex` or `inline-flex` to avoid the `float: right` you're adding below (`float` isn't great for layout with all its quirks, now that we can use `flex`), but there seems to be more changes needed for that, so let's defer this for a future PR (I might have a look myself).

resources/developer-mode.mjs
julienw: I wonder if `rangeLabel` should be outside of the `label` element (it's not labeling the input), but this would need more CSS changes that we're willing to do here, I believe.

resources/developer-mode.mjs
```diff
@@ -63,33 +68,56 @@ export function createUIForWarmupSuite() {
...
-
-    let label = document.createElement("label");
-    let countLabel = document.createElement("span");
-    countLabel.textContent = params.iterationCount;
-    label.append("iterations: ", countLabel, document.createElement("br"), range);
-
-    range.oninput = () => {
-        countLabel.textContent = range.value;
+function createUIForIterationCount() {
+    const { range, label } = createTimeRangeUI("Iterations: ", params.iterationCount, "#", 1, 100);
```
julienw, line 72: can we use `200` as a max value ? I think some folks at Mozilla routinely use `200` for their local tests.

julienw (approved):
Thanks, this seems to work perfectly fine!

Here are a few comments. Only the simple ones are mandatory, I'm leaving the ones that need more work for a future PR.

resources/developer-mode.mjs
camillobruni: done, switched over to flex.

resources/developer-mode.mjs
camillobruni: DONE.

resources/developer-mode.mjs
camillobruni: Renamed the variable to rangeValueAndUnit, let's maybe move the rest for later.

rniwa:
What's the status of this PR? Are we merging it for 3.0?

camillobruni:
I'd be happy to land this if possible, but only if everybody is fine with it :)

rniwa:
Let's land it for 3.0 then :)

PR #324, opened 2023-10-04
Todomvc svelte cleanup

flashdesignory:
Some cleanup items from this pr: https://github.com/tastejs/todomvc/pull/2194
Since the changes are coming directly from the Svelte team, it would be beneficial to align. 

two items I didn't carry over to Speedometer:
1. uuid: I kept our implementation 
2. vite: I didn't touch rollup, which is used here

scores
chrome before:
chrome_before
chrome after:
chrome_after

safari before:
safari_before
safari after:
safari_after

firefox before:
firefox_before
firefox after:
firefox_after

edge before: 
edge_before
edge after:
edge_after


julienw (approved):
Interesting that these simple changes make the app so much faster. Looks like that using idiomatic structures helps svelte compile the app to more efficient code.

Should you regenerate the complex version of svelte too?

Small request: can you add 
```js
watch: {
  clearScreen: false,
}
```
to the rollup config
so that `npm run dev` keeps the server URL?

Thanks :-)

resources/todomvc/architecture-examples/svelte/package.json
```diff
@@ -11,7 +11,7 @@
         "build": "rollup -c",
         "watch": "rollup -c -w",
         "serve": "http-server ./dist -p 7002 -c-1 --cors",
-        "dev": "npm run serve & npm run watch"
+        "dev": "npm run serve & npm run watch --no-watch.clearScreen"
```
flashdesignory, line 14: @julienw - let me know if this works for you

resources/todomvc/architecture-examples/svelte-complex/package-lock.json
```diff
@@ -10,7 +10,6 @@
             "devDependencies": {
                 "big-dom-generator": "file:../../big-dom-generator",
                 "http-server": "^14.1.1",
-                "jsdom": "^22.1.0",
```
flashdesignory, line 13: @lpardosixtosMs - this change happened when I ran npm install, to be able to rebuild the new svelte version.

resources/todomvc/architecture-examples/svelte/package.json
flashdesignory: it threw an error for me 🤷

resources/todomvc/architecture-examples/svelte-complex/package-lock.json
issackjohn: That is okay. We've removed jsdom as a devDependency in package.json. I was using npm ci which doesn't update the lock files. This change is okay though.

resources/todomvc/architecture-examples/svelte/package.json
julienw: very weird because this worked for me:
```diff
diff --git a/resources/todomvc/architecture-examples/svelte/rollup.config.js b/resources/todomvc/architecture-examples/>
index b9bf6c67..4c1f9c26 100644
--- a/resources/todomvc/architecture-examples/svelte/rollup.config.js
+++ b/resources/todomvc/architecture-examples/svelte/rollup.config.js
@@ -12,16 +12,19 @@ const production = !process.env.ROLLUP_WATCH;
 export default {
     input: "src/index.js",
     output: {
         file: "dist/app.js",
         format: "iife",
         sourcemap: true,
         name: "app",
     },
+    watch: {
+        clearScreen: false,
+    },
     plugins: [
         css({
             minify: true,
         }),
         svelte({
             include: "src/**/*.svelte",
         }),
         resolve({
```

resources/todomvc/architecture-examples/svelte-complex/package-lock.json
julienw: This makes me think that we probably want `npm install` and not `npm ci` when updating the complex versions. Indeed we might run into problems when adding dependencies to the big-dom-generator. It would be good to double check.

rniwa (approved):

resources/todomvc/architecture-examples/svelte/package.json
flashdesignory: maybe it took a while for the coffee to kick in, but I moved it in the config and it works of course.

Issue #316, opened 2023-09-27
Using display: none instead of visibility: hidden to hide elements in the complex DOM UI.


lpardosixtosMs:
Filing this issue to follow up on the discussion from the discussion in #308.

@julienw:
> I wonder why you decided to use these rules for the closed expandable items: ![image](https://github.com/WebKit/Speedometer/assets/94007365/83e263a1-c23e-4168-b48d-30bb629371d5) 
> instead of display: none which is IMO more common. Do you have data about that? When debugging with Firefox' devtools, I can see that the invisible elements were still laid out. I wonder if this might be the cause of the extra slowness in Firefox. If this is common on websites I'll all for keeping it, but otherwise I think this should be changed to display: none so that it's not part of the layout (but still part of the restyle).

@lpardosixtosMs:
> This is how the spectrum library is coded. We just used the is-open classname to control whether the item is expanded or not. We can override it if there is consensus that using display: none is better.

Performance-wise we can measure the impact of changing this rule, but that doesn't answer the questions:

- Which one is more realistic?
- Why use one against the other?

Here is a comment from our experts about the latter:
> one reason to not use display: none is that, historically, it's not possible to transition/animate from that value. So, if you wanted to have an opening animation where the element grows in height or something like that you might avoid display: none.

lpardosixtosMs:
I'd like us to rethink the questions above within the context of what the complex DOM workloads intend to accomplish. The complex DOM workloads evaluate the impact of performing seemingly simple small independent actions in a complex web page. In real web pages, most of the current DOM nodes have a display value different from `none`, so they are part of both: the layout and the re-style. I think that changing to `display: none` would take us further from the goal of the workload. Here is data for some websites about how many DOM elements have a computed style display value different from none.


| Website |  % of DOM elements with `display` value other than `none` |
|--------|--------|
| https://www.amazon.com/| 91.93% | 
| https://youtube.com/ | 76.53% |
| https://www.google.com/search?q=v8+javascript+engine | 94.67% |
| https://stackoverflow.com/questions/114543/how-can-i-horizontally-center-an-element | 90.76% |
| https://reddit.com/ | 97.79% |
| https://www.bestbuy.com/ | 88.09% |
| https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise| 91.17% |
| https://github.com/WebKit/Speedometer|  93.01% |
| https://www.bing.com/search?q=javascript+v8+engine | 85.92% | 
[...]

PR #327, opened 2023-10-19
[Complex-Dom] Combine visibility: hidden and display: none for side TreeView hidden elements.

lpardosixtosMs:
Following the discussion on #316 this PR adds a new class name to TreeView elements that should be hidden using `display: none`.

After this change the sidebar has 6004 elements and 2996 of them had the `display: none` value or are descendants of elements with `display: none` value. 

Hosted version: https://lpardosixtosms.github.io/SpeedometerHiddenNone/#home


lpardosixtosMs:
### Before and after results

image


### Comparison to standalone versions

image

resources/todomvc/big-dom-generator/src/components/tree-area.jsx
```diff
@@ -8,10 +9,10 @@ const TreeItem = (props) => {
     const { treeNode, currentDepth } = props;
 
     const isExpandableItem = treeNode.type === "expandableItem";
-    const treeViewItemIsOpen = isExpandableItem && currentDepth < MAX_VISIBLE_TREE_VIEW_ITEM_DEPTH ? "is-open" : "";
+    const treeViewItemIsOpen = isExpandableItem && (currentDepth !== MAX_VISIBLE_TREE_VIEW_ITEM_DEPTH || treeNode.isDisplayNone);
```
julienw, line 12: I don't understand this condition, it would be good to explain the logic of deciding if a node uses is-open or display-none or both.

resources/todomvc/big-dom-generator/src/app.css
```diff
@@ -86,6 +86,14 @@ Importing the CSS from the Spectrum CSS library.
     min-width: 30px;
 }
 
+.display-none {
+    display: none;
+}
+
+.spectrum-TreeView-item.is-open > .spectrum-TreeView {
+    visibility: unset;
+}
```
julienw, line 95: The impact of this CSS class is not clear to me. Which node does this apply to? Can you please add some more comments around that?

julienw (commented):
It's difficult to review without more comments in the code. I left a few comments to outline the parts where I'd need more explanations.
Thanks!

resources/todomvc/big-dom-generator/src/tree-generator.js
```diff
@@ -3,6 +3,39 @@ import { DEFAULT_SEED_FOR_RANDOM_NUMBER_GENERATOR, MAX_GENERATED_DOM_DEPTH, MAX_
...
+const fillSubtreeWeights = (node, expandableItemWeight, nonExpandableItemWeight) => {
+    if (node.type === "expandableItem")
+        node.subTreeWeight = node.children.reduce((acc, child) => acc + fillSubtreeWeights(child, expandableItemWeight, nonExpandableItemWeight), expandableItemWeight);
+    else
+        node.subTreeWeight = nonExpandableItemWeight;
+
+    return node.subTreeWeight;
+};
+
+const markDisplayNoneNodes = (node, expandableItemWeight, nonExpandableItemWeight) => {
```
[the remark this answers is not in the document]
julienw: I see you added a comment in the next commit, sorry I missed it! Thanks!
 It could be expanded with more explanations about the logic, especially an example of the result.

resources/todomvc/big-dom-generator/src/tree-generator.js
```diff
@@ -3,6 +3,39 @@ import { DEFAULT_SEED_FOR_RANDOM_NUMBER_GENERATOR, MAX_GENERATED_DOM_DEPTH, MAX_
 
 const random = new LCG(DEFAULT_SEED_FOR_RANDOM_NUMBER_GENERATOR);
 
+const fillSubtreeWeights = (node, expandableItemWeight, nonExpandableItemWeight) => {
```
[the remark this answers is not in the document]
lpardosixtosMs: Done.

resources/todomvc/big-dom-generator/src/tree-generator.js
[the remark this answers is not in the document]
lpardosixtosMs: Done.

resources/todomvc/big-dom-generator/src/components/tree-area.jsx
lpardosixtosMs: Added this comment "Every expandable TreeItem is open by default unless it is at the MAX_VISIBLE_TREE_VIEW_ITEM_DEPTH threshold. If it is marked as display none, set it as open so the `visibility: hidden` spectrum rule is removed.".

resources/todomvc/big-dom-generator/src/app.css
lpardosixtosMs: Added this comment:
```
/*
The spectrum rule making the open tree view items visible is
.spectrum-TreeView-item.is-open > .spectrum-TreeView {
    visibility: visible;
}
This can cause descendants of closed tree view items to be visible.
We unset here to let it inherit the visibility of its parent.
*/
```

This is an example of the pathological behavior:
image

resources/todomvc/big-dom-generator/src/app.css
```diff
@@ -86,6 +86,22 @@ Importing the CSS from the Spectrum CSS library.
...
 
+.display-none {
+    display: none;
+}
+
+/*
+The spectrum rule making the open tree view items visible is
+.spectrum-TreeView-item.is-open > .spectrum-TreeView {
+    visibility: visible;
+}
```
julienw, line 97: Can you add a few additional lines explaining that the normal TreeView has `visibility: hidden`?

Then explain that `unset` has the same effect as `visible` for the visible nodes, but doesn't have the side effect of showing descendants?

resources/todomvc/big-dom-generator/src/app.css
julienw: I see, but why wasn't this happening before?

resources/todomvc/big-dom-generator/src/components/tree-area.jsx
```diff
@@ -8,10 +9,14 @@ const TreeItem = (props) => {
     const { treeNode, currentDepth } = props;
 
     const isExpandableItem = treeNode.type === "expandableItem";
-    const treeViewItemIsOpen = isExpandableItem && currentDepth < MAX_VISIBLE_TREE_VIEW_ITEM_DEPTH ? "is-open" : "";
+    /**
+     * Every expandable TreeItem is open by default unless it is at the MAX_VISIBLE_TREE_VIEW_ITEM_DEPTH threshold.
+     * If it is marked as display none, set it as open so the `visibility: hidden` spectrum rule is removed.
```
julienw, line 14: IMO this makes things a bit difficult to follow, what do you think about adding `visibility: unset` to the `.display-none` class?
BTW I think we also want to reset `height` to `auto` are we? Not sure this is necessary.

julienw (approved):
The comments help a lot, thank you

I still don't understand everything :D I left a few questions that I'd love to get answers about, but approving still.

flashdesignory (approved):
thanks @julienw for all the questions, that made it easier to somewhat follow along.

resources/todomvc/big-dom-generator/src/app.css
lpardosixtosMs: > I see, but why wasn't this happening before?

Because before, our code was assuming that every tree item under a closed item is also closed, I changed it to only close the items at the threshold level. It makes more sense to me that the visibility of the children depends on the visibility of the parent instead of relying on its own `is-open` value.

resources/todomvc/big-dom-generator/src/components/tree-area.jsx
lpardosixtosMs: This was just a cheap way of removing the `visibility: hidden` class from `.display-none` elements, doing `visibility: unset` inside the `.display-none` rule wouldn't work because the specificity of the original rule is higher. I think it should be fine to let `.display-none` elements have both `visibility: hidden` and `dispaly: none` properties, since `display: none` will take the elements out of the layout. Alternatively, I could use `!important` to unset them, but that seems even more artificial. WDYT?

resources/todomvc/big-dom-generator/src/app.css
julienw: I see, can you write it as a comment too? I assume it's not just your code but spectrum in general assuming this.

resources/todomvc/big-dom-generator/src/components/tree-area.jsx
julienw: I think I'd use !important here, in this context, so that the 2 rules are in the same place. With some comment explaining why this is present.

lpardosixtosMs: Done.

resources/todomvc/big-dom-generator/src/app.css
lpardosixtosMs: Done.

resources/todomvc/big-dom-generator/src/tree-generator.js
```diff
@@ -1,8 +1,58 @@
...
+    return node.subTreeWeight;
+};
+
+/*
+ * Iterate over the exapandableItem nodes in a breadth-first manner until the
+ * sum of weights of the subtrees with root nodes in the queue is less than
+ * the target number of elements we want to have display none. Mark the
+ * nodes in the queue as display none.
+ * Consider the following example with the weights as displayed in the figure
+ * and a 10 as the target of display none elements. The iteration will stop
```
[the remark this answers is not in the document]
lpardosixtosMs: Done.

resources/todomvc/big-dom-generator/src/app.css
lpardosixtosMs: Done.

rniwa (approved):

lpardosixtosMs:
@julienw I addressed all your feedback, I'll go ahead and merge the PR now. Please let me know if there are any more changes you'd like to see in a follow-up.

PR #330, opened 2023-10-25
Add an option to prime the network cache by loading each test suite in an iframe first.

rniwa:


bgrins:
IIRC the server doesn't set cache headers (and if it did, then the reported effect would be at least mitigated by being contained within the first iteration without requiring a first pass like this)

resources/benchmark-runner.mjs
```diff
@@ -346,7 +346,7 @@ export class BenchmarkRunner {
         const iterationEndLabel = "iteration-end";
         for (let i = 0; i < iterationCount; i++) {
             performance.mark(iterationStartLabel);
-            await this._runAllSuites();
+            await this._runAllSuites(!i && params.cacheResources);
```
camillobruni, line 349: How about putting this as explicit separate `"cacheResources"` step with explicit labels?

camillobruni:
I think it would still make sense to have this step for slower network speeds.
We've occasionally seen different scores on v2.1 depending on how slow/fast the network is. 
Hopefully this would limit the effects a bit?

We could give it a shot with [ts_proxy](https://github.com/catchpoint/WebPageTest.tsproxy) or a similar setup?

bgrins:
> I think it would still make sense to have this step for slower network speeds.

But would this have any effect today, since the server isn't sending cache headers? If/when we started caching then any potential benefit to this change shrinks since 9/10 iterations will already be able skip the network. 

And on the margin I worry this change will incentive engines to do too-aggressive per-process/tab caching the first time a page is loaded (because it will improve the score by pulling work out of measured iterations). That's not really a good outcome - lots of pages are just loaded once and it'd be better have that be faster than to speed up potential subsequent load+interactions that may not ever happen.

camillobruni:
On our perf waterfall we measure with a clean profile, which adds additional noise for the first load (and even worse for local runs).
I'd say it's not something we want to cover directly in Speedometer v3, so I'd vouch for limiting these effects.

bgrins:
The most straightforward way to mitigate any network speed side effects would be to update browserbench.org to set cache headers for static content. Without doing that, AFAICT this PR won't have any practical effect other than incentivizing engines to warm up JIT caches etc during page load (which seems like a bad incentive).

bgrins:
Here are a couple of preview URLs with headers for all resources:

`Cache-Control: public, max-age=15552000`: https://with-cache-headers--speedometer-preview.netlify.app/
`Cache-Control: no-store`: https://with-no-store--speedometer-preview.netlify.app

camillobruni:
At least the effects on v2.1 are not mitigate with cache-headers (maybe this is a different issue here?)

The main issue is that on slower-networks you keep the CPU busy for longer (and thus warm), while sometimes on faster networks we drop to the lower clockspeeds in the setTimeout delays.

Let's have this as an option in the developerMode so we can experiment easily with it. Whether or not we enable this by default is another discussion.

rniwa:
The consensus here seems to abandon this PR, and instead pursue adding caching header on browserbench.org instead.

PR #332, opened 2023-10-28
Add the source map for Angular TodoMVC build.

rniwa:


julienw (commented):
This looks good to me, but I'd like @lpardosixtosMs to have a look so that he can give some guidance to update the complex version.
Also I'd like Bas to check there's no issue with the Base as we had in the past (it seems to be it's OK but I'm not confident).

resources/todomvc/architecture-examples/angular/package.json
```diff
@@ -10,7 +10,7 @@
     "scripts": {
         "ng": "ng",
         "dev": "ng serve",
-        "build": "ng build --base-href /resources/todomvc/architecture-examples/angular/dist/",
+        "build": "ng build --source-map=true",
```
flashdesignory, line 13: we should keep the `base-href` in there.

julienw: Do you know what this does? It's not clear to me. If all it does is adding the `<Base>` to the html file, then maybe we don't need it after all?

flashdesignory: I know Angular needs to be able to find the files when running inside Speedometer, but @rniwa also pointed out before that there was an issue when using `<base>` in the html file when running on the webkit infra.

julienw: At least this works without it currently (I made sure to try it locally), so it's probably not needed unless I'm missing something.

bgrins: I remember this requiring manual fixups like #161 in the past (from @rniwa). I can't tell if we've fixed this with this config in the meantime or if it's still requiring manual work for each build (and if so we should, at minimum, document that)

lpardosixtosMs:
> This looks good to me, but I'd like @lpardosixtosMs to have a look so that he can give some guidance to update the complex version. Also I'd like Bas to check there's no issue with the Base as we had in the past (it seems to be it's OK but I'm not confident).

It should be enough to rebuild the complex version, the script will copy the source maps too, @issackjohn is verifying locally. We usually check that the <base> is not in the html too but looks like it is not generated anymore.

resources/todomvc/architecture-examples/angular/package.json
rniwa: Getting rid of `--base-href` gets rid of the base element in the build so I don't think we want to keep that. As far as I can tell, there is no other side effect from this change.

flashdesignory (approved):
LGTM

julienw (approved):

PR #267, opened 2023-07-03
WIP Use sanitize.js script

camillobruni:
Make sure all workloads run with the same removed APIs

PR #336, opened 2023-11-24
Add sanitize script

camillobruni:
This script will eventually replace the benchmark-specifics overrides we already employ for some workloads.

- clamp all setTimeout to 0ms delay to limit effects of CPU throttling
- Make requestIdleCallback and cancelIdleCallback noops
- Set a fixed userAgent string to avoid running potentially browser-specific code

The sanitize.js script is supposed to be included in all benchmarks.


camillobruni:
Using the sanitize script will happen in a separate PR, see #267 as an early example.

camillobruni:
As per meeting discussion, this needs some more detailed investigation on which workloads use which of these APIs.
Happy to land the non-controversial stuff first and set everything up for more common scripts we want to share between workloads.

PR #337, opened 2023-11-29
Keep the parameter 'tags' as an array in the body of the function

julienw:
Previously, we were transforming it into a set. This was giving problems in a later condition, so that no message/debugInfo were defined, and the user would get an alert with the message "undefined".

Now the set stays local to the code block where it's used, so that the parameter variable "tags" is an array for the whole function, for more consistency with the "names" variable.

### DIFF
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -26,9 +26,9 @@ Suites.enable = function (names, tags) {
             if (!Tags.has(tag))
                 console.error(`Unknown Suites tag: "${tag}"`);
         });
-        tags = new Set(tags);
+        const tagsSet = new Set(tags);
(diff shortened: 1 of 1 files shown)


resources/tests.mjs
```diff
@@ -46,7 +46,6 @@ Suites.enable = function (names, tags) {
             validNames: this.map((each) => each.name),
         };
     } else if (tags?.length) {
```
julienw, line 48: .length doesn't exist in an array.

camillobruni: ups, thanks for fixing.

camillobruni (approved):

PR #338, opened 2023-12-19
Fixes #335: Prevent complex DOM workloads from hitting a specific WebKit pathology

### LINKED ISSUE #335: Every complex DOM workloads should not hit a specific WebKit pathology
The complex DOM content contains an overflow scroller (`<div class="tree-area">`) with many `position:relative` descendants (`.spectrum-TreeView-itemIndicator`), and the scroller is not a CSS stacking context. This triggers specific code paths related to async scrolling that we don't want affecting the performance on a whole set of subtests. While it's OK to have Speedometer test this code path on one subtest, other "complex DOM" subtests should be constructed in a way to exercise other code paths.

Specifying `will-change: z-index` on these overflow scroller will advert this pathology for example.

To put it another way, the request here is to vary complex DOM contents across the different subtests so as to avoid hitting the same pathology in every complex DOM workload.

issackjohn:
The PR Introduces a new variant of the Complex DOM shell.

Fixes #335.

`big-dom.css` is the original complex DOM shell and this new variant - `big-dom-with-stacking-context-scrollable.css` - is an introduction of the CSS property `isolation: isolate` to the `.tree-area`.

The purpose of this variant is to is to vary complex DOM contents across the different subtests so as to avoid hitting the same pathology in every complex DOM workload.

# Performance Measurements
### ${\textsf{\color{lightblue} lightblue}}$: Running in Stand-alone Setting
We only report the performance results for the **six** TodoMVC tests that we run in the complex setting as well as the four tests that we run in the stand-alone setting but have been changed to use the new variant(big-dom-with-stacking-context-scrollable.css)
${\textsf{\color{green} }}$
| TodoMVC Test | big-dom.css (Original) | big-dom-with-stacking-context-scrollable.css |
|--------|--------|--------|
| ${\textsf{\color{green}TodoMVC-JavaScript-ES6-Webpack-Complex-DOM}}$ | ✅ | - |
| ${\textsf{\color{green}TodoMVC-Lit-Complex-DOM}}$ | - | ✅ |
| ${\textsf{\color{green}TodoMVC-Preact-Complex-DOM}}$ | - | ✅ |
| ${\textsf{\color{green}TodoMVC-Svelte-Complex-DOM}}$ | - | ✅ |
| ${\textsf{\color{green}TodoMVC-React-Complex-DOM}}$ | ✅ | - | 
| ${\textsf{\color{green}TodoMVC-Angular-Complex-DOM}}$ | - | ✅ | 
| ${\textsf{\color{lightblue}TodoMVC-JavaScript-ES5-Complex-DOM}}$ | -  | ✅ | - | 
| ${\textsf{\color{lightblue}TodoMVC-WebComponents-Complex-DOM}}$ |  ✅  | - | - | 
| ${\textsf{\color{lightblue}TodoMVC-React-Redux-Complex-DOM}}$ | -  | ✅ | - | 
| ${\textsf{\color{lightblue}TodoMVC-Backbone-Complex-DOM}}$ | -  | ✅ | - | 
| ${\textsf{\color{lightblue}TodoMVC-Vue-Complex-DOM}}$ |  ✅ | - | - | 
| ${\textsf{\color{lightblue}TodoMVC-jQuery-Complex-DOM}}$ | -  | ✅ | - | 

Screenshot 2023-12-20 at 5 07 41 PM

# Hosted at:
https://issackjohn.github.io/Speedometer5/


resources/todomvc/architecture-examples/angular-complex/scripts/build.js
```diff
@@ -8,11 +8,12 @@ const options = {
     callerDirectory: path.resolve(__dirname),
     sourceDirectory: path.join("..", "node_modules", "todomvc-angular", "dist"),
     title: "TodoMVC: Angular Complex DOM",
-    filesToMove: ["node_modules/big-dom-generator/dist/big-dom-generator.css", "node_modules/big-dom-generator/dist/logo.png"],
+    filesToMove: ["node_modules/big-dom-generator/dist/big-dom-generator-v2.css", "node_modules/big-dom-generator/dist/logo.png"],
```
rniwa, line 11: -v1 and -v2 suffix doesn't really explain the difference. Can we use suffix like "-isolate" so that names are more self explanatory?

rniwa:
I guess you left the ones we don't use complex DOM setup (i.e. only runs in non-complex setup) to be v1?? Is that what's happening here?

issackjohn:
> I guess you left the ones we don't use complex DOM setup (i.e. only runs in non-complex setup) to be v1?? Is that what's happening here?

Yes that's correct.

resources/todomvc/big-dom-generator/utils/createVariant.js
```diff
@@ -0,0 +1,45 @@
...
+        });
+    } catch (error) {
+        console.error("An error occurred while varying the CSS:", error);
+    }
+}
+
+try {
+    const css = fs.readFileSync(INPUT_FILE_PATH, "utf-8");
+    root = postcss.parse(css);
+    vary(".tree-area", [["isolation", "isolate"]], "add");
```
rniwa, line 41: Maybe we should use a dictionary instead? {"isolation": "isolate"}.
That reads slightly better.

rniwa:
> > I guess you left the ones we don't use complex DOM setup (i.e. only runs in non-complex setup) to be v1?? Is that what's happening here?
> 
> Yes that's correct.

Can we switch those to v2 as well so that when we enable them in developer menu, we'd get a meaningful measurement?

flashdesignory:
Is there a way to make the choice between v1 / v2 dynamic? 
Ideally the test suits would tell the complex dom which version to use in my opinion.

sulekhark:
> Is there a way to make the choice between v1 / v2 dynamic? Ideally the test suits would tell the complex dom which version to use in my opinion.

@flashdesignory,
In the current design, the Complex DOM TodoMVC tests are built by embedding the pre-built TodoMVC dist in a pre-built Complex DOM shell (built in the `big-dom-generator` folder). In the last-but-one working group meeting which Simon Fraser attended, I remember we discussed that we would need to have a Complex DOM shell 1 and a Complex DOM shell 2. Keeping these two things in mind, we chose to take the approach in this PR.
I think that what you suggest is possible, but it would need quite some code refactoring. If in the future, we see the need to frequently change the embedding of Complex DOM TodoMVC tests from one shell to another (even in the dev environment), we could make this embedding dynamic (i.e. as you said, specified by the test). WDYT?

bgrins:
I'll let @julienw review, but I took a quick look and it's a relatively small change. Other than the file naming issue in the inline comment I don't have any particular concerns.

I don't have a particular preference on the topic of which variant the non-enabled tests should use. We could always push preview branches in alternative configurations for testing locally & in CI. If I had to choose I suppose I would keep the current (non-isolation) variant as the default for non-enabled, since Lit & Web Components have unique styles from the other tests, so that would maintain the Web Components test as the only way to run (webcomponent style + non isolation).

julienw (changes requested):
I left a few comments, with "request changes" because I'd like to see the result of the changes.

Thanks

resources/todomvc/architecture-examples/angular-complex/scripts/build.js
issackjohn: Done.

resources/todomvc/big-dom-generator/utils/createVariant.js
issackjohn: Done.

resources/todomvc/big-dom-generator/utils/createVariant.js
```diff
@@ -0,0 +1,45 @@
...
+            }
+        });
+    } catch (error) {
+        console.error("An error occurred while varying the CSS:", error);
+    }
+}
+
+try {
+    const css = fs.readFileSync(INPUT_FILE_PATH, "utf-8");
+    root = postcss.parse(css);
```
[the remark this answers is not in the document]
issackjohn: Done.

resources/todomvc/big-dom-generator/dist/big-dom-generator-v1.css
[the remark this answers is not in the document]
issackjohn: Done.

issackjohn:
Screenshot 2023-12-20 at 9 58 12 AM

issackjohn:
Screenshot 2023-12-20 at 10 16 55 AM

sulekhark:
> > > > I guess you left the ones we don't use complex DOM setup (i.e. only runs in non-complex setup) to be v1?? Is that what's happening here?
> > > 
> > > 
> > > Yes that's correct.
> > 
> > 
> > Can we switch those to v2 as well so that when we enable them in developer menu, we'd get a meaningful measurement?
> 
> @rniwa, would it make sense to keep two of the six (which only run in non-complex setup) to be v1 and the rest four v2? Just to be able to see the impact in the developer mode too? Maybe JQuery (or, Vue) and WebComponents with v1? WDYT?

Confirmed with Ryosuke Niwa (via a 1:1 chat on slack) that the above proposed plan is good. So, for the todo tests that are **not enabled in the complex setting** Vue and WC will have variant 1 and Vanilla JS5, JQuery, Backbone, React-Redux will have Variant 2 - this will kick-in only in developer mode.

julienw (approved):
looks good to me, thanks

flashdesignory (approved):

lpardosixtosMs (approved):

bgrins (approved):

rniwa (approved):

issackjohn:
Thank you all!

PR #349, opened 2024-01-16
Allow empty hash for cleaner default URLs on the home screen

camillobruni:
This makes sure we don't have to add `#home` by default.

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -348,9 +348,15 @@ class MainBenchmarkClient {
     }
 
     _setLocationHash(hash) {
-        if (hash === "#home")
-            hash = "";
-        window.location.hash = hash;
+        if (hash === "#home" || hash === "")
+            this._removeLocationHash();
+        else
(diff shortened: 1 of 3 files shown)


resources/main.mjs
```diff
@@ -331,20 +331,26 @@ class MainBenchmarkClient {
...
                 return;
             }
         }
-        window.location.hash = hash || "#home";
+        this._setLocationHash(hash);
+    }
+
+    _setLocationHash(hash) {
+        if (hash === "#home")
+            hash = "";
```
julienw, line 352: actually, even setting it to `""` will add a `#` to the URL, so it could be good to just not do anything if we want to set it to empty and it's already empty.

julienw (approved):

resources/main.mjs
camillobruni: used now history.pushState to remove the hash in case we're on the home screen.

rniwa (approved):

PR #350, opened 2024-01-22
Small lint-related fixes

julienw:
Here are a few fixes for small problems I found while working on the project lately.
Commits should be fairly independent so feel free to look at them separately.

1. Remove tentative/ from the ignore lists. Indeed there's no tentative workload anymore, but working on some on local branches I find it easier that prettier and eslint are not disabled on these directories.
2. This makes tsx files checked by the eslint, as well as mirror the check with the fix configuration.
3. It's good practice to put `root: true` on the top level eslint configuration file. Reference: https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy
4. While running eslint, I found this false positive about the next eslint plugin. It was due to a misconfiguration in the newssite-next directory, so I fixed this in the process.
5. Some file wasn't properly formatted.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,8 +3,6 @@
 package-lock.json
 **/.angular
 
-/resources/tentative/*
-
 /resources/todomvc/architecture-examples/*
 !/resources/todomvc/architecture-examples/angular
 !/resources/todomvc/architecture-examples/angular-complex
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,6 +24,7 @@
  */
 
 {
+    "root": true,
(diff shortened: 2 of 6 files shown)


flashdesignory (approved):
Thanks for the cleanup!

julienw:
I believe just one r+ is enough for this, right?

rniwa (approved):

rniwa:
> I believe just one r+ is enough for this, right?

In theory, yes.

julienw:
2 r+ is more than enough then :) thanks for the reviews

PR #353, opened 2024-01-24
Fix navigation for Firefox

camillobruni:
Temporarily set the `#home` hash before clearing it to ensure proper visibility of the home section.

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -348,10 +348,12 @@ class MainBenchmarkClient {
     }
 
     _setLocationHash(hash) {
-        if (hash === "#home" || hash === "")
+        if (hash === "#home" || hash === "") {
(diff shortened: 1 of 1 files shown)


rniwa (approved):
It's a bit funky bug ok.

bgrins:
Thanks for figuring out a workaround - we'll separately try to narrow down a reduced test case & fix whatever is happening Firefox. @julienw can you take a look at this PR?

camillobruni:
Might be worth having a look at the CSS selector again, maybe something is up with that:

```
section:target,
body:not(body:has(section:target)) #home {
    display: block;
}
```

camillobruni:
ups, this accidentally triggered recursive location updates... fix on the way.

PR #354, opened 2024-01-24
Update document title

camillobruni:
Add a `data-title` to each section to update document.title.

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -358,10 +358,8 @@ class MainBenchmarkClient {
     }
 
     _updateDocumentTitle(hash) {
-        let sectionTitle = "";
         const maybeSection = document.querySelector(hash);
-        if (maybeSection)
-            sectionTitle = maybeSection.getAttribute("data-title") ?? "";
+        const sectionTitle = maybeSection?.getAttribute("data-title") ?? "";
         document.title = `Speedometer 3 ${sectionTitle}`.trimEnd();
     }
 
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -348,15 +348,29 @@ class MainBenchmarkClient {
     }
 
     _setLocationHash(hash) {
-        if (hash === "#home" || hash === "")
+        if (hash === "#home" || hash === "") {
+            window.location.hash = hash = "#home";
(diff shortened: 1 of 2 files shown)


resources/main.mjs
```diff
@@ -348,10 +348,24 @@ class MainBenchmarkClient {
...
+
+    _updateDocumentTitle(hash) {
+        let title = "Speedometer 3";
+        const maybeSection = document.querySelector(hash);
+        if (maybeSection) {
+            const sectionTitle = maybeSection.getAttribute("data-title");
+            if (sectionTitle)
+                title += ` ${sectionTitle}`;
+        }
+        document.title = title;
```
rniwa, line 368: Can we instead just have this instead?:
document.title = `Speedometer 3 ${sectionTitle}`

rniwa (approved):

resources/main.mjs
bgrins: Not all sections have titles, and some kind of non-space delimiter is probably nice so I think the conditional is helpful. But if we wanted to golf this function down I think this would work

```
let sectionTitle = document.querySelector(hash)?.getAttribute("data-title") || "";
document.title = `Speedometer 3${sectionTitle && ` - ${sectionTitle}`}`
```

resources/main.mjs
```diff
@@ -349,11 +349,19 @@ class MainBenchmarkClient {
 
     _setLocationHash(hash) {
         if (hash === "#home" || hash === "") {
-            window.location.hash = "#home";
+            if (window.location.hash !== hash)
+                window.location.hash = hash = "#home";
```
rniwa, line 353: Did you mean to include this change here?

julienw (approved):
same comment as @rniwa but otherwise this looks good to me with the new code.

resources/main.mjs
camillobruni: used simpler code and `trimEnd`

resources/main.mjs
camillobruni: I can land that separately (fix for the navigation recursion that I accidentally landed previously)

camillobruni: ACK, created separate #357

PR #356, opened 2024-01-25
Adjust dot size to iterationCount

camillobruni:
Old:
Screenshot 2024-01-25 at 10 43 50

New:
Screenshot 2024-01-25 at 10 42 54

### DIFF
--- a/resources/metric-ui.mjs
+++ b/resources/metric-ui.mjs
@@ -239,8 +239,12 @@ function renderScatterPlot({ values, width = 500, height, trackHeight, xAxisPosi
     // Axis + labels height:
     const axisHeight = 18;
     const axisMarginY = 4;
-    const markerSize = 5;
     const trackMargin = 2;
+    let markerSize = 5;
+    // Auto-adjust markers to [2px, 5px] for high iteration counts:
+    const iterationsLimit = 20;
+    if (values.length > iterationsLimit)
+        markerSize = 1 + (4 / values.length * iterationsLimit);
     // Recalculate height:
     if (height)
         trackHeight = (height - axisHeight - axisMarginY) / trackCount;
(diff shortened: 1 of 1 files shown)


rniwa (approved):

resources/metric-ui.mjs
```diff
@@ -239,8 +239,12 @@ function renderScatterPlot({ values, width = 500, height, trackHeight, xAxisPosi
     // Axis + labels height:
     const axisHeight = 18;
     const axisMarginY = 4;
-    const markerSize = 5;
     const trackMargin = 2;
+    let markerSize = 5;
+    // Auto-adjust markers to [2px, 5px] for high iteration counts:
```
julienw, line 244: I think this comment is outdated from a previous implementation, now it's `[1px, 5px]`

resources/metric-ui.mjs
```diff
@@ -239,8 +239,12 @@ function renderScatterPlot({ values, width = 500, height, trackHeight, xAxisPosi
...
+    const iterationsLimit = 20;
+    if (values.length > iterationsLimit)
+        markerSize = 1 + (4 / values.length) * iterationsLimit;
```
julienw, line 247: I don't mind, but what do you think of a simpler:
```js
const markerSIze = values.length <= 20 ? 5 : 2; // adjust the marker size to the iteraction count
```
But if you prefer your formula that's also fine by me.

resources/metric-ui.mjs
julienw: that said I find 1px is a bit small. Maybe adjusting the formula to use 2px as the minimum would be better. What do you think?
(but my screen has a high dpi without zoom, so maybe that's why).

julienw (approved):

resources/metric-ui.mjs
camillobruni: Switched back to min size = 2px 👍

resources/metric-ui.mjs
camillobruni: The formula seems easy enough IMO (compared the surrounding complexity :P)

Issue #352, opened 2024-01-24
Add general run instructions


camillobruni:
Let's collect some general (and browser specific) instructions to get consistent Speedometer 3 scores across browsers and platforms.

WIP [google document](https://docs.google.com/document/d/1J4TXULBtnbBLtYYkwlrbK1RWgBWl-M5jTTA_n2gbkkY/edit?tab=t.0#heading=h.5aqetjfqhr8s) to collect information.

bgrins:
Based on the discussion today here, the likely outcome is 

1) Some text to replace the current text on the homepage (`Speedometer is a browser benchmark that measures the responsiveness of Web applications. It uses demo web applications to simulate user actions such as adding to-do items.`)
2) A more detailed runbook for people who want to report the most accurate scores

rniwa:
Yeah, I guess it would be good to point out things like clearing the cache, disabling all browser extensions, etc... for the detailed instructions.

PR #359, opened 2024-01-25
Add separate about.html and instructions.html

camillobruni:
Adds a basic instructions section.
To be extended once the document in #352 is filled in.

Screenshot 2024-01-25 at 11 43 32
Screenshot 2024-01-25 at 11 43 54


julienw:
Is it ready for review or are you waiting for the document?
If it's not ready for review, it would be good to be explicit and mark the PR as a draft :-)
Thanks

camillobruni:
Let's try to land this already so everything is in place and we only need to update the content.

julienw (approved):
looks good to me, thanks
make sure to run `npm run format` before landing.

I noticed one issue, but I believe it's unrelated to this patch, rather related to one of the setLocationHash patches:
1. Start the benchmark.
2. Before the benchmark is finished, press F5.

=> empty page

Can you please look at this issue?

camillobruni:
I forgot to redirect the #running view to #home by default if there are no results yet.

bgrins:
> I forgot to redirect the #running view to #home by default if there are no results yet.

I know it was added for a reason, but I wonder if we should reconsider the hash navigation.

It seems to me the benefit would be if we wanted to allow users to drive UI states with the back button - and in general we don't, in the sense that you can't go back from #summary to #running by design. If we don't need that property there would be simpler ways to manage the UI state that avoid these types of bugs: just set some kind of attribute on the body to toggle display for the corresponding section (or even toggling an attribute on the sections themselves).

One thing that's nice about the current behavior is that you can deep link to the About dialog, but we could handle that as a special case (reading location.hash at page load and toggling the about dialog if it matches "#about"). Any thoughts on this @camillobruni @rniwa ? Not volunteering Camillo to make that change, and I don't think it's a blocker, but if we generally agree on this then maybe @julienw or I could work on a patch if you'd like.

julienw:
In addition to the good suggestion from bgrins, I was also thinking that all the instructions and about texts could be in a separate plain HTML page, so that the runner would stay fairly simple with its 4 states (home / running / details / summary).

rniwa:
> In addition to the good suggestion from bgrins, I was also thinking that all the instructions and about texts could be in a separate plain HTML page, so that the runner would stay fairly simple with its 4 states (home / running / details / summary).

Yeah, that could be a simpler solution overall.

index.html
```diff
@@ -176,6 +177,44 @@ <h1 class="section-header">About Speedometer 3</h1>
...
+                <div class="section-grid">
+                    <h1 class="section-header">Speedometer 3 Test Instructions</h1>
+                    <div class="section-content">
+                        <h2>General Instructions</h2>
+                        <ol>
+                            <li>Before running Speedometer:</li>
+                            <ul>
+                                <li>Close all other applications.</li>
+                                <li>Close other browser windows and tabs.</li>
+                                <li>Use a fresh browser profile and a single window.</li>
```
rniwa, line 195: Maybe we want to explicitly call out that browser extensions should be disabled?

camillobruni:
I'm fine either way.
Maybe hosting the instructions and about text as separate documents is solving most complexity of the hash-based navigation. I'll create WIP PR with that change.

camillobruni:
I've now moved the about and instructions sections into standalone .html files which makes the navigation logic much simpler.

rniwa (approved):

Issue #365, opened 2024-01-27
jQuery test only ever marks the first todo item as completed


rniwa:
Step through jQuery suite in https://speedometer-preview.netlify.app/interactiverunner?suite=TodoMVC-jQuery

We observe that the only first item is marked as completed in "CompletingAllItems" test.

rniwa:
The issue here seems to be that the entire todo list is re-generated each time any checkbox is toggled. So we need to execute `querySelectorAll` in each iteration, or update jQuery implementation to not nuke the whole list each time any todo item's completeness is toggled.

flashdesignory:
we had several prs closed for jQuery, both would address the whole list re-render issue:

https://github.com/WebKit/Speedometer/pull/133

https://github.com/WebKit/Speedometer/pull/140

comment from me from the last example:
"The main difference, which I think is beneficial, is the fact that this implementation doesn't re-render the whole list with every update, which is an optimization that other frameworks / libraries do out of the box."

Happy to take a look at both and open a pr for an updated version!

julienw:
Another possibility would be to use [getElementsByClassName](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName) which is a live collection. This might change the performance characteristics of the workload though.

rniwa:
> Another possibility would be to use [getElementsByClassName](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName) which is a live collection. This might change the performance characteristics of the workload though.

Yeah, that's another alternative. I'd be happy with either approach.

bgrins:
At this point, my preference is to take a change like #366 as opposed to a broader refactor to optimize re-renders. I'd like to make an as-narrow-as-possible-correctness-fix for 3.0, and to consider whether we should make a refactor for a future version.

PR #366, opened 2024-01-27
This PR fixes the issue #365: jQuery test only ever marks the first todo item as completed

rniwa:
The bug was caused by jQuery replacing the whole todo list whenever each todo item is toggled.

Fixed it by running querySelectorAll in each iteration.

### DIFF
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -566,9 +566,8 @@ Suites.push({
             }
         }),
         new BenchmarkTestStep("CompletingAllItems", (page) => {
-            const checkboxes = page.querySelectorAll(".toggle");
             for (let i = 0; i < numberOfItemsToAdd; i++)
-                checkboxes[i].click();
+                page.querySelectorAll(".toggle")[i].click();
         }),
         new BenchmarkTestStep("DeletingAllItems", (page) => {
             for (let i = numberOfItemsToAdd - 1; i >= 0; i--)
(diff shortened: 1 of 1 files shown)


rniwa:
An alternative fix is modify the jquery todo implementation so that it won't nuke the whole todo list whenever one of them is toggled but that seemed like a bigger overhaul.

julienw:
This produces a lot of GC due to the full rerenders... Maybe actually we should actually fix this for toggles? And maybe even for creates and destroys...

FTR I tried different approaches than the one with `querySelectorAll` here, but I got the same scores as a result. I tried using 
```js
querySelector(`li:nth-child(${i}) .toggle`)
```
and with `getElementsByClassName`.
If you want to try, here are the links:
[with querySelectorAll](https://fix-jquery-complete-all-items--speedometer-preview.netlify.app/?suite=TodoMVC-jQuery)
[with nth-child](https://with-nth-child--speedometer-preview.netlify.app/?suite=TodoMVC-jQUery)
[with getElementsByClassName](https://with-getelementsbyclassname--speedometer-preview.netlify.app/?suite=TodoMVC-jQUery)

I'd like to hear other opinions.

camillobruni:
I probably slightly prefer `getElementsByClassName`, but given the little difference in score I don't have a strong opinion.

bgrins:
#367 is fine with me, if works is for @camillobruni / @flashdesignory

julienw:
fixed by #367

PR #367, opened 2024-01-31
Use :nth-child() to fix #365

julienw:
Alternative to #366 using a `:nth-child` selector.

PR #368, opened 2024-01-31
Use getElementsByClassName so that we don't need to use querySelectorAll at every step of the loop

julienw:
Alternative to #366

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -96,6 +96,10 @@ class Page {
         return this._wrapElement(element);
     }
 
+    getElementsByClassName(className) {
+        return this._frame.contentDocument.getElementsByClassName(className);
+    }
+
     call(functionName) {
         this._frame.contentWindow[functionName]();
         return null;
--- a/resources/tests.mjs
+++ b/resources/tests.mjs
@@ -566,8 +566,9 @@ Suites.push({
             }
         }),
         new BenchmarkTestStep("CompletingAllItems", (page) => {
+            const allItems = page.getElementsByClassName("toggle");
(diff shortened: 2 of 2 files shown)


rniwa (approved):

resources/benchmark-runner.mjs
```diff
@@ -96,6 +96,10 @@ class Page {
         return this._wrapElement(element);
     }
 
+    getElementsByClassName(className) {
+        return this._frame.contentDocument.getElementsByClassName(className);
```
rniwa, line 100: Hm... I guess this doesn't wrap each element in `PageElement`. Maybe we want to wrap this in a `Proxy` object which creates `PageElement` for each item in the list?

julienw: Yeah exactly. This more a cheap experiment to look at the performance characteristics. Spoiler: this doesn't change much from your querySelectorAll run in every step of the loop.

IMO wrapping in a Proxy is too complex for what we want to do.
This also led me to take a step back and think more about the PageElement wrapper. What is the purpose of this wrapper? My understanding is that this serves 2 purposes:
1/ Limits the things we can do with a retrieved element. Is this purpose useful in the context of this benchmark? I'd like to argue that it's not.
2/ Provides some common functionality (all the event dispatchers especially). This is useful, but could be provided with a more functional model taking the element as parameter.

Therefore my proposal (for speedometer 4) would be to get rid of the wrapper that brings more complexity than it resolves problems. I'll file an issue about that.

Note: I also wrote https://github.com/WebKit/Speedometer/pull/367 which is IMO more suitable in this case. Tell me what you think about it :-)

julienw:
fixed by https://github.com/WebKit/Speedometer/pull/367

PR #371, opened 2024-02-09
Update prettier, change print width to 120, and enable dangling commas for functions

### LINKED ISSUE #370: Change the prettier printWidth option back to the default
The current value is `250`, this makes very long lines.
Can we put it back to the default, or at least to a smaller value such a 120 if the default is too low because of the tab width?

julienw:
Fixes #370

I didn't regenerate all workloads (yet). Please do not land this just yet, this PR is here so that we can decide about #370.

I made separate commits so that we can easily see the effect of each action:
1. update Prettier
2. add dangling commas to functions
3. change printWidth to 120


rniwa:
At this point, I'd rather not take a huge PR like this into the repo right before the release even if it should not have any real impact on performance in theory.

camillobruni (approved):
Loooking good. I guess we can redo the formatting now that we've released?

julienw:
> Loooking good. I guess we can redo the formatting now that we've released?

yeah, I was waiting to see if there was some interest before doing more work :-)

camillobruni:
I'd definitely prefer some width-limit, especially for the html files.

julienw:
I rebased and refreshed all affected benchmarks.

On my Linux:

Firefox Before:
image

Firefox After:
image

Chrome Before:
image

Chrome After:
image

(I don't have a Mac so I can't try Safari)

instructions.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
```
rniwa, line 1: What's up with these DOCTYPE change? These don't seem desirable.

.prettierrc
```diff
@@ -1,12 +1,4 @@
 {
-    "arrowParens": "always",
```
rniwa, line 2: Why are we making all these other changes to prettier?

julienw: You can find out more if you look at the individual commits, as there's one commit for each logical change.

In short a lot of them are the defaults, except `printWidth` and `trailingComma` which are the changes by this PR

instructions.html
julienw: This comes from the prettier update to v3 (see https://prettier.io/blog/2023/07/05/3.0.0.html#html). There's also no option to disable this behavior.

The doctype is case insensitive so IMO it's not important. I'm glad somebody else has an opinion about it so that I don't need to have one.

julienw:
ping @rniwa

.prettierrc
mdubet: Not removing the default ones would make this PR easier to review/less scary

rniwa (approved):
This seems fine.

rniwa:
The consensus in today's meeting was that we want to put this into version 4, not 3.1 since it affects so many times.

julienw:
Yes, I was waiting for v3.1 to be released to minimize conflicts.

PR #373, opened 2024-02-15
TodoMVC workload based on React 18 / React Router / Material UI

julienw:
Hi,
I'd like to propose this new workload in `tentative/`.
It's based on React 18, React Router and Material UI.

Links:
* Try the app: https://react-18--speedometer-preview.netlify.app/resources/tentative/todomvc-react-18/dist/
* Run SP3 with it: https://react-18-no-autofill--speedometer-preview.netlify.app/?suite=TodoMVC-React-18

I decided to use Material UI as a way to get deeper React trees.
I also used React Router because it's very much used. The way I used it to control the model might lead to discussions though: I thought of it as a way to "emulate" a server side environment. I believe that if we'd have to implement a TodoMVC app with a server side environment, we could use the same architecture. Here of course I used a local model, not a server.
Because of React 18 (and I believe the CSS-in-JS library used in Material UI) I had to override `rAF` and `setImmediate` so that the SP3 runner could capture all the work.
Especially I had to use microtasks (with promises), not normal tasks (with `setTimeout`) like before. Indeed with just `setTimeout`, it was happening that we'd get a `setTimeout` called in another `setTimeout`, and this was sometimes escaping the benchmark runner.

I also decided to use some more steps in the benchmarks, than in the other TodoMVC workloads. I added:
* move between "active" and "all" items
* clear completedi tems

For review, I split this in several commits to make it easier to review, but the bulk of the implementation is in [Implement TodoMVC with all its actions](https://github.com/WebKit/Speedometer/commit/1f6a46246c00f2332e6336f34e568342561b57ec).

If we agree on the general approach, but there are small changes to be done, I'd rather land this first and implement the changes in follow-ups. Hopefully this makes sense!

Happy to hear feedback, folks

### DIFF
--- a/resources/tentative/todomvc-react-18/README.md
+++ b/resources/tentative/todomvc-react-18/README.md
@@ -1,30 +1,43 @@
-# React + TypeScript + Vite
+# Speedometer 3.0: TodoMVC with React 18 / Material UI / React Router
 
-This template provides a minimal setup to get React working in Vite with HMR and some ESLint rules.
+## Description
 
-Currently, two official plugins are available:
+This application uses React 18.0.0 to implement a todo application.
 
--   [@vitejs/plugin-react](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/README.md) uses [Babel](https://babeljs.io/) for Fast Refresh
--   [@vitejs/plugin-react-swc](https://github.com/vitejs/vite-plugin-react-swc) uses [SWC](https://swc.rs/) for Fast Refresh
+-   [React](https://react.dev/) is a JavaScript library for creating user interfaces.
+-   [Material UI](https://mui.com/) is a UI React-based library that makes it easier to implement a consistent style in an application.
+-   [React Router](https://reactrouter.com/) is a Router for React, that is a library that handles URL changes.
 
-## Expanding the ESLint configuration
+## Implementation details
 
-If you are developing a production application, we recommend updating the configuration to enable type aware lint rules:
+React:\
+ Model: /logic/todo-model.js
+View: /components/\*
+controller: The React Router configuration in /components/App.tsx
 
--   Configure the top-level `parserOptions` property like this:
+    MVC:\
+      Model: maintains the data and behavior of an application\
+      View: displays the model in the ui\
+      Controller: serves as an interface between view & model components
 
-```js
-export default {
-    // other rules...
-    parserOptions: {
-        ecmaVersion: "latest",
-        sourceType: "module",
-        project: ["./tsconfig.json", "./tsconfig.node.json"],
-        tsconfigRootDir: __dirname,
-    },
-};
+## Built steps
+
+      To build the static files, this application utilizes vite. It minifies and optimizes output files and copies all necessary files to a `dist` folder.
+
+## Requirements
+
+      The only requirement is an installation of Node, to be able to install dependencies and run scripts to serve a local server.
+
+```
+* Node (min version: 18.13.0)
+* NPM (min version: 8.19.3)
 ```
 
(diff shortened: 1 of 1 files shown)


resources/tests.mjs
```diff
@@ -304,6 +304,50 @@ Suites.push({
...
+
+    tests: [
+        new BenchmarkTestStep("AddingLotsOfItems", (page) => {
+            const button = page.querySelector(".add-lots-of-items-button");
+            button.click();
+        }),
+        new BenchmarkTestStep("CompletingSomeItems", (page) => {
+            const checkboxes = page.querySelectorAll("input[type='checkbox']");
+
+            // We'll be checking 5 checkboxes, in different parts of the list.
```
julienw, line 324: it's 20, not 5

resources/tentative/todomvc-react-18/src/components/App.tsx
```diff
@@ -0,0 +1,94 @@
...
+        path: "/:filter?",
+        element: <TodoListPage />,
+        loader,
+    },
+    {
+        path: "/api/todos",
+        children: [
+            {
+                path: "new",
+                action: async ({ request }) => {
```
flashdesignory, line 17: oh, I like the way the router is set up

julienw: Thanks!
Yeah, it's the new way introduced in react-router v6.4, I like it as well!

camillobruni:
Thanks, looking nice! ...and sorry for the delay.

Q1: Do think it would make sense to display the summary footer as well in the speedometer view?
Screenshot 2024-06-05 at 12 06 40

Q2: I do see a fair amount of unmeasured style and layout updates in chrome. 
  a) Could you confirm this is the same in firefox?
  b) Probably harder: Could we somehow fore measure this as well?
Screenshot 2024-06-05 at 12 08 11

Q3: Would it be interesting to also add a step scroll/jump to an item in the middle of the list?

camillobruni:
Both Chrome and Firefox seem to have an ok score, Safari Version `17.5 (19618.2.12.11.6)` and `Release 195 (Safari 17.4, WebKit 19619.1.15.0.1)` both seem to struggle massively in the delete step.
@rniwa is this something you might want to have a look at?

Screenshot 2024-06-05 at 12 15 19


Compare to Chrome Canary 127.0.6519.0 :
Screenshot 2024-06-05 at 12 16 46

And Firefox Nightly 127.0a1 (2024-05-12) (64-Bit):
Screenshot 2024-06-05 at 12 17 22

resources/tests.mjs
flashdesignory: do you want to update the comment? 🤔 😄

resources/tentative/todomvc-react-18/src/logic/todo-loader.ts
```diff
@@ -0,0 +1,17 @@
+import type { Params } from "react-router-dom";
+import { getTodos } from "./todo-model";
+
+export function loader({ params }: { params: Params<"filter"> }) {
```
flashdesignory, line 4: super duper nit: the name `loader` threw me off... would it be possible to rename it to something like:
`getTodos` or `getFilteredTodos` ?

flashdesignory (approved):
Overall this looks like a nice addition. 
nit: lint step fails at the moment.

bgrins:
@rniwa we discussed getting this one landed in an experimental folder. Care to review before we do that?

rniwa (commented):
seems okay to add this experimentally.

resources/tests.mjs
```diff
@@ -304,6 +304,50 @@ Suites.push({
     ],
 });
 
+Suites.push({
+    name: "TodoMVC-React-18",
```
rniwa, line 308: Can we disable this workload by default since it's experimental?

rniwa (approved):

PR #375, opened 2024-02-15
Rebuild workloads

flashdesignory:
rebuild the following workload directories:
* newssite
* todomvc

Changes are mostly changes in new hashes created for the filenames.
As far as I can tell, no actual changes to the workloads behavior have been added. 
Running charts and editors didn't create any changes. 
React stockcharts and perf.webkit.org has been skipped.

@kara


issackjohn (approved):

bgrins:
Thanks @flashdesignory. @julienw, could you please review?

julienw (approved):
Looks reasonable.

My understanding is that:
* angular-complex gets sourcemaps, it didn't have any before.
* vue doesn't change, only the file names change. Eventually it would be good to configure the vue workload so that the file names don't change if the content doesn't change (use a hash that depends on the content)
* newssite: similarly a filename changed but the content doesn't seem to change. 

So this PR could have only the angular-complex change, not the others. But I don't mind if they're included.

PR #376, opened 2024-02-16
Fix Hidden dropdown menus

### LINKED ISSUE #362: Hidden dropdown menus on TodoMVC: React Complex DOM would be drawn under page content if visible
This page creates a bunch of dropdown menu objects, hidden using `opacity: 0; visibility:hidden`.

If they were ever unhidden by a real user, they'd appear underneath some of the page content. Toggling the opacity and visibility properties on one of them gives this result:

Screenshot 2024-01-26 at 1 19 00 PM

This affects layerization if a browser is trying to pre-render hidden content that might become visible, and it seems unlikely that anyone would actually want them to appear at this position.

I think we should set the `z-index` property, so that these are ordered correctly.

issackjohn:
closes #362

actual change is in [resources/todomvc/big-dom-generator/src/app.css](https://github.com/WebKit/Speedometer/pull/376/files#diff-6f982665322d1ca718143e34506bd78913672736c292b2a5840071ef5846a012).

Before:
Screenshot 2024-02-16 at 9 55 40 AM

After:
Screenshot 2024-02-16 at 9 56 18 AM


rniwa:
What is getting fixed here?

issackjohn:
This PR fixes #362 which describes how the popover menus are being hidden behind the tree-area in the Complex-DOM shell.

image

This changes make it so that they show up on top.

Screenshot 2024-02-16 at 8 45 34 PM

mattwoodrow:
It looks like the "show-more" class also has a popover div that isn't affected by these changes.

issackjohn:
Thanks @mattwoodrow. Will update.

rniwa (approved):
Thanks for the fix!

julienw (approved):
I finally understood the issue and agree with the fix now.
Thanks!

flashdesignory (approved):

lpardosixtosMs:
I think this is good to merge, but I'll confirm with Isaack first. I'll merge it tomorrow at the latest.

PR #377, opened 2024-02-16
Update General Instructions

camillobruni:
- Text taken from the shared doc: http://docs.google.com/document/d/1J4TXULBtnbBLtYYkwlrbK1RWgBWl-M5jTTA_n2gbkkY?tab=t.0#heading=h.lgob4gofqlwz

- Changed `Restart the browser and close all other tabs and windows.`


instructions.html
```diff
@@ -1,52 +1,63 @@
...
+            <div class="section-grid">
+                <h1 class="section-header">Speedometer 3 General Instructions</h1>
+                <div class="section-content">
+                    <p>
+                        To get reasonably accurate results your device and browser need to be focused on the benchmark.
+                        Here are some basic instructions to allow that:
+                    </p>
+                    <ul class="compact">
+                        <li>
+                            Ensure you are running the latest browser version.
```
julienw, line 28: should we mention we want `the latest browser released version`? (as opposed to beta/nightly versions)

instructions.html
```diff
@@ -1,52 +1,63 @@
...
+                        </li>
+                        <li>
+                            Prefer using a separate clean browser profile - extensions and non-default browser settings
```
julienw, line 31: super nit: please use a larger dash to separate the 2 parts of the sentence.
According to https://www.scribbr.com/language-rules/dashes/, this is either a `endash` with spaces, or a `emdash` without spaces, depending on whether we want to look british or american :-)

julienw (approved):
lgtm, thanks

instructions.html
camillobruni: latest is quite a bit ambigous: either date vs release number, none of which are really unique :).
Maybe "latest stable version" is easier to read?

julienw: "stable" was my first thought actually :)

camillobruni:
Update the text again.

instructions.html
```diff
@@ -15,30 +16,19 @@
...
-                        <h3>Safari</h3>
-                    --></div>
+                        <p>To get reasonably accurate results your device and browser need to be focused on the benchmark. Here are some basic instructions to allow that:</p>
+                        <ul class="compact">
+                            <li>Ensure you are running the latest stable browser version.</li>
+                            <li>Prefer using a separate clean browser profile — extensions and non-default browser settings can have a large impact on the score.</li>
+                            <li>Don't run any programs in the background.</li>
+                            <li>Restart the browser and close all other tabs and windows.</li>
+                            <li>Keep the Speedometer page focused and do not interact with the device while running the benchmark.</li>
+                            <li>Keep your mobile device plugged in to your charger.</li>
```
rniwa, line 28: I don't think this applies to just mobile devices like phones. Even laptops can have a variance in its performance if it's connected to AC vs on battery.

instructions.html
```diff
@@ -15,30 +16,19 @@
...
+                            <li>Some devices might need time to cool down between subsequent runs.</li>
```
rniwa, line 29: Maybe "between consecutive runs" would read better?

instructions.html
camillobruni: Changed this to "battery-powered device" to be a bit more generic.

rniwa (approved):

julienw:
The new text looks good to me as well.

PR #391, opened 2024-03-02
Enable source maps in charts and editors

julienw:
Sorry this is also a bit late.

There's no change in the production code except for the comment pointing to the sourcemap file.
Also I've checked that we don't have the issue about referenced directories in the source map file.


rniwa:
Please post a PR against main instead. I'll remaster release/3.0 once these are merged.

julienw:
> Please post a PR against main instead. I'll remaster release/3.0 once these are merged.

done
I was thinking the opposite was easier and more common, so that we could merge to main things unrelated to v3.0.

rniwa (approved):

flashdesignory (approved):

PR #395, opened 2024-03-05
Avoid :has selector in CSS.

### LINKED ISSUE #394: :has selector breaks speedometer in older browsers
The `:has` selector introduced by #349 makes it so that Speedometer only runs in very recent browsers:
https://github.com/WebKit/Speedometer/blob/7efb0ac57b25195aebd7a9b719002cb40db0944f/resources/main.css#L309-L311

Can we change the CSS so that older browsers can still run Speedometer 3?

For example, Firefox ESR  does not support `:has`. Firefox 121 shipped `:has` support on December 19, 2023.

We'd like to use Speedometer 3 to get a sense of how much browser performance has improved over the last 2 years or so, and we can't use it for that purpose if it doesn't run in older versions.

mstange:
Fixes #394.

This changes how section visibility is controlled. All sections are still `display:none` by default.
Rather than becoming `display:block` when they match `:target`, they now become `display:block` if their ID matches the value of the root element's `data-visible-section` attribute - we hardcode the four section names "home", "running", "summary" and "details".

This lets us default to the home section just by setting `<html data-visible-section="home">` in index.html, and we can avoid the `:has` selector.

Now Speedometer can run in older browsers which don't support `:has`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html data-visible-section="home">
     <head>
         <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
         <meta name="viewport" content="width=850" />
--- a/resources/main.css
+++ b/resources/main.css
@@ -304,9 +304,11 @@ section {
     border-radius: 20px;
 }
 
-section:target,
 section.visible,
-body:not(body:has(section:target)) #home {
+:root[data-visible-section="home"] #home,
+:root[data-visible-section="running"] #running,
+:root[data-visible-section="summary"] #summary,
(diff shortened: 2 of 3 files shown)


rniwa (approved):

julienw (approved):
Looks good to me. IMO It's also simple enough to land in v3.0 but of course please test it in your respective browsers. (I tested myself in firefox nightly, stable and esr, as well as chrome dev and stable)

At this point the hash is only useful so that we can press back/forward buttons between the summary and details views. Is it worth the extra code? (Question for v3.1)

camillobruni (approved):

flashdesignory (approved):

PR #398, opened 2024-03-12
chore: upgrade sinon to latest

mroderick:
The updated version has a few bug fixes and a newer version of `fake-timers`.
Nothing really spectacular, but it's nice to not fall too far behind on major versions.

See https://github.com/sinonjs/sinon/blob/main/CHANGES.md


julienw (approved):
looks reasonable to me, thanks for the update!

PR #404, opened 2024-03-22
[fix]: update nuxt config: remove hash from file names

flashdesignory:
Changing nuxt config, to avoid hash in file names.

moving discussion from https://github.com/WebKit/Speedometer/pull/400 to this pr, in regards to build files creation.

@kara


rniwa (approved):

resources/newssite/news-nuxt/nuxt.config.ts
```diff
@@ -39,4 +39,15 @@ export default defineNuxtConfig({
         cdnURL: "./",
         baseURL: "./",
     },
+    vite: {
+        build: {
+            rollupOptions: {
+                output: {
+                    chunkFileNames: "_nuxt/[name].js",
+                    entryFileNames: "_nuxt/[name].js",
+                    assetFileNames: "_nuxt/[name].[ext]",
```
julienw, line 48: I'm not convinced: if we do that, when we'll update the file, the old version might be served from the cache. This will lead to bugs and frustrations.

I think the hash changing is a symptom that the content changes. I'm not too concerned about the hash change, it is easy to review. I'm more concerned to know why the content changes.
That's not a super big deal. Though I wonder if removing the minify step would be acceptable, as this would make easier to review the rendered code. Set [this to false in vite's configuration](https://vitejs.dev/config/build-options#build-minify), though [it looks like there's an option in nuxt](https://nuxt.com/docs/api/nuxt-config#minimize) too.
What do you think?

PR #406, opened 2024-04-08
Don't display invalid results

### LINKED ISSUE #399: With privacy.resistFingerprinting set to true in Firefox, Speedometer shows the result Infinity
This was reported by some people on social networks. We found that setting privacy.resistFingerprinting to true was triggering this. But this might happen in some other cases too.

When doing it, some subtests take 0ms on every run, this may mess up the results.
Any chance we can fix this?

camillobruni:
This partially fixes #399 by not display non-finite scores which are caused by 0-measured suite results.
- For non-positive or non-finite scores, the following summary page is displayed. Note that the detailed page is still accessible since you might find some useful information there.
- Additionally a `console.error` is generated for each zero-sum suite

Screenshot 2024-04-08 at 11 17 26


flashdesignory:
should we display a message to the user, explaining what happened and that the details view might have useful information?

camillobruni:
Updated version with a better description:
Screenshot 2024-04-08 at 16 57 52

rniwa:
Rather than "Invalid Score", how about "Error"? Also, we should probably stop right away when we detect an invalid value instead of running 'til completion.

camillobruni:
Screenshot 2024-04-09 at 11 04 00

camillobruni:
Updated the code to stop after the first iteration with a broken score (tracking errors across the TestInvoker seems a bit more tricky at the moment).

julienw (approved):
I don't have a strong opinion about the exact behavior but this looks reasonable to me. Thanks!

resources/benchmark-runner.mjs
```diff
@@ -528,6 +540,8 @@ export class BenchmarkRunner {
             const values = [];
             for (const suiteName in this._measuredValues.tests) {
                 const suiteTotal = this._measuredValues.tests[suiteName].total;
+                if (suiteTotal <= 0)
```
rniwa, line 543: Instead of waiting until the very end, we should check `this._measuredValues.tests[suite.name].total` is non-zero at the end of `_runSuite`.

julienw, line 543: could we have a comment with this check, explaining some cases where this might happen. For example:
```
// When the test is fast and the precision is low (for example with Firefox'
// privacy.resistFingerprinting preference), it's possible that the measured
// duration is 0.
```

Another thought, that might be another topic unrelated to this patch: if for a suite, `suiteTotal` is between 0 and 1, that will get the geomean down (if I'm not wrong), is that something we want? I kind of think we'd like to set a lower bound at 1.
(I'm really bad at statistics, have I said this already ? :-) )

camillobruni: Added comment.

As for the geoeman, it doesn't really matter whether values are < or > 1, it scales the same way:

value 1 | value 2 | value 3 | value 4 | geomean | mean
-- | -- | -- | -- | -- | --
0.1 | 0.2 | 0.6 | 0.9 | 0.3224 | 0.4500
1 | 2 | 6 | 9 | 3.2237 | 4.5000
10 | 20 | 60 | 90 | 32.2371 | 45.0000

Hope this helps. The main property we're trying to exploit is that relative changes of any workload (independent of absolute duration) has the same impact on the total score, as to not favour longer workloads over short running ones.

camillobruni: Added `_validateSuiteTotal` which is called from `_runSuite`

camillobruni: Moved the comment to `_validateSuiteTotal`

flashdesignory (approved):

resources/benchmark-runner.mjs
julienw: > Hope this helps. The main property we're trying to exploit is that relative changes of any workload (independent of absolute duration) has the same impact on the total score, as to not favour longer workloads over short running ones.

thanks this makes sense

camillobruni:
@rniwa could you have another look maybe?

camillobruni:
@rniwa do you have additional feedback or would be ok with landing with your proposed changes?

rniwa (approved):

Issue #407, opened 2024-04-16
Arithmetic mean is displayed in the detailed results instead of geometric mean


guillaumenibert:
Hello,

By analysing some benchmarks on [browserbench.org/Speedometer3.0](https://browserbench.org/Speedometer3.0/), I noticed an inconsistency in my results: the value displayed in the detailed results page is an arithmetic mean, not a geometric mean, and the associated deltaPercentage is also related to the arithmetic mean.

![Screenshot from 2024-04-16 17-53-49](https://github.com/WebKit/Speedometer/assets/40465432/2b191038-3bf3-43e1-8ad8-ad83ff4cee62)

Looking at the associated JSON, it becomes clear that the displayed value is not the geomean.

```json
...
"Geomean": {
        "name": "Geomean",
        "unit": "ms",
        "mean": 103.97435840648527,
        "geomean": 103.72593600263131,
        "delta": 5.503709786390318,
        "percentDelta": 5.293333732220493,
        "sum": 1039.7435840648527,
        "min": 96.63929714231493,
        "max": 118.13667941264073,
        "values": [
            98.03324073445613,
            99.50013883138098,
            97.04862399436112,
            107.64823016376299,
            103.19284210227842,
            97.74388483491845,
            96.63929714231493,
            107.28020836431752,
            114.52043848442153,
            118.13667941264073
        ],
...
```

I also double-checked the calculations using the functions in the [metric.mjs](https://github.com/WebKit/Speedometer/blob/13cc928140e2a48926c229c1a9febf004b747085/resources/metric.mjs) and [statistics.mjs](https://github.com/WebKit/Speedometer/blob/13cc928140e2a48926c229c1a9febf004b747085/resources/statistics.mjs) files, the percentDelta value is calculated in relation to the arithmetic mean.

Best regards,

Guillaume Nibert

rniwa:
Oh, this is a pretty bad bug! Thank you for reporting!

camillobruni:
I think this is a bit of a confusion in how the data is structured.

 -`Geomean` (uppercase) refers to metric which contains the geomean of the subtest times.
- `values` contains an entry per iteration so that's `values[0] = geomean(iterations[0].subtest_1.total, ..., iterations[0].subtest_N.total)`
- `geomean` is the geomean of all the `Geomean` `values` (this is mostly auto-added for completeness but doesn't add much value in this context)
- `mean` is the geomean of all the `Geomean` `values` (this is what's displayed)

[...]

PR #408, opened 2024-05-13
Deconfuse Geomean Metric and geomean aggregate value

camillobruni:
The metric's "geomean" aggregate value can be easily confused with the top-level Geomean metric (issue #407).

- Don't serialize "geomean" in the JSON data
- Add descriptions to Score, Geomean, and Iteration metrics that are serialized in the JSON data
- Make the detail view titles more descriptive

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -574,9 +574,9 @@ export class BenchmarkRunner {
             // Prepare all iteration metrics so they are listed at the end of
             // of the _metrics object, before "Total" and "Score".
             for (let i = 0; i < this._iterationCount; i++)
-                iterationTotalMetric(i);
-            getMetric("Geomean");
-            getMetric("Score", "score");
+                iterationTotalMetric(i).description = `Test totals for iteration ${i}`;
+            getMetric("Geomean", "ms").description = "Geomean of test totals"
+            getMetric("Score", "score").description = "Scaled inverse of the Geomean";
         }
 
         const geomean = getMetric("Geomean");
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -575,7 +575,7 @@ export class BenchmarkRunner {
             // of the _metrics object, before "Total" and "Score".
             for (let i = 0; i < this._iterationCount; i++)
                 iterationTotalMetric(i).description = `Test totals for iteration ${i}`;
-            getMetric("Geomean", "ms").description = "Geomean of test totals"
+            getMetric("Geomean", "ms").description = "Geomean of test totals";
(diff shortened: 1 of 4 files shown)


resources/metric.mjs
```diff
@@ -10,9 +10,13 @@ export class Metric {
...
         this.name = name;
         this.unit = unit;
+        this.description = "";
 
         this.mean = 0.0;
-        this.geomean = 0.0;
+        Object.defineProperty(this, "geomean", {
+            writable: true,
+            value: 0,
+        });
```
julienw, line 19: Is there a reason you're using defineProperty here? If yes can you please add a comment explaining why?

resources/benchmark-runner.mjs
```diff
@@ -574,9 +574,9 @@ export class BenchmarkRunner {
             // Prepare all iteration metrics so they are listed at the end of
             // of the _metrics object, before "Total" and "Score".
             for (let i = 0; i < this._iterationCount; i++)
-                iterationTotalMetric(i);
-            getMetric("Geomean");
-            getMetric("Score", "score");
+                iterationTotalMetric(i).description = `Test totals for iteration ${i}`;
+            getMetric("Geomean", "ms").description = "Geomean of test totals";
+            getMetric("Score", "score").description = "Scaled inverse of the Geomean";
```
julienw, line 579: You're not doing anything with these `description` properties, aren't you?

julienw:
BTW the CI doesn't pass for some reason...

resources/benchmark-runner.mjs
camillobruni: Correct, this is just for them to show up the serialized JSON as sort of an explainer.

resources/metric.mjs
camillobruni: This is the `Don't serialize "geomean" in the JSON data` part, since it's non-enumerable.

resources/benchmark-runner.mjs
julienw: oh I see, makes sense

resources/metric.mjs
julienw: The comment below was mentioning circular references though, it's not the case here I guess.
You only want to avoid that it's serialized? The reason isn't clear to me, that's why explaining the "why" in a comment would be useful.

camillobruni: added more explicit comment.

resources/benchmark-runner.mjs
camillobruni: I might add a help-text  / hover alt text in the future with the description.

julienw (approved):

PR #410, opened 2024-05-20
Remove obsolete eslint ignore

kkuegler:
The offending `new Promise(async ...)` code was already removed in 19184dc69432fa7e511e13f3318c8c3e1acd28ca but this linting comment remained.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -453,7 +453,6 @@ export class BenchmarkRunner {
     }
 
     async _runTestAndRecordResults(suite, test) {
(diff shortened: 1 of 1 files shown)


rniwa (approved):

julienw:
We could use `reportUnusedDisableDirectives: true,` in our eslint configuration as well as add `--report-unused-disable-directives` when calling eslint in our CI and package.json. Note it's useful to use both, because the property `reportUnusedDisableDirectives` only warns (so it doesn't error the test) but shows up in editors, while the command line parameter errors and fails the test.

julienw (approved):

camillobruni (approved):

PR #414, opened 2024-06-18
Add local storage to todoMVC-es5

flashdesignory:
This pr switches the storage from in-memory to localStorage for todomvc-es5.

Scores snapshot:

| browser  | before | after |
| -------- | -------- | -------- |
| chrome  | 30.5 | 30.3 |
| firefox | 26.12 | 25.6 |
| safari | 25.5 | 25.3 |

@kara

### DIFF
--- a/experimental/todomvc-localstorage/README.md
+++ b/experimental/todomvc-localstorage/README.md
@@ -1,0 +1,42 @@
+# Speedometer 3.0: TodoMVC: JavaScript Es5
+
+## Description
+
+This application uses JavaScript with ES5 language features to implement a todo application.
+
+JavaScript (JS) is a lightweight, interpreted, or just-in-time compiled programming language with first-class functions. While it is most well-known as the scripting language for Web pages, many non-browser environments also use it..
+
+[JavaScript - developer.mozilla.org](http://developer.mozilla.org/en-US/docs/JavaScript)
+
+## Implementation details
+
+This implementation uses an explicit MVC pattern, with a clear file structure to reflect the architecture. The storage solution uses an in-memory data object that implements a simple array to hold the todos.
+
+## Built steps
+
+A simple build script copies all necessary files to a `dist` folder.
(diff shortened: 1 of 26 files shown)


julienw:
Hey Thorsten, can you remind me when/where we've discussed about introducing localStorage into a workload?

As far as I know, the latest decision about that was https://github.com/WebKit/Speedometer/pull/251#discussion_r1237232511, that is localStorage introduces I/O in the workloads and therefore is another source of noise, so we decided to not include it in workloads.

It's totally possible I forgot about a past discussion too.

flashdesignory:
@julienw - there was previous concern and we did opt out of using localStorage for our workloads for 3.0.

The reason I think that it's a valid addition:
1. it's widely used in real world applications and therefore browsers should be able to optimize for the api usage
2. although we need to be careful with not persisting localStorage between tests, this is absolutely do-able and should not be a reason to avoid it.

I'm also happy to include this in an experimental branch, if there's still concern. Although I believe that we are missing out on opportunities to optimize for our users, if we're not taking storage apis in consideration. They ARE widely used, if we optimize for them or not.

camillobruni:
Let's maybe incubate this as an experimental workload and see how it behaves on various devices (I have to test this on mobile as well which are known to have slow flash storage).

Maybe we could prime the local storage (aka. a single access in prepare) so we don't necessarily measure the first-time setup costs if they're disproportional to the runtime costs?

flashdesignory:
changes since last comment:

- added a dedicated experimental folder in the root of speedometer (this replaces 'tentative')
- updated benchmark-runner to not prepend 'resources', since the experimental folder is in the root
- local storage example accesses localStorage once in the prepare phase (added a helper function to expose localStorage to the Page class)

rniwa (approved):

bgrins (approved):
Experimental for testing: let's land it

PR #418, opened 2024-07-11
install speedometer-workloads

flashdesignory:
# Use speedometer-workloads package instead of local workloads

This pr installs a package that contains all workloads, in favor of using workloads that are currently in the resource folder.

## Testing

for testing purposes, only installed chart.js workload and added a tag to compare "before&after".

http://127.0.0.1:7000/?developerMode=&tags=compare

## What is speedometer-workloads?

[Speedometer-workloads](https://github.com/GoogleChromeLabs/speedometer-workloads) is a standalone monorepo that contains example apps and utilities that can be installed and used, instead of the current workloads that are included in the speedometer benchmark itself.

## Benefits of using speedometer-workloads

Decoupling workloads from the benchmark would allow more flexibility and control over what workloads to use for local testing and how they can be configured for these tests.

Having a more modern and dynamic development setup will promote speed of development and motivate experimentation of new workloads, without being forced to adhere to technical constraints of Speedometer itself.

Speedometer-workloads can adhere to its own release cycle and will establish a new versioning of these workloads, which will result in more granular control on what workloads to install for each release of Speedometer.

A nice side effect of this change is that the workloads would also be re-usable for other benchmarks, if the need arises in the future.

## Workloads install

Workloads are now installed through a (private) npm package. Currently, one single package is used for all workloads, instead of installing each workload separately. The Benefit of using a single package is simplicity. Downside is that we can’t decide in Speedometer, which version of a specific workload should be installed. This should be an easy step to change though.

### speedometer-workloads

This package should be installed as a regular dependency of Speedometer.
For testing purposes, the current package is private, until the changes are approved and agreed on.
Since the package is private, the npm package name is scoped: @tkober/speedometer-workloads.
Link to the speedometer-workloads source:  [https://github.com/GoogleChromeLabs/speedometer-workloads](https://github.com/GoogleChromeLabs/speedometer-workloads)

```json
"dependencies": {
    "@tkober/speedometer-workloads": "^1.3.0"
}
```

### workloads.config.folder.json

This file determines which workload should get copied over from the @tkober/speedometer-workloads package.

```json
{
    "workloads": [
        { 
            "name": "charts-chartjs",
            "type":"static"
        }
    ]
}
```

### copy-workloads

The copy-workloads script in the package.json file, creates a workloads folder in the root of Speedometer with a copy of the dist folders of each work that is listed in the workloads.config.folder.json file.

```json
"scripts": {
    "copy-workloads": "npm explore @tkober/speedometer-workloads pnpm run move:apps:custom"
}
```

## Notes

- For local testing, please reach out to me and I'll add you to the private npm package

@kara


resources/benchmark-runner.mjs
```diff
@@ -448,7 +448,7 @@ export class BenchmarkRunner {
                 await suite.prepare(this._page);
                 resolve();
             };
-            frame.src = `resources/${suite.url}`;
+            frame.src = `${suite.url}`;
```
flashdesignory, line 451: frame source is now relative to the root of the repo, so that we can point to either `/workloads` or `/resources`

resources/tests.mjs
```diff
@@ -58,7 +58,7 @@ Suites.enable = function (names, tags) {
 
 Suites.push({
     name: "TodoMVC-JavaScript-ES5",
-    url: "todomvc/vanilla-examples/javascript-es5/dist/index.html",
+    url: "resources/todomvc/vanilla-examples/javascript-es5/dist/index.html",
```
flashdesignory, line 61: Changes in this file are mainly to point to the workloads from the root

PR #420, opened 2024-07-24
WIP: base async step implementation

camillobruni:
This WIP PR adds `AsyncBenchmarkTestStep` and completely separate async run paths so the default runs are not affected.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -7,6 +7,13 @@ export class BenchmarkTestStep {
     constructor(testName, testFunction) {
         this.name = testName;
         this.run = testFunction;
+        this.isAsync = false;
+    }
+}
+export class AsyncBenchmarkTestStep extends BenchmarkTestStep {
+    constructor(testName, testFunction) {
+        super(testName, testFunction);
+        this.isAsync = true;
     }
 }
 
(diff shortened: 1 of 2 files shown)


resources/tests.mjs
```diff
@@ -953,7 +953,27 @@ Suites.push({
...
+            page.querySelector("#prepare").click();
+            page.querySelector("#add-scatter-chart-button").click();
+        }),
+        new AsyncBenchmarkTestStep("Show tooltip", (page) => {
+            page.querySelector("#open-tooltip").click();
+        }),
+        new AsyncBenchmarkTestStep("Draw opaque scatter", (page) => {
+            page.querySelector("#opaque-color").click();
+            page.querySelector("#add-scatter-chart-button").click();
+        }),
```
smaug----, line 974: Hmm, what is async here? Should the test functions be async?

resources/benchmark-runner.mjs
```diff
@@ -311,6 +337,23 @@ class RAFTestInvoker extends TestInvoker {
     }
 }
 
+class AsyncRAFTestInvoker extends BaseRAFTestInvoker {
+    _scheduleCallbacks(resolve) {
+        requestAnimationFrame(async () => {
+            await this._syncCallback();
```
smaug----, line 343: Ok, so this makes sync callback the async thingie. Which makes sense given how callbacks are used currently, but naming is a bit confusing then. Perhaps we could rename the callbacks to not be sync and async, but something else. Not sure what.

camillobruni: yeah absolutely, this is more confusing than ever :). 
- `main` vs .`trailing` or so?
- `blocking` vs. `delayed`?

resources/benchmark-runner.mjs
```diff
@@ -311,6 +337,23 @@ class RAFTestInvoker extends TestInvoker {
...
+            requestAnimationFrame(() => {
```
smaug----, line 344: This rAF is possibly a bit problematic. It is likely that after resolving syncCallback, one may need to wait quite a while before getting vsync, at least if the callback was very fast.
Another option could be to just immediately call setTimeout() => { this._asyncCallback(), but then that has similar issues what SP2 had where a paint might or might not be included in the measured time. So guess we do need rAF  here somehow.

Or, what if we run this._asyncCallback(); using queueMicrotask. That way promise callbacks triggered by _syncCallback() would get run before asyncCallback, but possible setTimeout(0)s wouldn't be.

resources/benchmark-runner.mjs
flashdesignory: How about something like this?

class TestInvoker {
    constructor(captureTest, captureLayout, reportResults) {
        this._captureTest = captureTest;
        this._captureLayout = captureLayout;
        this._reportResults = reportResults;
    }
}

PR #421, opened 2024-08-22
Use more helper runner helper methods

camillobruni:
Refactor the runner a bit to be more flexible for upcoming experimental PRs.


flashdesignory (approved):

resources/benchmark-runner.mjs
```diff
@@ -474,14 +494,12 @@ export class BenchmarkRunner {
...
+    async _loadFrame(suite) {
+        return new Promise((resolve, reject) => {
             const frame = this._page._frame;
-            frame.onload = async () => {
-                await suite.prepare(this._page);
-                resolve();
-            };
-            frame.src = `${suite.url}`;
+            frame.onload = () => resolve();
+            frame.onerror = () => reject();
```
julienw, line 501: 👍

tests/benchmark-runner-tests.mjs
```diff
@@ -106,22 +108,31 @@ describe("BenchmarkRunner", () => {
...
+                for (const suite in runner.suites)
+                    spy(suite, "prepare");
 
-                await runner._runAllSuites();
+                await runner.runAllSuites();
+            });
+
+            it("should call prepare on all suites", () => {
+                for (const suite in runner.suites)
+                    suite.prepare.calledOnce();
```
julienw, line 127: I'm surprised this is working. I think `calledOnce` is a property on a spy, not a function?
You probably wanted `assert.calledOnce(suite.prepare)`. This makes me wonder if we really execute this line.

Debugging further it looks like indeed `runner.suites` is undefined, that's why the test passes (the for loop never runs anything). For loops are dangerous in tests :-)
I think you wanted `_suites`, not `suites` (also above in the `before`). I tried this but then I get more errors, so I'll let you debug :-D

Also it would be good to add something such as `expect(runner._suites).not.to.have.length(0)` to make sure we enter the loop.

resources/benchmark-runner.mjs
```diff
@@ -408,23 +408,32 @@ export class BenchmarkRunner {
...
 
+        return suites;
+    }
+
+    _shuffleSuites(suites) {
+        // We just do a simple Fisher-Yates shuffle based on the repeated hash of the
+        // seed. This is not a high quality RNG, but it's plenty good enough.
+        for (let i = 0; i < suites.length - 1; i++) {
+            let j = i + (this._suiteOrderRandomNumberGenerator() % (suites.length - i));
+            let tmp = suites[i];
```
rniwa, line 425: Use const maybe?

rniwa (approved):

tests/benchmark-runner-tests.mjs
camillobruni: good catch, added some more checks.

tests/benchmark-runner-tests.mjs
```diff
@@ -106,22 +108,37 @@ describe("BenchmarkRunner", () => {
...
+            });
 
-                await runner._runAllSuites();
+            it("should call prepare on all suites", () => {
+                let suitesPrepareCount = 0;
+                for (const suite of runner._suites) {
+                    suitesPrepareCount += 1;
+                    assert.calledOnce(suite.prepare);
+                }
+                expect(suitesPrepareCount).equal(2);
```
julienw, line 132: uber nit:
```suggestion
                expect(suitesPrepareCount).equal(SUITES_FIXTURE.length);
```

PR #423, opened 2024-08-29
Avoid key update for every re-render of ArticleContent with the same content

### LINKED ISSUE #422: NewsSite-Next workload is generating keys for <li> element on the fly
Since the data doesn't contain an `id` field for each item, the workload is now generating keys for each `<li>` element dynamically while rendering, using `uuidv4()`.
https://github.com/WebKit/Speedometer/blob/3150f4117389931ba36871c4a0ab4e289d273965/resources/newssite/news-next/src/components/article/article-content.jsx#L23-L32

This will cause keys to never match up between renders, leading to these elements being recreated every time, which defeats the purpose of keys and slows the performance [[1](https://react.dev/learn/rendering-lists#rules-of-keys)].

Besides, this implementation seems inconsistent with that of NewsSite-Nuxt, which uses `item.id` (which is undefined) as the key binding directly.
https://github.com/WebKit/Speedometer/blob/3150f4117389931ba36871c4a0ab4e289d273965/resources/newssite/news-nuxt/components/atoms/ArticleContent.vue#L29-L34

[1] https://react.dev/learn/rendering-lists#rules-of-keys

lingyuncai:
This is to address #422.

Use `useMemo()` to avoid `uuidv4()` call on the same content, which introduced unnecessary re-renders of the same `<li>` element.


julienw:
This looks reasonable to me, but I wonder if the `id` prop could be generated instead wherever we create the items in the first place. I'd like @flashdesignory to chime in here.

Also you mention in #422 that nuxt has a similar issue where `id` is undefined. I believe this could be solved the same way.

flashdesignory:
@julienw - you are correct, the id should be coming from the data itself. 
Looks like I didn't add an id to the article content. 

I'll update the data and open a pr

Thanks @lingyuncai for surfacing the issue

lawchingman (approved):

lawchingman (approved):

lingyuncai:
Hi @julienw and @flashdesignory, thanks for the review!

Storing the `id` prop in locally generated data is indeed a more clean way to achieve the goal, while it also unifies the behavior across the Next/Nuxt workloads.

Thanks again for the effort on this :)

lingyuncai:
Yes, closing now.

PR #426, opened 2024-08-31
News Site Next

flashdesignory:
This fixes  next.js issue: https://github.com/WebKit/Speedometer/issues/422
A separate pr for Nuxt will do the same. 

The fix here is to add unique ids to any article content that is a list.
This is done in the actual data files, so no need for the app to generate them at any point.

@kara


camillobruni (approved):
Generally LGTM: might be worth adding a little test to check for consistency across languages and uniqueness of IDs within a language.

flashdesignory:
yes, absolutely!
Tests are needed and will come in a future PR at some point.

julienw (approved):
This looks good to me, thanks for the fix.
One more thing: in `section.jsx` you can use `article.id` instead of concatenating the section id with `index`.

rniwa:
Do we have before/after numbers?

flashdesignory:
> Do we have before/after numbers?

here's a snapshot
| browser | before  | after  |
|---|---|---|
| chrome | 72.27 | 70.27 |
| firefox | 77.47 | 72.43 |
| safari | 54.03 | 57.47 |

rniwa:
Is that runtime or pts?

flashdesignory:
> Is that runtime or pts?

we're displaying runtime in the detail screen, right?

flashdesignory:
@rniwa - any concerns to merge this?

flashdesignory:
closing in favor of: https://github.com/WebKit/Speedometer/pull/447

PR #427, opened 2024-08-31
News Site Nuxt

flashdesignory:
This fixes next.js issue for Nuxt: https://github.com/WebKit/Speedometer/issues/422

The fix here is to add unique ids to any article content that is a list.
This is done in the actual data files, so no need for the app to generate them at any point.

@kara


resources/newssite/news-nuxt/data/index.js
```diff
@@ -1,22 +1,22 @@
-import { content as contentEn } from "~/data/en/content";
-import { content as contentJp } from "~/data/jp/content";
-import { content as contentAr } from "~/data/ar/content";
+import { content as contentEn } from "@/data/en/content";
```
julienw, line 1: Just curious, why have you changed this?

julienw (approved):
Thanks!

similarly to the other PR, I believe you can use `article.id` in `Section.vue` (instead of concatenating the section id and `index`).

resources/newssite/news-nuxt/data/index.js
flashdesignory: looks like nuxt supports both ways and this is consistent with next.js. 
At least it still worked after I copied and pasted it by accident 🤣 

I can revert that change though if there are concerns

julienw: I couldn't find any doc about that except from a babel plugin that I don't think we're using... Whatever works works also for me here

rniwa:
Do we have before/after numbers?

julienw:
I checked just Firefox on my computer, couldn't see any difference.

flashdesignory:
> Do we have before/after numbers?

Here's a snapshot:
| browser | before  | after  |
|---|---|---|
| chrome | 64.80 | 63.02 |
| firefox | 66.03 | 67.27 |
| safari | 60.47 | 65.13 |

flashdesignory:
@rniwa - I did some testing with @camillobruni and here are our crossbench results:

current workload:
| browser | safari         |           Firefox      |          Google Chrome               |
|---------|------------|-----------------|-------------------------------|
| duration |  61.80 ± 1.5%        |      66.96 ± 0.72%     |     66.0 ± 1.8%      |
|    score     |           16.27 ± 1.3%      |        14.98 ± 0.68%     |     15.19 ± 1.8% |

pr workload:
| browser | safari         |           Firefox      |          Google Chrome               |
|---------|------------|-----------------|-------------------------------|
| duration |   61.68 ± 0.74%        |      66.04 ± 1.2%     |    65.40 ± 1.2%     |
|    score     |            16.31 ± 0.68%      |        15.18 ± 1.1%     |     15.33 ± 1.1% |

flashdesignory:
closing, in favor of: https://github.com/WebKit/Speedometer/pull/448

Issue #325, opened 2023-10-07
(dev)-server port is already occupied on macOS by default


denschub:
Okay, this is a bit of a weird issue, but this felt like worth highlighting: the default dev-server/server port set in `package.json` is 7000. On a default macOS, this port is already occupied by the AirPlay Receiver feature (`ControlCenter` is listening on it).

This isn't a big deal, of course, but I wonder if the default port should be something else, just to avoid people being confused for a couple of moments. :)

PR #429, opened 2024-09-09
Change default port to 8900 

mdubet:
Port 7000 is occupied by default on macOS.
This also makes it simpler how to customize it
(use env var instead of hardcoded in the package.json)

https://github.com/WebKit/Speedometer/issues/325

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     },
     "license": "SEE LICENSE IN LICENSE",
     "scripts": {
-        "dev": "http-server ./ -p 7000 -c-1 --cors",
-        "server": "http-server ./ -p 7000 --cors",
+        "dev": "http-server ./ -c-1 --cors",
+        "server": "http-server ./ --cors",
         "lint:check": "eslint **/*.{js,mjs,jsx,ts,tsx}",
         "lint:fix": "eslint \"**/*.{js,mjs,jsx,ts,tsx}\" --fix",
         "pretty:check": "prettier --check ./",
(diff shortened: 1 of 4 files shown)


rniwa (approved):

flashdesignory:
should we also update the port when running speedometer in dev mode?

https://github.com/WebKit/Speedometer/blob/main/package.json#L18-L19

mdubet:
> should we also update the port when running speedometer in dev mode?
> 
> [`main`/package.json#L18-L19](https://github.com/WebKit/Speedometer/blob/main/package.json?rgh-link-date=2024-09-09T22%3A52%3A12Z#L18-L19)

The current patch changes both. (dev and server)

package.json
```diff
@@ -15,8 +15,8 @@
     },
     "license": "SEE LICENSE IN LICENSE",
     "scripts": {
-        "dev": "http-server ./ -p 7000 -c-1 --cors",
-        "server": "http-server ./ -p 7000 --cors",
+        "dev": "http-server ./ -c-1 --cors",
+        "server": "http-server ./ --cors",
```
flashdesignory, line 19: removing the dedicated port, will always choose the next open port on your machine. 
The updates in the readme files mention `8900`, which might not always be the port chosen on a user's machine. The same is the case for the charts and editors that use vite. 

That said, the removal here is correct, just the readme files need to reflect that somehow.

mdubet: The README change "suggests" to force the port to 8900 via `PORT=8900 npm run ...`.
(`http-server` uses the env var if it's defined - it chooses the next open port only if `$PORT` is not defined).

flashdesignory: sweet - thanks!

flashdesignory (approved):

PR #430, opened 2024-10-09
Add complexity param

camillobruni:
Depends on params.mjs changes in  #469

Work in progress proposal for adding a variable complexity param.
- `complexity === 1.0`: workloads run in default configuration
- `complexity > 1.0`: ideally workloads run proportionally slower
- `complexity < 1.0`: ideally workloads run proportionally fast

Background:
We're occasionally seeing issues with gc timing / optimisations heuristics with the current workloads. If we can dynamically change the duration / complexity of workloads it's a bit easier to assess how good these heuristics are and how much we're potentially overfitting.

Current Limitations:
- DeveloperMode UI cannot reload tests.mjs causing stale BenchmarkTestStep titles
- Only TodoMVC are currently scaling the workload size (other workloads are depend on whether we're fine with the general idea here)


resources/tests.mjs
```diff
@@ -66,7 +71,8 @@ Suites.push({
         page.getLocalStorage().getItem("javascript-es5");
     },
     tests: [
-        new BenchmarkTestStep(`Adding${numberOfItemsToAdd}Items`, (page) => {
+        new BenchmarkTestStep(`Adding${getNumberOfItemsToAdd()}Items`, (page) => {
```
camillobruni, line 74: Maybe I should just stick to using `defaultNumberOfItemsToAdd` here, so that the metric names don't change?

flashdesignory: I think the name should reflect the complexity level, in an effort to avoid comparing a simple test run with a complex test run

bgrins:
Without looking closely at the code change, I think this is a neat idea and I could imagine us increasing the complexity of charts, text editing, etc if we move forward with it. Will be curious for feedback from @julienw.

resources/tests.mjs
bgrins: Maybe we could remove the number here and handle this during reporting time - update the code that consumes the name after the run when reporting in the UI or JSON as either `TodoMVC-LocalStorage (Complexity 1.5)` -> `AddingItems`, or `TodoMVC-LocalStorage` -> `AddingItems (Complexity 1.5)` when complexity !== 1? One tradeoff there is it would report with complexity even in tests that aren't complexity-aware, but that seems maybe fine given that this is an opt-in thing that you'd only use if you know what you're doing.

@rniwa WDYT?

camillobruni: Definitely open to either suggestion.  `getNumberOfItemsToAdd()` would work partially if the complexity param is passed-in via url-params only (but that's a bit brittle).

resources/developer-mode.mjs
```diff
@@ -101,11 +102,21 @@ function createUIForSyncStepDelay() {
     return label;
 }
 
-function createTimeRangeUI(labelText, initialValue, unit = "ms", min = 0, max = 1000) {
+function createUIForComplexity() {
+    const { range, label } = createTimeRangeUI("Relative complexity", params.complexity, "x", 0, 10, 0.01);
```
julienw, line 106: I was thinking that rather than a float, maybe this can be an integer between 1 and 100, with 50 being the default? 

Then later you can divide this number by 50 to get the multiplier.
And you can cut down a lot of the changes.

But this is merely a suggestion, I'm also happy with what you did.

resources/tests.mjs
julienw: Naively I'd do this:
* add the complexity value somewhere in the json report, but not in all tests. Also display it somewhere in the results interface.
* remove any mention of actual counts in the test titles

But I'm not convinced this is right, because we might want the tests themselves to hold this value so that it's clear in reports that we're not at the default complexity.

Also if we want to change the actual titles, I'd like to see the complexity in the test names more than in the steps.

julienw (commented):
I like the idea :-)

resources/tests.mjs
camillobruni: - I can add a `non-standard-params` warning in the detailed results page to surface what's happening
- I'd vouch for dropping mentioning the complexity params (much like other score-influencing params, we don't surface them anywhere else)

camillobruni:
- Added table to highlight non-standard params
Screenshot 2024-10-11 at 14 15 17

resources/tests.mjs
camillobruni: In the latest commit I've just removed the reference to 100 in the step title.

resources/developer-mode.mjs
camillobruni: I kinda prefer a normalized float value, aiming quite literaly at a complexity-factor, 2x => it should run roughly twice as slow.

camillobruni:
I've now changed the complexity slider to have an exponential scale which makes it a bit more comfortable to use.

resources/newssite/news-next/dist/404.html
```diff
@@ -1 +1 @@
...
...01e5cf.css" as="style"/><link rel="stylesheet" href="./_next/static/css/a0dca1379a01e5cf.css" data-n-g=""/><noscript data-n-css=""></noscript><script defer="" nomodule="" src="./_next/static/chunks/polyfills-c67a75d1b6f99dc8.js"></script><script src="./_next/static/chunks/webpack-e50e9853db18b759.js" defer=""></script><script src="./_next/static/chunks/framework-2c79e2a64abdb08b.js" defer=""></script><script src="./_next/static/chunks/main-2ba37e62325cc71b.js" defer=""></script><script src="./_next/static/chunks/pages/_app-4fe6038a10d156ec.js" defer=""></script><script src="./_next/static/chunks/pages/_error-54de1933a164a1ff.js" defer=""></script><script src="./_next/static/3gZGkbShGQnNLB0qAzRia/_buildManifest.js" defer=""></script><script src="./_next/static/3gZGkbShGQnNLB0qAzRia/_ssgManifest.js" defer=""></script></head><body><div id="__next"></div><div id="settings-container"></div><div id="notifications-container"></div><script id="__NEXT_DATA__" type="application/json">{"props":{"pageProps":{"statusCode":404}},"page":"/_error","query":{},"buildId":"3gZGkbShGQnNLB0qAzRia","assetPrefix":".","nextExport":true,"isFallback":false,"gip":true,"scriptLoader":[]}</script></body></html>
```
rniwa, line 1: Why are these files getting re-generated?

camillobruni: If params.mjs is changed and thus we should update news-next as well since it's included as a helper package there.

resources/tests.mjs
```diff
@@ -1,9 +1,14 @@
 import { BenchmarkTestStep } from "./benchmark-runner.mjs";
 import { todos } from "./translations.mjs";
+import { params } from "./shared/params.mjs";
 
-const numberOfItemsToAdd = 100;
+const defaultNumberOfItemsToAdd = 100;
 const defaultLanguage = "en";
 
+function getNumberOfItemsToAdd() {
```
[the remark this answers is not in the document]
camillobruni: - Added `numberOfItemsToAdd` module variable
- Add `handleComplexityChange` to update `numberOfItemsToAdd` dynamically from the developer menu

julienw (commented):
Looks good!

can we also add the complexity value to the json?

resources/shared/params.mjs
```diff
@@ -70,17 +82,24 @@ export class Params {
...
         if (parsedValue < minValue)
             throw new Error(`Invalid ${paramKey} param: '${parsedValue}', value must be >= ${minValue}.`);
         searchParams.delete(paramKey);
         return parsedValue;
     }
 
+    _parseIntParam(searchParams, paramKey, minValue) {
+        const parsedValue = this._parserNumberParam(searchParams, paramKey, minValue);
+        if (!Number.isInteger(parsedValue))
+            throw new Error(`Invalid ${paramKey} param: '${parsedValue}', expected int.`);
```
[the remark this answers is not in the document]
camillobruni: Move the range check to the _parseNumber helper

resources/tests.mjs
```diff
@@ -1,9 +1,16 @@
...
+import { params } from "./shared/params.mjs";
 
-const numberOfItemsToAdd = 100;
+const defaultNumberOfItemsToAdd = 100;
 const defaultLanguage = "en";
 
+let numberOfItemsToAdd = defaultNumberOfItemsToAdd;
+
+export function handleComplexityChange() {
+    numberOfItemsToAdd = Math.abs(Math.round(defaultNumberOfItemsToAdd * params.complexity));
```
julienw, line 11: Not sure why I suggested `Math.abs`, of course I meant `Math.round` 😅 

`Math.abs` is probably not needed actually, this can't be negative I guess?

julienw (approved):
thank you, looks good to me

resources/tests.mjs
camillobruni: fair, we do check that in the params parsing already.
removing it.

flashdesignory (approved):

PR #432, opened 2024-10-09
Remove interactive runner

camillobruni:
The interactive runner has low added values these days:
- We can easily run individual workloads via the developerMode menu
- The control-flow in the async-await-based runner is much easier to follow and thus it's easier to set the right breakpoints

To simplify future runner refactorings I suggest removing the interactive runner support all together (and if we need some additional support, we should expose that in the developerMode)


flashdesignory (approved):

bgrins (approved):
Works for me, as discussed. I do find this somewhat convenient at times when developing tests, but we can find ways to simulate it with individual steps or perhaps in developer mode if we find it necessary to re-add.

julienw (approved):
I agree with removing it, but it would be good to have a replacement for the "step by step" way of running tests. I found this mode valuable when I developed my tests last year.

camillobruni:
Can't you easily use breakpoints for step-by-step debugging in the current runner (in `_runTestAndRecordResults`)?

rniwa:
It would be useful to have a step-by-step mode. I use it all the time while debugging Speedometer content.

camillobruni:
Ok, let me add a "Step Workload" button in the devMode menu, that shouldn't be too hard to implement.

camillobruni:
Updated the code and included a stepping button and functionality, please have a look.
I've now added the promise wait on the `willRunTest` client hook as it seemed more natural, but maybe the `didRunTest` approach as in the current interactive runner is better?

resources/main.mjs
```diff
@@ -17,17 +17,49 @@ class MainBenchmarkClient {
...
+        if (!this._isRunning) {
+            this._startBenchmark();
+            this._showSection("#running");
+        }
+    }
+
+    _clearStepping() {
+        this._steppingResolver();
+        this._steppingPromise = null;
+        this._steppingResolver = null;
```
rniwa, line 51: Shouldn't we be clearing these two states before calling `_steppingResolver`?

resources/developer-mode.mjs
```diff
@@ -222,14 +222,20 @@ function createUIForSuites() {
 }
 
 function createUIForRun() {
-    let button = document.createElement("button");
-    button.textContent = "Start Test";
-    button.onclick = (event) => {
+    const stepTestButton = document.createElement("button");
+    stepTestButton.textContent = "Step Test ⏯";
```
rniwa, line 226: Can we use `&#9199;` or '\u23EF' instead of directly using the unicode character?

resources/developer-mode.mjs
```diff
@@ -222,14 +222,20 @@ function createUIForSuites() {
...
+    stepTestButton.onclick = (event) => {
+        globalThis.benchmarkClient.step();
+    };
+    const startTestButton = document.createElement("button");
+    startTestButton.textContent = "Start Test ⏵";
```
rniwa, line 231: Ditto for `&#9205;` or `\u23F5`

rniwa (approved):

resources/main.mjs
camillobruni: sounds good.

resources/developer-mode.mjs
camillobruni: done.

PR #436, opened 2024-10-11
improve developerMode styling

camillobruni:
- Remove display flex that had no effect
- Make sure the run settings are always visible
- Scroll only the suites part
- Resize more

### DIFF
--- a/resources/main.css
+++ b/resources/main.css
@@ -10,6 +10,7 @@
     --text-width: 650px;
     --metrics-line-height: 25px;
     --scrollbar-width: 10px;
+    --developer-mode-settings-height: 280px;
 }
 
 body {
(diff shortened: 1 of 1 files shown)


camillobruni:
Tested on Firefox, Safari and Chrome for small windows (unlikely use-case):
Screenshot 2024-10-11 at 09 01 20

Large screen sizes now also allows to increase the developer-mode:

Screenshot 2024-10-11 at 09 02 43

flashdesignory (approved):

resources/main.css
```diff
@@ -212,36 +213,38 @@ button,
 }
 
 .developer-mode {
+    /* max-height: max(90vh, var(--viewport-height)); */
```
julienw, line 216: leftover?

julienw (approved):
I don't mind!

Ideally I'd also like to have the "select all / deselect all" buttons fixed, and only have the suites themselves in a scrollable container. Would that be easy to do?

camillobruni:
I could move the tags above the "select" buttons  and make the tags + suits scrollable?
Adding the tags would be the easiest, I just though that they might grow over time and we might have an issue long-term (but we can also deal with this later)

julienw:
> I could move the tags above the "select" buttons and make the tags + suits scrollable? Adding the tags would be the easiest, I just though that they might grow over time and we might have an issue long-term (but we can also deal with this later)

mmm it's mostly "select/unselect" for me, because it's common I want to unselect them all to run just one test.
but I'm not sure what would be best... Maybe keep them inside the scrollable element, but move them at the top so that they're readily available?

Let's not focus too much on that if there's no clear path and let's move on with what you already have.

camillobruni:
Actually your suggestions was quite easy to add:

Screenshot 2024-10-15 at 12 59 01

resources/main.css
camillobruni: removed.

PR #438, opened 2024-10-22
Add more flexible SUITE_RUNNER_LOOKUP

camillobruni:
- Add SUITE_RUNNER_LOOKUP to dynamically look up the suite runner class
- Add dummy AsyncSuiteRunner and RemoteSuiteRunner classes

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -457,7 +457,8 @@ export class BenchmarkRunner {
     async runSuite(suite) {
         // FIXME: Encapsulate more state in the SuiteRunner.
         // FIXME: Return and use measured values from SuiteRunner.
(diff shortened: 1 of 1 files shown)


flashdesignory (approved):
LGTM

resources/benchmark-runner.mjs
```diff
@@ -663,3 +664,16 @@ export class SuiteRunner {
             await this._client.didRunTest(suite, test);
     }
 }
+
+// FIXME: implement async steps
+class AsyncSuiteRunner extends SuiteRunner {}
```
rniwa, line 669: I don't think it makes much sense to split the suite runner between sync & async. async runner is simply a superset of sync runner.

julienw (approved):

bgrins:
Looks like this can merge now

PR #440, opened 2024-10-23
Add measurePrepare param

camillobruni:
- Include the suite prepare time in total time if param.debugMetrics is set
- Add SuiteRunner._suiteResults local variable

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -251,10 +251,8 @@ class MainBenchmarkClient {
         document.documentElement.style.setProperty("--metrics-line-height", `${trackHeight}px`);
         const plotWidth = (params.viewport.width - 120) / 2;
         const aggregateMetrics = [metrics.Geomean];
-        if (params.measurePrepare) {
+        if (params.measurePrepare)
             aggregateMetrics.push(metrics.Prepare);
-            console.log(metrics.Prepare);
-        }
         document.getElementById("aggregate-chart").innerHTML = renderMetricView({
             metrics: aggregateMetrics,
             width: plotWidth,


rniwa:
What is the purpose of this change? Why do we ever want to include the prepare time in the total?

resources/params.mjs
```diff
@@ -80,11 +82,10 @@ class Params {
...
         searchParams.delete("developerMode");
-
-        if (searchParams.has("useWarmupSuite")) {
-            this.useWarmupSuite = true;
-            searchParams.delete("useWarmupSuite");
-        }
+        this.useWarmupSuite = searchParams.has("useWarmupSuite");
+        searchParams.delete("useWarmupSuite");
+        this.debugMetrics = searchParams.has("debugMetrics");
+        searchParams.delete("debugMetrics");
```
julienw, line 88: why are we calling "delete" BTW? It's not clear to me...

julienw (commented):
Thanks for implementing my request!

To be honest I find the patch difficult to follow, because you've put a lot of things together in the same commit: how you deal with measuredValues, the checkbox UI, the prepare phase record. It would be good if you can split it in several meaningful commits. (no need to have separate PR).

Also I would like that we remove the first iteration when computing the average of the Prepare phase, because it's so different. What do you think?

resources/suite-runner.mjs
```diff
@@ -108,8 +111,10 @@ export class SuiteRunner {
         if (this.#suite === WarmupSuite)
             return;
 
-        const total = syncTime + asyncTime;
-        this.#suiteResults.tests[test.name] = { tests: { Sync: syncTime, Async: asyncTime }, total: total };
+        let total = syncTime + asyncTime;
+        if (this.#params.measurePrepare)
+            total += this.#prepareTime;
```
rniwa, line 116: I think if we did this, we should probably refuse to compute/show the final score. It's going to be very misleading otherwise. We don't want people to start reporting score with prepare time included as some kind of Speedometer score variant.

resources/main.mjs
```diff
@@ -250,8 +250,13 @@ class MainBenchmarkClient {
         const trackHeight = 24;
         document.documentElement.style.setProperty("--metrics-line-height", `${trackHeight}px`);
         const plotWidth = (params.viewport.width - 120) / 2;
-        document.getElementById("geomean-chart").innerHTML = renderMetricView({
-            metrics: [metrics.Geomean],
+        const aggregateMetrics = [metrics.Geomean];
+        if (params.measurePrepare) {
+            aggregateMetrics.push(metrics.Prepare);
+            console.log(metrics.Prepare);
```
rniwa, line 256: Is this console.log intentional??

PR #445, opened 2024-11-04
Add local results variable to SuiteRunner

camillobruni:
We can migrate some of the measuredValue logic directly to the SuiteRunner and simplify the code a bit.
Additionally this will allow us to generate more flexible metrics in the future.


julienw (approved):
thanks, nothing to say!

PR #448, opened 2024-11-07
News Site Nuxt - Article Ids

flashdesignory:
This is an alternate solution for: https://github.com/WebKit/Speedometer/pull/427

The reason to keep creating unique ids, is that it'll allow us to potentially opt into a complexity param (https://github.com/WebKit/Speedometer/pull/430), which could append content by cloning the sections in the data file. This is only possible, if we don't hard-code list ids in the data file (which was the previous solution). 

Here, I am just creating the unique ids once, when the data gets assigned to the app, instead of the currently used version that had issues (https://github.com/WebKit/Speedometer/issues/422).

Difference to the Next.js version:
You'll notice that the article-content already uses article.id in some instances in the main branch - this is not throwing any errors, but does assign `undefined`, since the article.id doesn't exist unless we create it somewhere (which is now done in the pr branch).

note: feel free to ignore the dist folder - that's just the generated build files

Crossbench - 10 iterations:

this pr
```
browser			Safari                    Firefox                Google Chrome
NewsSite-Nuxt  		61.10 ± 1.1%              67.35 ± 1.4%           60.95 ± 0.66%
NewsSite-Next  		58.98 ± 1.5%              78.5 ± 1.6%            75.4 ± 1.7%
Score          		16.72 ± 0.79%             13.80 ± 1.1%           14.78 ± 1.00%
```

main
```
browser			Safari                    Firefox                Google Chrome
NewsSite-Nuxt  		62.2 ± 2.6%               67.4 ± 1.7%            61.45 ± 1.2%
NewsSite-Next  		59.86 ± 1.4%              78.81 ± 1.2%           75.7 ± 2.1%
Score          		16.46 ± 1.8%              13.76 ± 1.0%           14.69 ± 1.5%
```


resources/newssite/news-nuxt/composables/provide-locale.js
```diff
@@ -16,10 +18,41 @@ export function provideLocale() {
...
+
+    const selected = Object.create(null);
+    Object.keys(content).forEach((key) => {
+        const { sections } = content[key];
+
+        const selectedSections = [];
+        let index = 0;
+
+        for (let i = 0; i < sections.length; i++) {
+            selectedSections.push({ ...sections[index] });
```
camillobruni, line 31: nit: probably use a temp here `const selectedSection = { ...sections[index] };`

resources/newssite/news-nuxt/composables/provide-locale.js
```diff
@@ -16,10 +18,41 @@ export function provideLocale() {
...
+            const { articles } = selectedSections[i];
+            for (let j = 0; j < articles.length; j++) {
+                articles[j].id = uuidv4();
+                const { content } = articles[j];
+                if (Array.isArray(content)) {
+                    for (let k = 0; k < content.length; k++) content[k].id = uuidv4();
+                }
+            }
+
+            index = (index + 1) % sections.length;
```
camillobruni, line 42: isn't this just the same as `i`?

camillobruni (approved):

resources/newssite/news-nuxt/composables/provide-locale.js
```diff
@@ -16,10 +18,41 @@ export function provideLocale() {
         htmlAttrs: { dir, lang },
     });
 
+    const { content } = dataSource[lang];
+
+    const selected = Object.create(null);
```
julienw, line 23: I'm not sure where this term of "selected" comes from, where we apply this to all contents. I'd use `contentWithIds` or something similar for more clarity.

resources/newssite/news-nuxt/composables/provide-locale.js
```diff
@@ -16,10 +18,41 @@ export function provideLocale() {
...
+    Object.keys(content).forEach((key) => {
+        const { sections } = content[key];
+
+        const selectedSections = [];
+
+        for (let i = 0; i < sections.length; i++) {
```
julienw, line 29: nit: rewrite this for loop as a `map` operation

julienw (changes requested):
The general approach looks good to me, but we need to avoid touching the original singleton to avoid surprises in the future.

Using `map` with some object spread operators should also make it more explicit that we're not touching it.

Thanks!

resources/newssite/news-nuxt/composables/provide-locale.js
```diff
@@ -16,10 +18,41 @@ export function provideLocale() {
...
+        const currentSections = sections.map((section) => {
+            const currentSection = { ...section };
+            currentSection.articles = section.articles.map((article) => {
+                const currentArticle = { ...article };
+                currentArticle.id = uuidv4();
+                if (Array.isArray(article.content)) {
+                    currentArticle.content = article.content.map((item) => {
+                        const currentItem = { ...item };
+                        currentItem.id = uuidv4();
+                        return currentItem;
```
julienw, line 37: I wanted to (optionally) suggest this:
```suggestion
                        return {
                          ...item,
                          id: uuidv4();
                        };
```
but I'm not sure this looks better.

julienw (approved):
thanks!
note you'll need to rebase and regenerate the dist

resources/newssite/news-nuxt/composables/provide-locale.js
flashdesignory: I know what you mean. I think I'd rather keep things consistent (visually) and treat currentArticle and currentItem the same way.

rniwa (approved):
Seems sane to me.

PR #452, opened 2024-11-14
Runner Cleanup: SuiteRunner & TestRunner classes

flashdesignory:
- move SuiteRunner class into a separate file
- create a new TestRunner class, which a remote workload could consume in the future
- rename _runTestAndRecordResults to TestRunner.runTest()
- create class methods for functions used in the runSync & measureAsync functions

### DIFF
--- a/resources/test-runner.mjs
+++ b/resources/test-runner.mjs
@@ -10,59 +10,6 @@ export class TestRunner {
         this._frame = frame;
     }
 
-    _runWarmupSuite() {
-        if (this._params.warmupBeforeSync) {
-            performance.mark("warmup-start");
-            const startTime = performance.now();
-            // Infinite loop for the specified ms.
-            while (performance.now() - startTime < this._params.warmupBeforeSync)
-                continue;
-            performance.mark("warmup-end");
-        }
-    }
-
-    _measureSyncTime({ syncStartLabel, syncEndLabel }) {
-        performance.mark(syncStartLabel);
-        const syncStartTime = performance.now();
-        this._test.run(this._page);
-        const syncEndTime = performance.now();
-        performance.mark(syncEndLabel);
-
-        const syncTime = syncEndTime - syncStartTime;
-        return syncTime;
-    }
-
-    _initAsyncTime({ asyncStartLabel }) {
-        performance.mark(asyncStartLabel);
-
-        const asyncStartTime = performance.now();
-        return asyncStartTime;
-    }
-
-    _measureAsyncTime({ asyncStartTime, asyncEndLabel }) {
-        const asyncEndTime = performance.now();
-        performance.mark(asyncEndLabel);
-
-        const asyncTime = asyncEndTime - asyncStartTime;
-        return asyncTime;
-    }
-
-    _measureTestPerformance({ suiteName, testName, syncStartLabel, syncEndLabel, asyncStartLabel, asyncEndLabel, params }) {
-        if (this._params.warmupBeforeSync)
-            performance.measure("warmup", "warmup-start", "warmup-end");
-        performance.measure(`${suiteName}.${testName}-sync`, syncStartLabel, syncEndLabel);
-        performance.measure(`${suiteName}.${testName}-async`, asyncStartLabel, asyncEndLabel);
-    }
-
-    _forceLayout() {
-        const bodyReference = this._frame ? this._frame.contentDocument.body : document.body;
-        const windowReference = this._frame ? this._frame.contentWindow : window;
-        // Some browsers don't immediately update the layout for paint.
-        // Force the layout here to ensure we're measuring the layout time.
-        const height = bodyReference.getBoundingClientRect().height;
-        windowReference._unusedHeightValue = height; // Prevent dead code elimination.
-    }
-
     async runTest() {
         // Prepare all mark labels outside the measuring loop.
         const suiteName = this._suite.name;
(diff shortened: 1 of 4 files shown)


bgrins:
I like the direction here - would like @julienw to review the change if possible.

@flashdesignory to confirm - the big picture idea here is that remote tests will (eventually) consume resources/test-runner.mjs themselves, and it will grow functionality to manage communication back and forth with the suite (postMessage etc)?

flashdesignory:
> @flashdesignory to confirm - the big picture idea here is that remote tests will (eventually) consume resources/test-runner.mjs themselves, and it will grow functionality to manage communication back and forth with the suite (postMessage etc)?

remote tests would use the test-runner class, that's correct 😄.
postMessage communication will happen on a higher level in the workloads, not on a test-runner level.

resources/test-runner.mjs
```diff
@@ -0,0 +1,96 @@
+import { TEST_INVOKER_LOOKUP } from "./test-invoker.mjs";
+
+export class TestRunner {
+    constructor(suite, test, params, callback, page, frame) {
```
rniwa, line 4: Can we use a consistent parameter order as SuiteRunner? so "frame, page, params"

resources/test-runner.mjs
```diff
@@ -0,0 +1,96 @@
...
+        this._suite = suite;
```
rniwa, line 5: We should use real private instance variables.

resources/test-runner.mjs
```diff
@@ -0,0 +1,96 @@
...
+        };
+
+        const report = () => this._callback(this._test, syncTime, asyncTime);
+        const invokerClass = TEST_INVOKER_LOOKUP[this._params.measurementMethod];
+        const invoker = new invokerClass(runSync, measureAsync, report, this._params);
+
+        return invoker.start();
+    }
+
+    _runWarmupSuite() {
```
rniwa, line 44: This isn't really a suite. We should probably just call it _runWarmup or something

resources/test-runner.mjs
```diff
@@ -0,0 +1,96 @@
...
+        const syncEndLabel = `${suiteName}.${testName}-sync-end`;
+        const asyncStartLabel = `${suiteName}.${testName}-async-start`;
+        const asyncEndLabel = `${suiteName}.${testName}-async-end`;
+
+        let syncTime;
+        let asyncStartTime;
+        let asyncTime;
+        const runSync = () => {
+            this._runWarmupSuite();
+            syncTime = this._measureSyncTime(syncStartLabel, syncEndLabel);
```
rniwa, line 28: This code is harder to follow with nested function calls. We should go back to having all the code here instead.

resources/test-runner.mjs
```diff
@@ -0,0 +1,96 @@
...
+            asyncStartTime = this._initAsyncTime(asyncStartLabel);
+        };
+        const measureAsync = () => {
+            this._forceLayout();
```
rniwa, line 32: Ditto.

resources/test-runner.mjs
flashdesignory: I was trying to find a good way to be able to reuse as much as possible for remote workloads. 
If I keep everything in one function, I'll end up overriding the whole function, when I might just need to change one line in one nested function call. 

With this, I should be able to create a new class that extends TestRunner, but only overrides one method, which is more targeted and introduces less CLs.

rniwa: Which is the line that needs to be changed for remote workloads?

flashdesignory: Some examples, that are slightly different for remote workloads:

`this._test.run(this._page); `
The remote workloads don't use a page object to wrap elements. 
Also, if we opt into async workloads, we might want to await the run function. 

`_forceLayout()`
It works, although I won't use or pass in a frame reference.

These are just some examples

flashdesignory: Both examples can work as is (even if page and frame are not passed in), but I still feel that smaller methods might be easier to follow and reuse. 

This shouldn't be a blocker though and if I'm the only one feeling this way, I can certainly change it.

bgrins: Perhaps we should make it more of a copy/paste for this PR and then follow up with refactors? I think it'll be easier to say one way or the other with (a) the change unbundled from the file move and (b) when we have a remote implementation to consider.

rniwa: Yeah, doing the split as a follow up seems like a good idea here.

flashdesignory: sounds good - i'll fix it up!

resources/test-runner.mjs
flashdesignory: changed order for both SuiteRunner and TestRunner

resources/test-runner.mjs
flashdesignory: done

resources/test-runner.mjs
flashdesignory: done

resources/test-runner.mjs
flashdesignory: done

resources/test-runner.mjs
flashdesignory: reverted

rniwa (approved):

bgrins (approved):
I'll go ahead and approve in place of Julien with the simplified change in place

julienw:
FTR looks good to me!

PR #456, opened 2024-11-19
Remote Runner / PostMessge api

flashdesignory:
This pr introduces Remote Runner workloads. 
Workloads that are local OR that point to a remote url (when running Speedometer locally), can communicate and run their own tests via PostMessage.

Basic flow:

1. SuiteRunner loads the iframe of the local or remote workload.
2. SuiteRunner waits for the "app-ready" message from the workload.
3. Workload sends "app-ready" message, once its able to listen and receive a command to start the tests.
4. SuiteRunner sends the "benchmark-suite" command to start the testing (with an optional suite name).
5. Workload runs through the test suite and collects performance metrics.
6. After each completed test step it sends a "step-complete" message, to update progress in speedometer.
7. With the "suite-complete" message, the workload sends the metrics along.
8. Speedometer receives metrics.

image

Notes
- TestRunner and TestInvoker files are installed as dependencies in the news-site-next workload, to ensure they stay synced with updates.


resources/newssite/news-next/public/test-invoker.mjs
```diff
@@ -0,0 +1,55 @@
+class TestInvoker {
```
bgrins, line 1: I was surprised this is in public/ since you are loading this file via npm (I would have expected it to only exist in dist/). Is this left over or is this how npm works?

resources/newssite/news-next/public/test-runner.mjs
```diff
@@ -0,0 +1,79 @@
+import { TEST_INVOKER_LOOKUP } from "./test-invoker.mjs";
```
bgrins, line 1: Ditto

resources/newssite/news-next/public/test-invoker.mjs
flashdesignory: yeah, so basically I'm doing this:

1. install dependency through [package.json](https://github.com/WebKit/Speedometer/blob/ee3ceb4c4598ae7517d6a403e99cefdb56a38a53/resources/newssite/news-next/package.json#L24) (this copies the two files in my node_modules folder)
2. run a [script](https://github.com/WebKit/Speedometer/blob/ee3ceb4c4598ae7517d6a403e99cefdb56a38a53/resources/newssite/news-next/scripts/prepare.js#L39) to copy both files to the public folder

The reason is that the code needs to execute in the hosted directory when you visit the workload. 
The node_modules folder is not available at that point.

bgrins: if you were to update the `<script>` src in the development source to point to the full node_modules path wouldn't the build step be able to figure it out to get it into dist without the extra script to copy the file? Not sure how the build works exactly but that's what I figured would happen conceptually

flashdesignory: let me take a look if I can simplify....

flashdesignory: @bgrins - looking at the code, the files I am referencing in the `script` component are all files created by the news-site-workload:

- benchmark-connector.mjs
- workload-test.mjs
- workload-testing-utils.mjs

The last file imports the test-runner, but since I am using the script component, I have to put the 3 files above in the public folder. This is because of the way Next.js resolves file paths.

Second problem: I can't reference an import from the node_modules folder, since the three files above are in the public folder, which doesn't have access to the node_modules.

bgrins: As a test, I ran `npm create vite@latest` and then installed the entire "speedometer" package rather than one-by-one mjs, then referenced it as `import { TestRunner } from 'speedometer/resources/test-runner.mjs'` and it seems to work in both dev and build (see screenshot). I assume with this approach we could also move workload-testing-utils.mjs into the main resources dir instead of implementing it directly in the workload.

I'm not quite sure how Next `<Script>` elements integrate with this, but presumably they have a way to reference something from an installed package (maybe make the script element reference a local file and then have that file ESM import the relevant helpers). Or maybe I'm missing something?

Screenshot 2024-11-21 at 10 33 55 AM

bgrins: #459 shows an example of this

flashdesignory: we need a solution that works for various frameworks though. 
For Next.js, I can't point the Script component to the node_modules folder... 
for example below won't work

```
<Script src="node_modules/benchmark-connector.mjs"...>
```

flashdesignory: > #459 shows an example of this

yeah imports are different from static js files though.

resources/newssite/news-next/dist/benchmark-connector.mjs
```diff
@@ -0,0 +1,97 @@
+/*
+    Note: This will eventually come from a separate package and should be minified.
+*/
+window.name = "news-site-next";
+window.version = "1.0.0";
+
+/** **********************************************************************
+ * Params
```
bgrins, line 8: @flashdesignory I don't see the following logic used in the main runner. Since it also reads params from the URL, would it be possible to share some code here instead?

flashdesignory: we could - the params.mjs is slightly different, but I can take a look at both and see if I can clean up and reuse

resources/newssite/news-next/public/test-invoker.mjs
flashdesignory: @bgrins was convincing 😄 .... refactored and introduced a shared folder that's installable.

bgrins:
Now that this has mostly incorporated #459 (I'd still like to see the params sharing) I think it's going to be easier to review & hopefully get landed for testing @rniwa @julienw

resources/newssite/news-next/src/workload-test.mjs
```diff
@@ -0,0 +1,41 @@
+import { BenchmarkTestStep, BenchmarkTestSuite, BenchmarkSuitesManager, forceLayout, getElement } from "speedometer-utils/workload-testing-utils.mjs";
+
+export function initWorkload() {
+    window.benchmarkSuitesManager = new BenchmarkSuitesManager(window.name, [
```
bgrins, line 4: I guess we could actually just pass this object as a parameter into the call to `connectToBenchmark` now that we're just calling the function (instead of sticking it on the window object). So just have this module call connectToBenchmark directly instead of sticking it on benchmarkSuitesManager and then having the App object call both. WDTY @flashdesignory ?

flashdesignory: done!

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,153 @@
...
+
+    window.onmessage = async (event) => {
+        // ensure we only let legit functions run...
+        if (event.data.id !== appId || event.data.key !== "benchmark-connector")
+            return;
+
+        switch (event.data.type) {
+            case "benchmark-suite":
+                // eslint-disable-next-line no-case-declarations
+                const { params } = await import("./params.mjs");
```
flashdesignory, line 143: for SSR compatibility!

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,154 @@
...
+            return;
+
+        switch (event.data.type) {
+            case "benchmark-suite":
+                // eslint-disable-next-line no-case-declarations
+                const params = new Params(new URLSearchParams(window.location.search));
+                // eslint-disable-next-line no-case-declarations
+                const { result } = await benchmarkSuitesManager.getSuiteByName(event.data.name).runAndRecord(params, (test) => sendMessage({ type: "step-complete", status: "success", appId, name, test }));
+                sendMessage({ type: "suite-complete", status: "success", appId, result });
+                break;
```
julienw, line 148: nit: add a default clause with a console.error line writing the received `event.data.type`

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * postMessage is used to communicate between app and benchmark.
+ * When the app os ready, an 'app-ready' message is sent to signal that the app can receive instructions.
+ *
+ * A prepare script within the apps appends window.name and window.version from the package.json file.
+ * The appId is build by appending name-version
+ * It's used as an additional safe-guard to ensure the correct app responds to a message.
+ *************************************************************************/
+export function connectToBenchmark(benchmarkSuitesManager, name, version) {
+    const appId = name && version ? `${name}-${version}` : -1;
```
julienw, line 130: is it an error that we don't have name and version? Should we throw or at least output a warning?

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,154 @@
...
+    const rect = document.body.getBoundingClientRect();
+    const e = document.elementFromPoint((rect.width / 2) | 0, (rect.height / 2) | 0);
+    return e;
+}
+
+/** **********************************************************************
+ * Benchmark Connector
+ *
+ * postMessage is used to communicate between app and benchmark.
+ * When the app os ready, an 'app-ready' message is sent to signal that the app can receive instructions.
```
julienw, line 123:
```suggestion
 * When the app is ready, an 'app-ready' message is sent to signal that the app can receive instructions.
```

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,154 @@
...
+    getSuiteByName(name) {
+        return this.suites.find((suite) => suite.name === name);
+    }
+}
+
+/**
+ * Helper Methods
+ *
+ * Various methods that are extracted from the Page class.
+ */
```
julienw, line 92: could the Page class use them directly (maybe as a follow-up, I don't know)
I don't feel comfortable with the duplication.

resources/suite-runner.mjs
```diff
@@ -89,13 +111,101 @@ export class SuiteRunner {
...
+
+    async _prepareSuite() {
+        const suiteName = this.suite.name;
+        const suitePrepareStartLabel = `suite-${suiteName}-prepare-start`;
+        const suitePrepareEndLabel = `suite-${suiteName}-prepare-end`;
+
+        performance.mark(suitePrepareStartLabel);
+
+        // Wait for the app-ready message from the workload.
+        const promise = this._subscribeOnce("app-ready");
```
julienw, line 154:
```suggestion
        const appReadyPromise = this._subscribeOnce("app-ready");
```

resources/suite-runner.mjs
```diff
@@ -89,13 +111,101 @@ export class SuiteRunner {
...
+        // Update progress bar with each completed step.
+        this._startSubscription("step-complete", async (e) => this._updateClient(e.data.test, e.data.name));
+        // Ask workload to run its own tests.
+        this.frame.contentWindow.postMessage({ id: this.appId, key: "benchmark-connector", type: "benchmark-suite", name: this.suite.config?.name || "default" }, "*");
+        // Capture metrics from the completed tests.
+        const response = await this._subscribeOnce("suite-complete");
+        this._stopSubscription("step-complete");
+
+        this.suiteResults.tests = response.result.tests;
+        this.suiteResults.total += response.result.total;
```
julienw, line 176: why are we overriding the `tests` property but incrementing the `total` value?
Should we append the `tests` instead?
(I haven't researched closely but it would be good to comment about that)

resources/shared/params.mjs
```diff
@@ -146,21 +146,25 @@ class Params {
         return shuffleSeed;
     }
 
-    toSearchParams() {
+    toSearchParams(forRemote = false) {
```
julienw, line 149: can we have an option object parameter instead of a boolean? it's really not clear what this does from the caller site.

(and actually from here neither, a comment would be nice)

resources/shared/params.mjs
```diff
@@ -146,21 +146,25 @@ class Params {
...
+            delete rawParams["tags"];
+        }
         return new URLSearchParams(rawParams).toString();
     }
 }
 
 export const defaultParams = new Params();
 
-const searchParams = new URLSearchParams(window.location.search);
+const searchParams = new URLSearchParams(typeof window !== "undefined" ? window.location.search : undefined);
```
julienw, line 163: when is it happening that window isn't present?

optional nit: use `?.` operator

julienw (approved):
I haven't tried locally but this looks pretty good to me. We can always fix things later if there's anything broken anyway.

r+ with my comments fixed.

Note: it would be a lot easier to review if you'd reworked the PR to have consistent commits, for example:

Commit 1: move files around without changing them
Commit 2: add the remote runner + the workload tools
Commit 3: change the workload itself

If you do this for future complex PR this will make them a joy to review (and hence faster to review as well). Thanks!

resources/newssite/news-next/src/pages/index.js
```diff
@@ -1,10 +1,16 @@
...
+import { useLayoutEffect } from "react";
 import { HashRouter as Router, Routes, Route } from "react-router-dom";
 import Page from "@/partials/page/page";
 import Head from "next/head";
 import { DataContextProvider } from "@/context/data-context";
+import { connectToBenchmark } from "speedometer-utils/workload-testing-utils.mjs";
+import { getBenchmarkSuitesManager } from "@/workload-test.mjs";
 
 export default function App() {
+    useLayoutEffect(() => {
```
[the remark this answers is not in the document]
flashdesignory: Added a comment. Basically, we want to wait until the workload is ready to accept a test suite. 
Waiting until all DOM mutations are done ensures that elements are present that are needed.
The Workload should decide when the 'app-ready' message should be sent to the benchmark.

resources/newssite/news-next/src/pages/index.js
```diff
@@ -1,10 +1,16 @@
...
+        connectToBenchmark(getBenchmarkSuitesManager(), "news-next", 1);
```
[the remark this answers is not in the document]
flashdesignory: added a cleanup function. It's less important here, but I am now also disconnecting after the "suite-complete' message has been sent.

resources/suite-runner.mjs
```diff
@@ -25,6 +23,30 @@ export class SuiteRunner {
...
+    }
+
+    get client() {
+        return this.#client;
+    }
+
+    get suiteResults() {
+        return this.#suiteResults;
+    }
+
```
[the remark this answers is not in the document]
flashdesignory: I feel like this should be a discussion point outside of the current prs.

resources/newssite/news-next/src/pages/index.js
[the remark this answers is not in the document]
julienw: but why useEffect and not useLayoutEffect?

resources/shared/workload-testing-utils.mjs
flashdesignory: Agreed - This should be a follow up!

resources/suite-runner.mjs
flashdesignory: done

resources/suite-runner.mjs
flashdesignory: cleaned this up - thank you!

resources/shared/params.mjs
flashdesignory: I added some comments in that function.
We don't use option params in our current code base and it feels too much, just for readability. 
We can rename "forRemote" or I can also add a comment on the caller site, if that helps.

julienw: a comment on the caller site would be good, thanks

resources/newssite/news-next/src/pages/index.js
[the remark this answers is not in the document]
flashdesignory: You're right with questioning me 😄 - switched to useEffect since that waits for the paint.

camillobruni:
Still looking good from my side.

resources/shared/params.mjs
```diff
@@ -146,21 +146,28 @@ class Params {
...
-    toSearchParams() {
+    toSearchParams(forRemote = false) {
         const rawParams = { ...this };
         rawParams["viewport"] = `${this.viewport.width}x${this.viewport.height}`;
+
+        // Only returning params that are useful for the workload.
+        // Both, 'suites' and 'tags', are specific to the debug menu.
+        if (forRemote) {
+            delete rawParams["suites"];
+            delete rawParams["tags"];
```
rniwa, line 157: I'd prefer having an allowed list of params instead of letting every param not listed here pass through.

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,166 @@
+import { TestRunner } from "./test-runner.mjs";
+import { Params } from "./params.mjs";
+
+/**
+ * BenchmarkTestStep
+ *
+ * A single test step, with a common interface to interact with.
+ */
+export class BenchmarkTestStep {
```
rniwa, line 9: "Test" doesn't seem to add much value here. Why not just `BenchmarkStep`?

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,166 @@
...
+        return result;
+    }
+}
+
+/**
+ * BenchmarkTestSuite
+ *
+ * A single test suite that contains one or more test steps.
+ */
+export class BenchmarkTestSuite {
```
rniwa, line 27: Ditto for BenchmarkSuite.

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,166 @@
...
+        };
+    }
+}
+
+/**
+ * BenchmarkSuitesManager
+ *
+ * A collection of test suites for a single workload.
+ */
+export class BenchmarkSuitesManager {
```
rniwa, line 77: Why do we need this? It doesn't seem like we're gonna have more than one suite per document?

resources/newssite/news-next/src/pages/index.js
```diff
@@ -1,10 +1,19 @@
-import React from "react";
+import { useEffect } from "react";
 import { HashRouter as Router, Routes, Route } from "react-router-dom";
 import Page from "@/partials/page/page";
 import Head from "next/head";
 import { DataContextProvider } from "@/context/data-context";
+import { connectToBenchmark } from "speedometer-utils/workload-testing-utils.mjs";
+import { getBenchmarkSuitesManager } from "@/workload-test.mjs";
```
rniwa, line 7: So a bunch of tests are all going to have their own getBenchmarkSuitesManager?
That would make grep'ing harder. Why not name this after each suite?
e.g. `import { NewsNextSuite } from "@/workload-test.mjs";`

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * postMessage is used to communicate between app and benchmark.
+ * When the app is ready, an 'app-ready' message is sent to signal that the app can receive instructions.
+ *
+ * A prepare script within the apps appends window.name and window.version from the package.json file.
+ * The appId is build by appending name-version
+ * It's used as an additional safe-guard to ensure the correct app responds to a message.
+ *************************************************************************/
+
+export function connectToBenchmark(benchmarkSuitesManager, name, version) {
```
rniwa, line 130: I'd rather have this as a class like `BenchmarkConnector` or maybe `BenchmarkRunnerProxy`.

resources/shared/workload-testing-utils.mjs
```diff
@@ -0,0 +1,166 @@
...
+        this.suites = suites;
+    }
+
+    getSuiteByName(name) {
+        return this.suites.find((suite) => suite.name === name);
+    }
+}
+
+/**
+ * Helper Methods
```
rniwa, line 89: I'd rather have a separate file for these helper methods and rename this file to "benchmark.mjs" or something since all classes here will have Benchmark prefix then.

resources/suite-runner.mjs
```diff
@@ -89,13 +112,107 @@ export class SuiteRunner {
...
+            ...response.result.tests,
+        };
+
+        this.suiteResults.total += response.result.total;
+
+        this._validateSuiteTotal();
+    }
+
+    _handlePostMessage(e) {
+        if (this.postMessageCallbacks.has(e.data.type))
```
rniwa, line 189: Can we call get here and use a local variable?
```js
if (count callback = this.postMessageCallbacks.get(e.data.type))
    callback(e);
```

resources/suite-runner.mjs
```diff
@@ -89,13 +112,107 @@ export class SuiteRunner {
...
+            ...this.suiteResults.tests,
+            ...response.result.tests,
+        };
+
+        this.suiteResults.total += response.result.total;
+
+        this._validateSuiteTotal();
+    }
+
+    _handlePostMessage(e) {
```
rniwa, line 188: Please don't use a single letter variable name like "e". Please spell out event.

resources/tests.mjs
```diff
@@ -894,6 +894,19 @@ Suites.push({
...
+Suites.push({
+    name: "NewsSite-PostMessage",
+    url: "resources/newssite/news-next/dist/index.html",
+    tags: ["experimental", "newssite", "language"],
+    disabled: true,
+    async prepare() {},
+    type: "remote",
+    config: {
+        // name: "default", // optional param to target non-default tests locally
+        steps: 3, // manual step for the progress display on the bottom: this is the number of BenchmarkSteps in a BenchmarkSuite.
```
rniwa, line 906: Hm... so whenever we update benchmark suite, we need to remember to update this magic number?
That doesn't sound like a scalable solution.

Can we load `resources/newssite/news-next/src/workload-test.mjs` in the harness as well and get the number of suites from there?

Alternatively, can we define some kind of JSON file somewhere which outlines each step of suite?

resources/shared/workload-testing-utils.mjs
flashdesignory: There is a reason for this. 
Each workload has a 'default' suite, the one that runs by default in Speedometer. 

This setup allows for one workload to have multiple suites. one default and let's say an experimental one. 

It lets us reuse the same workload, with different suites to run.

resources/newssite/news-next/src/pages/index.js
flashdesignory: A workload can have multiple suites, but we could name it after the workload.
What's the benefit of renaming it?

resources/shared/params.mjs
flashdesignory: An allowlist is something we'd need to maintain, vs here we just remove what breaks the remote workload.

resources/shared/workload-testing-utils.mjs
flashdesignory: done

resources/shared/workload-testing-utils.mjs
flashdesignory: done

resources/tests.mjs
camillobruni: An alternative solution we discussed is a little shared helper library that communicates back the progress / step sizes on Test setup.
Given that we don't really have many workloads yet I'd prefer sticking to slightly more cumbersome manual approach and work towards a more maintainable solution in the long run.

resources/suite-runner.mjs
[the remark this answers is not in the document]
camillobruni: I tend to be on your side here julien... but I totally fine with cleaning / discussing this after landing this PR :)

resources/suite-runner.mjs
```diff
@@ -89,13 +112,108 @@ export class SuiteRunner {
...
+export class RemoteSuiteRunner extends SuiteRunner {
+    #appId;
+
+    get appId() {
+        return this.#appId;
+    }
+
+    set appId(id) {
+        this.#appId = id;
+    }
```
camillobruni, line 133: this can definitely be just a public field if we have public getters *and* setters.

resources/tests.mjs
bgrins: +1 on doing this as a followup. We can discuss there whether JSON manifest with hardcoded step count or encoding it in the setup message would be preferable.

resources/shared/params.mjs
```diff
@@ -108,7 +108,7 @@ class Params {
         return defaultParams.suites;
     }
 
-    _parseTags() {
+    _parseTags(searchParams) {
```
flashdesignory, line 111: oversight?

resources/shared/params.mjs
flashdesignory: we fixed the params obj without filtering keys.

resources/suite-runner.mjs
flashdesignory: implemented above

resources/shared/workload-testing-utils.mjs
flashdesignory: as discussed, using a 'suites' object now.

resources/newssite/news-next/src/pages/index.js
flashdesignory: with the mentioned suites object, which is exported as default, we could easily rename the import if that's desired.

resources/tests.mjs
flashdesignory: removed steps magic number!

resources/newssite/news-next/src/workload-test.mjs
```diff
@@ -0,0 +1,41 @@
+import { BenchmarkStep, BenchmarkSuite, forceLayout, getElement } from "speedometer-utils/workload-testing-utils.mjs";
```
rniwa, line 1: Don't we need to import from benchmark.mjs and helpers.mjs now that they're split?

resources/shared/benchmark.mjs
```diff
@@ -0,0 +1,127 @@
...
+ */
+export class BenchmarkStep {
+    constructor(name, run) {
+        this.name = name;
+        this.run = run;
+    }
+
+    async runAndRecord(params, suite, test, callback) {
+        const testRunner = new TestRunner(null, null, params, suite, test, callback);
+        const result = await testRunner.runTest();
```
rniwa, line 18: Do we need an explicit await here? Can't we just return the result of `testRunner.runTest()`?

resources/newssite/news-next/src/workload-test.mjs
flashdesignory: the imported files re-exports them

resources/suite-runner.mjs
```diff
@@ -25,6 +23,30 @@ export class SuiteRunner {
         this.#params = params;
     }
 
+    get frame() {
```
rniwa, line 26: Do we really need to expose all these methods now?
It doesn't seem like this is ever called anywhere??
Ditto for other methods in this class.

rniwa (approved):

resources/newssite/news-next/src/workload-test.mjs
rniwa: Do we really need that? Directly importing from benchmark.mjs & helpers.mjs seem better?

resources/shared/benchmark.mjs
flashdesignory: yes, since the RAFTestInvoker has this line:

`
const result = await this._reportCallback();
`

resources/suite-runner.mjs
flashdesignory: yeah, since the RemoteSuiteRunner extends the SuiteRunner and the private vars are not available in the RemoteSuiteRunner.

I have to use getters to be able to reference the frame like:
`this.frame.contentWindow.postMessage`

rniwa: Oh, I missed that these are getters, not methods.

resources/newssite/news-next/src/workload-test.mjs
flashdesignory: changed to two imports

PR #457, opened 2024-11-19
add message-channel

flashdesignory:
Alternate solution to discuss, which might help mitigate https://github.com/WebKit/Speedometer/issues/441

Issue #83, opened 2023-03-03
Add the ability to have async steps


bgrins:
There are some workloads that we hard to capture synchronously in a BenchmarkTestStep. Areas this has come up

* Moanco (https://github.com/WebKit/Speedometer/pull/81#issue-1599855829) since it relies on a worker to do some processing
* React-Stockcharts (https://github.com/WebKit/Speedometer/pull/11#issuecomment-1366851864), and @camillobruni has raised some concerns with bimodal distribution on Chrome that is quite possibly related.
* In the research on data grids I see some grid libraries that delay rendering and so spend 0ms during the step. I've also experimented with and think we should consider using in-memory SQLite WASM as a backend for these tests, and that runs in a worker so async is a must-have.
* In the research on charting libraries that @julienw has been doing - where certain libraries delay rendering (especially React-based wrappers AIUI) in a way that causes significant work to happen outside of the timed window.

@rniwa and I discussed a bit on WebKit Slack, and concluded

> yeah, making the sync step compatible with promise seems okay to me. We just need to make sure we don't accidentally start measuring truly async stuff like network loads and such.

I don't know mechanically what will be involved with a change here, and the main runner function comments about not using Promise https://github.com/WebKit/Speedometer/blob/24b094d1cc620be0fecb04e31863dba851fa900e/resources/benchmark-runner.mjs#L230. There's also https://github.com/WebKit/Speedometer/pull/28#discussion_r1054531972 which may or may not make sense to tackle at the same time if we're going to be tweaking the core measurement function.

bgrins:
Status update: @camillobruni mentioned last meeting that was on his radar in the coming weeks

rniwa:
Sounds like we want to push this to v4 based on this week's sync?

flashdesignory:
@bgrins - can we close this issue?

PR #458, opened 2024-11-20
Async Runner with Message Channel

flashdesignory:
As discussed in our sync, here is a combination of the [AsyncRunner](https://github.com/WebKit/Speedometer/pull/442) and [Message Channel](https://github.com/WebKit/Speedometer/pull/457) pr.

Relevant code snippet (udpated):

```
class AsyncRAFTestInvoker extends BaseRAFTestInvoker {
    static mc = new MessageChannel();
    _scheduleCallbacks(resolve) {
        let gotTimer = false;
        let gotMessage = false;
        let gotPromise = false;

        const tryTriggerAsyncCallback = () => {
            if (!gotTimer || !gotMessage || !gotPromise)
                return;

            this._asyncCallback();
            setTimeout(async () => {
                await this._reportCallback();
                resolve();
            }, 0);
        };

        requestAnimationFrame(async () => {
            await this._syncCallback();
            gotPromise = true;
            tryTriggerAsyncCallback();
        });

        requestAnimationFrame(() => {
            setTimeout(() => {
                gotTimer = true;
                tryTriggerAsyncCallback();
            });

            AsyncRAFTestInvoker.mc.port1.addEventListener(
                "message",
                function () {
                    gotMessage = true;
                    tryTriggerAsyncCallback();
                },
                { once: true }
            );
            AsyncRAFTestInvoker.mc.port1.start();
            AsyncRAFTestInvoker.mc.port2.postMessage("speedometer");
        });
    }
}
```

For testing purposes I kept an explicit type of "async" in the test file, but I opted all default suites into using it. 

Initial testing shows that it should also fix the react specific issue.

Other issue solved by this pr: https://github.com/WebKit/Speedometer/issues/83


resources/test-invoker.mjs
```diff
@@ -48,8 +67,51 @@ export class RAFTestInvoker extends TestInvoker {
...
+
 export const TEST_INVOKER_LOOKUP = {
     __proto__: null,
     timer: TimerTestInvoker,
     raf: RAFTestInvoker,
 };
+
+export const ASYNC_TEST_INVOKER_LOOKUP = {
+    __proto__: null,
+    timer: AsyncTimerTestInvoker,
```
flashdesignory, line 115: If we don't want an AsyncTimerTestInvoker, I could simply assign the same AsyncRAFTestInvoker..

resources/test-invoker.mjs
```diff
@@ -48,8 +67,51 @@ export class RAFTestInvoker extends TestInvoker {
     }
 }
 
+class AsyncRAFTestInvoker extends BaseRAFTestInvoker {
+    _scheduleCallbacks(resolve) {
+        requestAnimationFrame(async () => {
+            await this._syncCallback();
```
smaug----, line 73: _syncCallback and _asyncCallback should be rename to something else,  since, well, the patch makes _syncCallback rather asynchronous :)

resources/test-invoker.mjs
```diff
@@ -48,8 +67,51 @@ export class RAFTestInvoker extends TestInvoker {
...
+                    resolve();
+                }, 0);
+            };
+
+            setTimeout(() => {
+                gotTimer = true;
+                tryTriggerAsyncCallback();
+            });
+
+            const mc = new MessageChannel();
```
smaug----, line 94: It might be nice to use the same channel for all the tests, just to avoid extra garbage created by the runner. Not sure where there channel could live. Perhaps a static property on the class?

resources/test-invoker.mjs
```diff
@@ -48,8 +67,51 @@ export class RAFTestInvoker extends TestInvoker {
...
+                setTimeout(async () => {
+                    await this._reportCallback();
+                    resolve();
+                }, 0);
+            };
+
+            setTimeout(() => {
+                gotTimer = true;
+                tryTriggerAsyncCallback();
+            });
```
smaug----, line 92: oh, hmm, does this non-double rAF setup have a bug after all.  If the promise callbacks create new promises which get resolve and those callbacks then trigger setTimeout or postMessage, does anything guarantee that test runners setTimeout/messageport gets handled after those?

flashdesignory: We are stopping and waiting for `await this._syncCallback();`. 
The setTimeout / messageport happens after that resolved, correct?
The reason of having the boolean flags is that we can't guarantee in which order setTimeout / postMessage gets called. At least that's how I understood our conversation 🙂

resources/test-invoker.mjs
flashdesignory: LOL, we had the same convo in a similar pr... what do you think about:

```
class TestInvoker {
    constructor(captureTest, captureLayout, reportResults) {
        this._captureTest = captureTest;
        this._captureLayout = captureLayout;
        this._reportResults = reportResults;
    }
} 
```

resources/test-invoker.mjs
flashdesignory: I did test it with the current tests as well as with a test that had an artificial promise resolving after a long period of time in each step.

resources/test-invoker.mjs
flashdesignory: are you thinking maybe something like:
```
class AsyncRAFTestInvoker extends BaseRAFTestInvoker {
    static mc = new MessageChannel();
    _scheduleCallbacks(resolve) {
        requestAnimationFrame(async () => {
            await this._syncCallback();

            let gotTimer = false;
            let gotMessage = false;

            const tryTriggerAsyncCallback = () => {
                if (!gotTimer || !gotMessage)
                    return;

                this._asyncCallback();
                setTimeout(async () => {
                    await this._reportCallback();
                    resolve();
                }, 0);
            };

            const handleMessage = () => {
                AsyncRAFTestInvoker.mc.port1.removeEventListener("message", handleMessage);
                gotMessage = true;
                tryTriggerAsyncCallback();
            };

            setTimeout(() => {
                gotTimer = true;
                tryTriggerAsyncCallback();
            });

            AsyncRAFTestInvoker.mc.port1.addEventListener("message", handleMessage);
            AsyncRAFTestInvoker.mc.port1.start();
            AsyncRAFTestInvoker.mc.port2.postMessage("speedometer");
        });
    }
}
```

resources/test-invoker.mjs
smaug----: But await this._syncCallback(); might return in the same task, if one just does effectively
return new Promise(r => r());

In my example using MessageChannel there were still two rAFs.

I thought we were thinking to use the same runner for async tests and non-async, and I'm worried that not having two rafs would broke non-async tests (meaning that some time wouldn't be measured correctly).  But I could be missing something (I think I did when I said on Slack couple of days ago that this approach is ok :)).

flashdesignory: @smaug---- following up on the above snippet. Do you think this would satisfy async steps with two rafs?

resources/test-invoker.mjs
flashdesignory: @smaug---- and following up here as well in regards to a static MessageChannel

resources/tests.mjs
```diff
@@ -125,6 +125,7 @@ Suites.push({
     name: "TodoMVC-JavaScript-ES5",
     url: "resources/todomvc/vanilla-examples/javascript-es5/dist/index.html",
     tags: ["todomvc"],
+    type: "async",
```
smaug----, line 128: So in practice all the tests are now async?

resources/tests.mjs
```diff
@@ -857,11 +869,12 @@ Suites.push({
     name: "NewsSite-Next",
     url: "resources/newssite/news-next/dist/index.html#/home",
     tags: ["newssite", "language"],
+    type: "async",
     async prepare(page) {
         await page.waitForElement("#navbar-dropdown-toggle");
     },
     tests: [
-        new BenchmarkTestStep("NavigateToUS", (page) => {
+        new BenchmarkTestStep("NavigateToUS", async (page) => {
```
smaug----, line 877: Why is async callback needed here? The function doesn't seem to do anything async, no?

resources/tests.mjs
flashdesignory: This can be reverted. For testing I marked all default tests async. 
How do we decide which tests should be async? I know React / Next.js had issues with not capturing all work. Those could be candidates to mark async?

resources/tests.mjs
flashdesignory: removed!

resources/tests.mjs
flashdesignory: I'll leave it unchanged (opting every single workload into async steps), until I get more feedback from other folks.

resources/test-invoker.mjs
```diff
@@ -24,7 +24,24 @@ export class TimerTestInvoker extends TestInvoker {
     }
 }
 
-export class RAFTestInvoker extends TestInvoker {
```
rniwa, line 27: I'd prefer removing the TimerTestInvoker entirely since we've moved to RAF-based measurement in 3 and we haven't seen any issue with it.

flashdesignory: removed!

rniwa:
Can we split the change to remove `measurementMethod` into a separate PR? We should also update params.mjs and developer-mode.mjs.

flashdesignory:
> Can we split the change to remove `measurementMethod` into a separate PR? We should also update params.mjs and developer-mode.mjs.

ok, reverted for now. I agree a cleanup pr to remove this would be better.

bgrins:
@smaug---- I pushed this to https://feature-async-runner-message-channel--speedometer-preview.netlify.app/ for anyone who wants to test.

resources/shared/test-runner.mjs
```diff
@@ -13,25 +13,52 @@ export class TestRunner {
...
+
+    get syncStartLabel() {
+        return `${this.#suite.name}.${this.#test.name}-start`;
+    }
+
+    get syncEndLabel() {
+        return `${this.#suite.name}.${this.#test.name}-sync-end`;
+    }
+
+    get asyncStartLabel() {
```
rniwa, line 36: Do we really need to add all these getters now that we're only overriding _runSyncStep?

resources/shared/test-invoker.mjs
```diff
@@ -24,7 +24,24 @@ export class TimerTestInvoker extends TestInvoker {
     }
 }
 
-export class RAFTestInvoker extends TestInvoker {
+class AsyncTimerTestInvoker extends TestInvoker {
```
rniwa, line 27: Can we instead just delete `TimerTestInvoker` as we discussed in a separate PR?

resources/suite-runner.mjs
```diff
@@ -75,7 +75,8 @@ export class SuiteRunner {
             if (this.#client?.willRunTest)
                 await this.#client.willRunTest(this.#suite, test);
 
-            const testRunner = new TestRunner(this.#frame, this.#page, this.#params, this.#suite, test, this._recordTestResults);
+            const testRunnerClass = TEST_RUNNER_LOOKUP[this.#suite.type ?? "default"];
```
rniwa, line 78: Instead of changing the test runner based on the suite type for some suites, can we initially add a toggle to developer menu so that we can try out async runner?

camillobruni: To my understanding the agreement in the meeting was to unconditionally enable the new runner for the preact workloads and add the global developer-mode menu in a separate PR.

resources/shared/test-runner.mjs
flashdesignory: I can revert this.

resources/suite-runner.mjs
flashdesignory: correct - below is the list on enabled suites (for now):

-  TodoMVC-React-Complex-DOM,
-  TodoMVC-React-Redux,
-  TodoMVC-Preact-Complex-DOM, (technically not react based, but it's similar)
-  NewsSite-Next,
-  React-Stockcharts-SVG

resources/shared/test-runner.mjs
flashdesignory: done

resources/suite-runner.mjs
rniwa: If we're doing that, we need before/after numbers across browsers.

flashdesignory: I'll run crossbench for the ones on the list and post results here.

flashdesignory: Here are metrics:
Screenshot 2024-12-18 at 5 54 40 PM

smaug---- (approved):
Looks reasonable

resources/shared/test-invoker.mjs
rniwa: We probably want to land the removal PR first.

resources/shared/test-invoker.mjs
```diff
@@ -48,8 +67,58 @@ export class RAFTestInvoker extends TestInvoker {
     }
 }
 
+class AsyncRAFTestInvoker extends BaseRAFTestInvoker {
+    static mc = new MessageChannel();
```
rniwa, line 71: Why does this need to be a static variable?

resources/shared/test-invoker.mjs
flashdesignory: thanks for opening https://github.com/WebKit/Speedometer/pull/466

resources/shared/test-invoker.mjs
rniwa: Oh I see. Maybe but can't we use an instance variable in that case? We don't re-create TestInvoker for each suite / step, right?

flashdesignory: each test creates a new instance:

https://github.com/WebKit/Speedometer/blob/main/resources/shared/test-runner.mjs#L74-L75

flashdesignory:
@rniwa - I believe we addressed all the comments. Anything else missing that I can help with?

camillobruni (approved):

resources/suite-runner.mjs
rniwa: So Safari's total time goes up while the total time for Chrome & Firefox both go down? With that kind of magnitude of a difference, we need a more thorough investigation before we can approve of the change. I suggest we land this method initially as an optional developer-menu only feature instead.

resources/shared/test-invoker.mjs
```diff
@@ -31,7 +33,56 @@ export class RAFTestInvoker extends TestInvoker {
...
+
+            this._asyncCallback();
+            setTimeout(async () => {
+                await this._reportCallback();
+                resolve();
+            }, 0);
+        };
+
+        requestAnimationFrame(async () => {
+            await this._syncCallback();
```
rniwa, line 55: It is a bit suspicious to alway trigger the workload in rAF in this asynchronous measurement method.
A real web apps don't necessarily always wait for rAF to trigger any work so for the maximal realism, it might be better if we just triggered sync callback at a pseudo-random time relative to rAF.

resources/shared/test-invoker.mjs
```diff
@@ -31,7 +33,56 @@ export class RAFTestInvoker extends TestInvoker {
...
+            gotPromise = true;
+            tryTriggerAsyncCallback();
+        });
+
+        requestAnimationFrame(() => {
+            setTimeout(() => {
+                gotTimer = true;
```
rniwa, line 62: So I think we need to wait for any promises scheduled in setTimeout get resolved so I think we need an await here.

resources/shared/test-invoker.mjs
```diff
@@ -31,7 +33,56 @@ export class RAFTestInvoker extends TestInvoker {
...
+                tryTriggerAsyncCallback();
+            });
+
+            AsyncRAFTestInvoker.mc.port1.addEventListener(
+                "message",
+                function () {
+                    gotMessage = true;
+                    tryTriggerAsyncCallback();
```
rniwa, line 70: Ditto here. I think we need to wait for any promises that got scheduled as a result of postMessage to get processed before we finish the measurement.

resources/shared/test-invoker.mjs
rniwa: What I mean is to do this:
```
await Promise.resolve();
```
This will ensure that any promises scheduled during `setTimeout` will get resolved before we set `gotTimer` to `true`.

flashdesignory: @rniwa - added

resources/shared/test-invoker.mjs
flashdesignory: added

resources/shared/params.mjs
```diff
@@ -50,12 +52,14 @@ export class Params {
...
         this.developerMode = this._parseBooleanParam(searchParams, "developerMode");
         this.useWarmupSuite = this._parseBooleanParam(searchParams, "useWarmupSuite");
+        this.useAsyncSteps = this._parseBooleanParam(searchParams, "useAsyncSteps");
         this.waitBeforeSync = this._parseIntParam(searchParams, "waitBeforeSync", 0);
         this.warmupBeforeSync = this._parseIntParam(searchParams, "warmupBeforeSync", 0);
         this.measurementMethod = this._parseMeasurementMethod(searchParams);
         this.shuffleSeed = this._parseShuffleSeed(searchParams);
 
         const unused = Array.from(searchParams.keys());
+        console.log("unused", unused);
```
rniwa, line 62: This is logging intentional??

resources/tests.mjs
```diff
@@ -343,6 +343,7 @@ Suites.push({
     name: "TodoMVC-React-Complex-DOM",
     url: "resources/todomvc/architecture-examples/react-complex/dist/index.html#/home",
     tags: ["todomvc", "complex", "complex-default"],
+    type: "async",
```
rniwa, line 346: Wouldn't this mean all these workloads will use async runner now?

resources/shared/params.mjs
flashdesignory: removed

resources/shared/test-invoker.mjs
```diff
@@ -31,7 +33,58 @@ export class RAFTestInvoker extends TestInvoker {
...
+        });
+    }
+}
+
 export const TEST_INVOKER_LOOKUP = {
     __proto__: null,
     raf: RAFTestInvoker,
 };
+
+export const ASYNC_TEST_INVOKER_LOOKUP = {
```
rniwa, line 87: Do we really need a separate lookup table? Now that timer-based measurement is gone,
why don't we just have a single lookup table with "raf" and "async" as measurement methods.

resources/shared/test-runner.mjs
```diff
@@ -41,7 +52,7 @@ export class TestRunner {
             }
             performance.mark(syncStartLabel);
             const syncStartTime = performance.now();
-            this.#test.run(this.#page);
+            await this._runSyncStep(this.test, this.page);
```
rniwa, line 55: This seems to have a side effect of always pushing the measurement end until when a promise resolves.
I don't think we want to do that for rAF based measurement method (since that can affect the score)
so maybe we should check the return value of this function and await conditionally?

rniwa: More concretely, the following code logs: 1, 2, 3:
```
(async function() { console.log(1); await (() => { })(); console.log(3); })(); console.log(2);
```
whereas the following code logs 1, 3, 2:
```
(async function() { console.log(1); (() => { })(); console.log(3); })(); console.log(2);
```

i.e. `await` will always delay the execution of the continuation until the end of the current micro-task. While that's highly desirable for async measurement, I don't think we want to affect rAF measurement.

resources/shared/test-invoker.mjs
rniwa: ok, we can keep the double rAF for now but I think it's worth exploring if we can execute at rAF un-aligned time.

flashdesignory:
cleaned up some stuff from the developer menu integration. Previously i only checked if the suite.type is "async", which always returned false, since we removed the async type from all tests. 

Now I am checking both: suite.type === "async" or "params.useAsyncSteps"

rniwa (approved):
Great. Thank you for addressing all the review comments!

smaug---- (approved):

PR #459, opened 2024-11-21
add a minimal example of a remote workload using speedometer as a package

bgrins:
This is copying the minimal set of changes from #456 necessary to wire up a trivial workload (followup from discussion in this thread: https://github.com/WebKit/Speedometer/pull/456/files#r1852664786)

The speedometer package is installed via

```
    "dependencies": {
        "speedometer": "../../"
    },
```

and then the workload consumes a single file from it `import { ... } from "speedometer/resources/workload-testing-utils.mjs"`. Then when you `npm run build` it's able to do manage all of the necessary imports within speedometer itself to share code without needing to copy/paste into the files from the runner.

We may want to first land something like this, and then follow up with a smaller change for the Next integration, if there are additional complications with that.


bgrins:
This can be previewed at https://remote-helloworld--speedometer-preview.netlify.app/?developerMode=&suites=Remote-Helloworld

bgrins:
There will need to be some kind of unification with the helper / Page / PageElement methods so the convenience functionality like dispatchMouseEvent etc is made available to remote tests.

The simplest thing might be to export the entire classes from a shared module and make them available to both contexts, though I think there's a desire to get rid of them (instead relying on helper functions that take in the actual element). 

I don't have a strong opinion on how this is resolved, but I think it could be done in a followup.

bgrins:
I'll close this one since #456 has (mostly) incorporated the approach

flashdesignory:
thanks - this was super helpful though!

Issue #460, opened 2024-12-02
Selenium tests time out frequently


flashdesignory:
Our Selenium tests randomly time out. 
We should increase the time Selenium waits before throwing a timeout error.

<img width="1054" alt="Screenshot 2024-12-02 at 9 39 18 AM" src="https://github.com/user-attachments/assets/b7550536-0464-44b5-9d99-1a0aa4ca0c48">

PR #461, opened 2024-12-02
Fix/selenium

flashdesignory:
Testing a potential fix for https://github.com/WebKit/Speedometer/issues/460

### DIFF
--- a/tests/index.html
+++ b/tests/index.html
@@ -64,11 +64,12 @@
 
                 runner.on("end", function () {
                     window.suite = createReport(runner.suite);
-                    const event = new Event("test-complete");
-                    window.dispatchEvent(event);
+                    window.dispatchEvent(new Event("test-complete"));
                 });
             }
+
             window.addEventListener("start-test", startTest);
+            window.dispatchEvent(new Event("test-ready"));
         </script>
     </body>
 </html>
--- a/tests/index.html
+++ b/tests/index.html
@@ -68,7 +68,7 @@
                 });
             }
 
-            window.addEventListener("start-test", () => startTest());
+            window.addEventListener("start-test", () => startTest(), { once: true });
             window.benchmarkReady = true;
             window.dispatchEvent(new Event("benchmark-ready"));
         </script>
--- a/tests/index.html
+++ b/tests/index.html
@@ -69,14 +69,16 @@
             }
 
             function checkIn(type) {
-                if (type === "runner") window.runnerReady = true;
-                else window.domReady = true;
-
                 if (window.runnerReady && window.domReady) startTest();
             }
 
             window.addEventListener("start-test", () => checkIn("runner"));
-            document.addEventListener("readystatechange", () => checkIn("dom"));
+            document.addEventListener("readystatechange", () => {
+                if(document.readyState === "complete") {
+                    window.domReady = true;
+                    checkIn("dom")
+                }
+            });
         </script>
     </body>
 </html>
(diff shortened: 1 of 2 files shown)


camillobruni:
I'll have a look tomorrow. 
I did something similar with JetStream, there is sometimes a race-condition when the selenium code is execute vs the ready-events were sent.

See https://github.com/WebKit/JetStream/pull/11

tests/index.html
```diff
@@ -64,11 +64,21 @@
...
                 runner.on("end", function () {
                     window.suite = createReport(runner.suite);
-                    const event = new Event("test-complete");
-                    window.dispatchEvent(event);
+                    window.dispatchEvent(new Event("test-complete"));
                 });
             }
-            window.addEventListener("start-test", startTest);
+
+            function checkIn(type) {
```
julienw, line 71: what's the type parameter for?

tests/run.mjs
```diff
@@ -108,8 +107,8 @@ async function test() {
                     }),
                 { once: true }
             );
-            const event = new Event("start-test");
-            window.dispatchEvent(event);
+            window.runnerReady = true;
```
julienw, line 110: I feel like it would be more self-contained to set this property inside event listener for "start-test", in index.html

tests/index.html
julienw: Please add a comment to future readers explaining what this function is for

tests/run.mjs
flashdesignory: This is a scenario, where we need global variables. 
The runner shares the same window object, which is the only scope I can add a variable in the runner, that the html page "can read". If the custom event fires, before the listener was added, the html page can read its window flags and decide if it can run the tests.

julienw: I meant that you could set the global in the `start-test` event listener above:

```js
window.addEventListener("start-test", () => {
  window.runnerReady = true;
  checkIn();
});
```

Then it's more self-contained because the globals are updated only in one file.

flashdesignory: What happens if the runner dispatched "start-test", before the html page can add its listener for the event?

julienw: Ah I see what you mean. I'm not very familiar with Selenium but my understanding is that `await driver.get(...)` should be resolved only when the page is fully loaded (by default). See in https://www.selenium.dev/documentation/webdriver/drivers/options/#pageloadstrategy.

Maybe simply removing `type="module"` on the script (ensuring that it would run synchronously) would fix the issue you see, without any other change?


Otherwise I'm thinking of another way of solving this without using any variable, by adding this here:
```js
await driver.executeAsyncScript(function (callback) {
  if (document.readyState === "complete") {
    callback();
    return;
  }
  document.addEventListener("readystatechange", function onReadyStateChange() {
    if (document.readyState === "complete") {
      document.removeEventListener("readystatechange", onReadyStateChange);
      callback();
    }
  });
});
```

flashdesignory: I tried both and neither worked consistently

camillobruni (approved):

tests/run.mjs
flashdesignory: @julienw - I refactored it a bit and using now the same solution @camillobruni did for JetStream. 
This seems to work consistently for me in CI and is a bit cleaner than before.

rniwa (approved):
Looks sane to me.

julienw (approved):
Thanks, I like this version much more than the previous one :-)

I just left 2 optional comments for correctness, but otherwise this looks good to me

PR #464, opened 2024-12-17
Avoid measurement dead-spot between sync and async phase

camillobruni:
We could potentially miss a `gc` between the `performance.mark` calls for `syncEndTime` and `asyncStartTime`.

This PR changes the code to reuse the sync end times and marker for the async start.
This means we'll now also measure one additional `performance.mark` call.

Summary 5 runs:
```
browser                                     Google Chrome          Firefox                Safari
version                                     131.0.6778.140         133.0                  18.1.1 (20619.2.8.11.12)
Score Before                                32.07 ± 0.67%          27.228 ± 0.13%         32.75 ± 0.48%
Score  After                                31.14 ± 0.54%          26.10 ± 0.79%          30.36 ± 0.68%
```

I have to re-run the detailed measurements tonight when my machine is not in use to reduce noise.

### DIFF
--- a/resources/test-runner.mjs
+++ b/resources/test-runner.mjs
@@ -24,7 +24,7 @@ export class TestRunner {
         const testName = this.#test.name;
         const syncStartLabel = `${suiteName}.${testName}-start`;
         const syncEndLabel = `${suiteName}.${testName}-sync-end`;
-        const asyncStartLabel = `${suiteName}.${testName}-async-start`;
+        const asyncStartLabel = syncEndLabel;
         const asyncEndLabel = `${suiteName}.${testName}-async-end`;
 
         let syncTime;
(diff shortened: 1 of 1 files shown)


rniwa (approved):
Seems sensible to me.

resources/test-runner.mjs
```diff
@@ -24,7 +24,7 @@ export class TestRunner {
         const testName = this.#test.name;
         const syncStartLabel = `${suiteName}.${testName}-start`;
         const syncEndLabel = `${suiteName}.${testName}-sync-end`;
-        const asyncStartLabel = `${suiteName}.${testName}-async-start`;
+        const asyncStartLabel = syncEndLabel;
```
julienw, line 27: I'd rather remove the name and use `syncEndLabel` directly below. I think this makes it more obvious what the PerformanceTiming marker will be about.

julienw (approved):
This looks reasonable to me as well.

resources/test-runner.mjs
camillobruni: done.

camillobruni:
Just rerun the numbers and there is no measurable change.
```
browser       Google Chrome          Firefox                Safari
version       131.0.6778.205         133.0.3                18.1.1 (20619.2.8.11.12)
Score Main    32.60 ± 0.58%          28.66 ± 0.39%          33.59 ± 0.59%
Score PR      32.54 ± 0.52%          28.63 ± 0.61%          33.51 ± 0.99%
```

PR #466, opened 2024-12-20
Delete the timer based measurement since we've switched over to

rniwa:
requestAnimationFrame based measurement for 3.0 and found no issues.

### DIFF
--- a/resources/shared/params.mjs
+++ b/resources/shared/params.mjs
@@ -15,7 +15,7 @@ export class Params {
     // Change how a test measurement is triggered and async time is measured:
     // "timer": The classic (as in Speedometer 2.x) way using setTimeout
     // "raf":   Using rAF callbacks, both for triggering the sync part and for measuring async time.
-    measurementMethod = "raf"; // or "timer"
+    measurementMethod = "raf";
     // Wait time before the sync step in ms.
     waitBeforeSync = 0;
     // Warmup time before the sync step in ms.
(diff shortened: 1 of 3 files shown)


resources/params.mjs
```diff
@@ -122,8 +122,8 @@ class Params {
         if (!searchParams.has("measurementMethod"))
             return defaultParams.measurementMethod;
         const measurementMethod = searchParams.get("measurementMethod");
-        if (measurementMethod !== "timer" && measurementMethod !== "raf")
-            throw new Error(`Invalid measurement method: '${measurementMethod}', must be either 'raf' or 'timer'.`);
+        if (measurementMethod !== "raf")
```
rniwa, line 125: I kept this here since we may want to support "async" as an option now.

resources/shared/params.mjs
```diff
@@ -122,8 +122,8 @@ export class Params {
         if (!searchParams.has("measurementMethod"))
             return defaultParams.measurementMethod;
         const measurementMethod = searchParams.get("measurementMethod");
-        if (measurementMethod !== "timer" && measurementMethod !== "raf")
-            throw new Error(`Invalid measurement method: '${measurementMethod}', must be either 'raf' or 'timer'.`);
+        if (measurementMethod !== "raf")
+            throw new Error(`Invalid measurement method: '${measurementMethod}', must be either 'raf'.`);
```
flashdesignory, line 126: should we change it to:
```
throw new Error(`Invalid measurement method: '${measurementMethod}', must be 'raf'.`);
```

flashdesignory (approved):

resources/shared/params.mjs
smaug----: yeah, drop "either"

smaug---- (approved):

julienw (approved):

camillobruni:
- Removing the timer-based code sounds good
- Can we keep the UI so we can switch between the current rAF and the new async runner?

rniwa:
> * Can we keep the UI so we can switch between the current rAF and the new async runner?

I don't know how keeping the UI mean here because it's a checkbox with "rAF timing" label. Are you thinking of replacing that with "Async timing" or something?

PR #467, opened 2024-12-23
Add Measurement and Scoring details to about  page

camillobruni:
Given that this comes up ever so often in discussion with internal and external folks it might be worthwhile documenting in a bit more detail which parts are measured and how we calculate the final score.

- Add example timeline to explain which parts contribute to the score
- Add basic formulas on how the test steps are combined into a final score


bgrins:
Instead of the about page, could we put this type of information in a markdown file in the repo? That way it wouldn't require publishes to browserbench.org (presumably requiring a version bump) to update the content.

camillobruni:
I can put it into a separate html file, markdown is likely the wrong file format for math formulas and tables, no?

bgrins:
I think you can embed the HTML into markdown to preserve the ability to host the preview directly out of the github repo but no strong opinion either way.

rniwa (approved):

camillobruni:
I've not pulled out the timeline table and the scoring math into separate SVGs (this seems to be the least brittle way to get externally formatted content into markdown).

WDYT?

resources/measurement-timeline.svg
rniwa, line 1: We can either fill the entire image with a background color or omit the background color?

resources/score.svg
rniwa, line 1: Ditto. We can either fill the entire image with a background color or omit the background color?

bgrins (approved):
Thanks. We can iterate on it in the future as necessary

PR #468, opened 2024-12-23
Auto run unittests in tests/index.html

camillobruni:
Instead of triggering the tests from the outside, we can always directly start the mocha runner in `tests/index.html`.
The selenium driver now syncs on either `globalThis.testResults` being there or the `test-complete` event.

This way we should be able to reliably run the tests both via selenium and the browser.

### DIFF
--- a/tests/index.html
+++ b/tests/index.html
@@ -27,7 +27,7 @@
                 },
             });
 
-            await import("./benchmark-runner-tests.mjs")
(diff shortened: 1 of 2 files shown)


flashdesignory (approved):
LGTM

rniwa (approved):
seems ok.

PR #470, opened 2025-01-07
Add end2end tests

camillobruni:
- Move unitests to run-unittests.mjs
- Move common test driver and server setup to separate helper.mjs module
- Add `SpeedometerReady` and `SpeedometerDone` events for easier testing
- Add simple end2end tests


resources/shared/benchmark.mjs
```diff
@@ -55,7 +55,7 @@ export class BenchmarkSuite {
             const result = await test.runAndRecord(params, this, test, this.record);
             measuredValues.tests[test.name] = result;
             measuredValues.total += result.total;
-            onProgress?.(test.name);
+            measuredValues.totalPrepare += result.onProgress?.(test.name);
```
flashdesignory, line 58: do we need to add an initial value of 0?

```
const measuredValues = {
            tests: {},
            total: 0,
            totalPrepare: 0,
        };
```

tests/run-end2end.mjs
```diff
@@ -0,0 +1,89 @@
...
+    assert(metric.mean >= 0);
+}
+
+async function testIterations() {
+    const metrics = await testPage("index.html?iterationCount=3");
+    Suites.forEach((suite) => {
+        if (!suite.disabled) {
+            const metric = metrics[suite.name];
+            assert(metric.values.length === 3);
+        }
```
julienw, line 44: I think we need to test that if the suite is disabled it doesn't appear, right?

tests/run-end2end.mjs
```diff
@@ -0,0 +1,89 @@
...
+    await driver.get(`http://localhost:${PORT}/${url}`);
+    await driver.executeAsyncScript((callback) => {
+        if (globalThis.benchmarkClient)
+            callback();
+        else
+            window.addEventListener("SpeedometerReady", () => callback(), { once: true });
+    });
+    console.log("    - Awaiting Benchmark");
+    const metrics = await driver.executeAsyncScript((callback) => {
+        window.addEventListener("SpeedometerDone", () => callback(globalThis.benchmarkClient.metrics));
```
julienw, line 19: nit: I believe this needs `{ once: true }` as well

tests/run-unittests.mjs
```diff
@@ -0,0 +1,57 @@
+#! /usr/bin/env node
+/* eslint-disable-next-line  no-unused-vars */
```
julienw, line 2: nit: it could be good to add a comment explaining a few things about the tests, such as "tests are run in the browser by the `index.html` file. Individual tests are defined in benchmark-runner-tests.html", things like that.

julienw (approved):
thanks, this looks good to me
I only have a few nits before landing :) see comments below
thanks for the new tests!

tests/run-end2end.mjs
```diff
@@ -0,0 +1,89 @@
...
+    assert(metrics.Score.mean > 0);
+}
+
+function validateMetric(name, metric) {
+    assert(metric.name === name);
+    assert(metric.mean >= 0);
+}
+
+async function testIterations() {
+    const metrics = await testPage("index.html?iterationCount=3");
```
julienw, line 39: maybe 2 is good enough and would take less time?

camillobruni:
Updated a few more things
- updated required selenium version to 4.27 to fix local runs
- installed better error handlers to catch runner errors in the end2end tests
- moved `benchmark-runner-tests.mjs` to `tests/unittests`

julienw:
Thanks, I like the fact we call a function better (didn't want to highlight it before :-) )

BTW I wonder if there's a way to run the tests headless too. How does it work on CI, is there a X server or something?

camillobruni:
headless would mostly depend on how the browser is set up.
I know how to run chrome headless e.g. it basically only needs the `--headless` command line flag, I think firefox has the same flag, no? (We can add that by default or make it customizable in an additional PR?)

Given that we currently only run on macOs it wont matter for the official testing (there is no real headless on macOS I think).

PR #473, opened 2025-01-22
Use local-web-server package for all servers

camillobruni:
- Merge the test, dev and npm run server commands to use local-web-server.
- Send COEP headers by default
- Run `npm ci` to regenerate package-lock and sort package.json entries 
- Resolve committed merge conflict in bundled news-site workload


flashdesignory (approved):

resources/newssite/news-next/dist/_next/static/IbZpY6lpxNCTHGQFLD2XQ/_buildManifest.js
julienw, line 1: I'm a bit surprised to see that here, I believe this was fixed by thorsten in another PR

tests/server.mjs
```diff
@@ -1,52 +1,38 @@
...
-        const found = !pathTraversal && exists;
-        const streamPath = found ? filePath : `${STATIC_PATH}/index.html`;
-        const ext = path.extname(streamPath).substring(1).toLowerCase();
-        const stream = fs.createReadStream(streamPath);
-        return { found, ext, stream };
+    const ws = await LocalWebServer.create({
+        port: port,
+        directory: ROOT_DIR,
+        corsOpenerPolicy: "same-origin",
+        corsEmbedderPolicy: "require-corp",
```
julienw, line 18: nit: add the `logFormat: "dev"` to get some diagnostics in the console.
(I think this is how this would be specified, the documentation is quite sparse about the API)

tests/server.mjs
```diff
@@ -1,52 +1,38 @@
...
+    });
+    console.log(`Server started on http://localhost:${port}`);
```
julienw, line 20: I noticed the log is output even though the server can't be started (for example run it twice in 2 terminals: the second one will return right away). I'm a bit surprised that `create` doesn't throw in that case...

julienw: Maybe something such as this would work:
```js
ws.server.on('listening', () => console.log(`Server started on http://localhost:${port}`));
ws.server.on('error', (e) => console.error(`Error while starting the server`, e);
``

Comment (julienw) [reply tests/server.mjs:20]: And even our returned promise could wait for these events, and be either fulfilled or rejected depending on them, that would work better with the tests I believe.

File: package.json [hunk 3]
```diff
@@ -15,8 +15,8 @@
     },
     "license": "SEE LICENSE IN LICENSE",
     "scripts": {
-        "dev": "http-server ./ -c-1 --cors",
-        "server": "http-server ./ --cors",
+        "dev": "node tests/server.mjs",
+        "server": "node tests/server.mjs",
```
julienw, line 19: do we need to keep both?

package.json
```diff
@@ -35,22 +35,23 @@
...
         "eslint-plugin-vue": "^9.10.0",
         "expect.js": "^0.3.1",
-        "http-server": "^14.1.1",
         "mocha": "^10.2.0",
         "prettier": "^2.8.3",
         "selenium-webdriver": "^4.27.0",
         "sinon": "^17.0.1",
-        "typescript": "^5.0.4"
+        "typescript": "^5.0.4",
+        "local-web-server": "^5.4.0"
```
julienw, line 55: I wonder if we should use `lws` with just the few middlewares we need (probably just log, cors, static, index?), instead of `local-web-server` with all the dependencies it comes with. What do you think?

julienw (commented):
This looks pretty good to me!
Here are some suggestions, tell me what you think!

package.json
julienw: (I would keep just "dev")

julienw:
Looks like the CI failed too

package.json
camillobruni: ah yeah, wanted to ask... I'd just keep `server` as an alias. but no strong opinion here.

tests/server.mjs
camillobruni: I think you can use any cli-option and make it came-case here. let me try.

package.json
camillobruni: was easy enough, and limited it now to "lws", "lws-log", "lws-cors", "lws-static" and "lws-index"

rniwa (approved):

julienw:
image
Also it's worth updating to latest version of koa (probably you can checkout the package-lock.json from the main branch, and run `npm i` again).

package.json
```diff
@@ -35,22 +34,27 @@
...
         "prettier": "^2.8.3",
         "selenium-webdriver": "^4.27.0",
         "sinon": "^17.0.1",
-        "typescript": "^5.0.4"
+        "typescript": "^5.0.4",
+        "lws": "^4.2.0",
+        "lws-cors": "^4.2.1",
+        "lws-index": "^3.1.1",
+        "lws-log": "^3.0.0",
+        "lws-static": "^3.1.1"
```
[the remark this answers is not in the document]
camillobruni: ah I forgot that I tihnk.

camillobruni:
I never know what the command for this is `npm i` didn't do the trick, so I now ran `npm audit fix` and updated a bunch of other packages too.

rniwa (approved):

PR #478, opened 2025-01-29
Update github workloads

camillobruni:
- Use latest action versions
- Bump node version to current stable v22
- Use caches for node modules to slightly speed up setup
- Mild formatting was applied


rniwa (approved):
Looks okay to me.

.github/workflows/test.yml
```diff
@@ -15,15 +15,27 @@ jobs:
...
+                restore-keys: ${{ runner.os }}-build-node-modules-
+
             - name: Setup Node
-              uses: actions/setup-node@v3
+              uses: actions/setup-node@v4
               with:
-                  node-version: 18.13.0
-            - name: Install
-              run: npm install
+                  node-version: '23.x'
```
julienw, line 31: odd versions are the development versions, the stable versions are what they call "active LTS" or "maintenance LTS" on https://github.com/nodejs/release?tab=readme-ov-file#release-schedule

Let's use 22.x :-)

julienw: Also it looks like it can use the value form `package.json` directly (see documentation in https://github.com/actions/setup-node) when using
```
 node-version-file: "package.json"
```

Might be worth trying :-)

.github/workflows/test.yml
```diff
@@ -15,15 +15,27 @@ jobs:
...
+              uses: actions/checkout@v4
+
+            - name: Cache node modules
+              uses: actions/cache@v4
+              with:
+                # npm cache files are stored in `~/.npm` on Linux/macOS
+                path: ~/.npm
+                key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+                restore-keys: ${{ runner.os }}-build-node-modules-
+
```
julienw, line 27: it's not needed, see https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data

Just add `cache: "npm"` in setup-node below

.github/workflows/test.yml
julienw: BTW you also need to update package.json of course :-)

and all comments also apply below

camillobruni:
Simplified the test.yml file based on the feedback.
Looks like the netlify deployment is now unhappy with the upped node version requirement. @julienw is this easy to fix or should I just lower the requirements again?

julienw:
> Simplified the test.yml file based on the feedback. Looks like the netlify deployment is now unhappy with the upped node version requirement. @julienw is this easy to fix or should I just lower the requirements again?

image
just changed it :)

[Another way would be to use a `.node-version` or `.nvmrc` file in the repository (this would also override any setting in the configuration UI BTW).](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript) I think `node-version-file` can also use that one BTW.
It's a bit too bad that netlify can't use the version specified in package.json :(

package.json
```diff
@@ -3,7 +3,7 @@
     "version": "3.0.0-alpha",
     "description": "An open source repository for the Speedometer benchmark.",
     "engines": {
-        "node": ">=18.13.0",
+        "node": ">=22.0.0",
```
julienw, line 6: do we want to ensure that we stay on v22 by adding `< 23` as well? I'm a bit afraid that this will switch to v24 automatically and we'll get build errors. We can always change that later if this happens of course.

julienw: I believe you need to run `npm i` to update `package-lock.json` as well.

julienw (approved):
Thanks, looks good to me!

.github/workflows/test.yml
```diff
@@ -15,15 +15,20 @@ jobs:
...
+              uses: actions/setup-node@v4
               with:
-                  node-version: 18.13.0
-            - name: Install
-              run: npm install
+                  node-version-file: "package.json"
+                  cache: "npm"
+
+            - name: Install Node Packages
+              run: npm ci
```
julienw, line 27: much better!

package.json
camillobruni: ah yeah... of course.
Let me add a CI check for that as well in a separate PR.

julienw: I think `npm ci` does it for dependencies, it's disappointing it doesn't work for everything that can update it!

PR #488, opened 2025-02-17
Perf-Dashboard.Render sync phase sets location.hash, but the actual DOM modifications happen asynchronously in the onhashchange() callback.

mattwoodrow:
This results in nothing interesting being measured during the sync phase, and the resulting rendering update (during the async phase) not including the change.

Change the test to manually and synchronously trigger the onhashchange() callback, and change the page content to detect and omit duplicate events.

Crossbench results before/after the change:

```
label                                       macos.arm64.local_0    macos.arm64.local_1    macos.arm64.local_2
browser                                     Firefox                Safari                 Google Chrome
version                                     135.0                  18.3 (20620.2.4)       133.0.6943.99
os                                          macos 15.3.1 arm64     macos 15.3.1 arm64     macos 15.3.1 arm64
device                                      MacBookPro18,2         MacBookPro18,2         MacBookPro18,2
cpu                                         Apple M1 Max 10 cores  Apple M1 Max 10 cores  Apple M1 Max 10 cores
runs                                        1                      1                      1
failed runs                                 0                      0                      0
Perf-Dashboard                              34.5                   33.9999999999995       29.539999999850988


Perf-Dashboard                              44.2                   40.2                   38.01000000089407
```

### DIFF
--- a/resources/perf.webkit.org/public/v3/pages/page-router.js
+++ b/resources/perf.webkit.org/public/v3/pages/page-router.js
@@ -97,10 +97,11 @@ class PageRouter {
 
     _hashDidChange()
     {
-        if (unescape(location.hash) == this._hash)
+        let unescapedHash = unescape(location.hash);
+        if (unescapedHash == this._hash)
             return;
         this.route();
-        this._hash = null;
+        this._hash = unescapedHash;
     }
 
     _serializeToHash(route, state)
--- a/resources/perf.webkit.org/public/v3/bundled-scripts.js
+++ b/resources/perf.webkit.org/public/v3/bundled-scripts.js
@@ -2608,8 +2608,8 @@ url(routeName,state)
 _updateURLState()
 {this._historyTimer=null;console.assert(this._currentPage);const currentPage=this._currentPage;this._hash=this._serializeToHash(currentPage.routeName(),currentPage.serializeState());location.hash=this._hash;}
 _hashDidChange()
-{if(unescape(location.hash)==this._hash)
-return;this.route();this._hash=null;}
+{let unescapedHash=unescape(location.hash);if(unescapedHash==this._hash)
+return;this.route();this._hash=unescapedHash;}
 _serializeToHash(route,state)
 {const params=[];for(const key in state)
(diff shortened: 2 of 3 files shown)


resources/perf.webkit.org/public/v3/pages/page-router.js
```diff
@@ -97,8 +98,12 @@ class PageRouter {
...
     {
-        if (unescape(location.hash) == this._hash)
+        let unescapedHash = unescape(location.hash);
+        if (unescapedHash == this._hash)
             return;
+        if(unescapedHash == this._previousHash)
+            return;
+        this._previousHash = unescapedHash;
         this.route();
         this._hash = null;
```
julienw, line 108: it's not clear to me why `this._hash` is set to null here. In `_updateURLState` above it looks like we want to keep it synced with location.hash.
Then if we don't reset it, we probably don't need _previousHash and the check above would work, right?

mattwoodrow: That's a good point, much simpler to do that. Fixed :)

julienw (approved):
Thanks, this looks better to me.
I also looked with a firefox profiler, this clearly captures more work, which sounds good to me.

It would be too to have a sign-off from somebody at Apple who initially brought this in speedometer.

There's just one small correctness issue left to fix, I believe.

Thanks for the fix!

camillobruni (approved):

rniwa (approved):

flashdesignory:
do we also want to merge the same changes into the 3.1 release branch?

rniwa:
Yeah, we should probably do that.

rniwa:
Here's a PR to merge the fix into release/3.1: https://github.com/WebKit/Speedometer/pull/496

resources/perf.webkit.org/public/v3/pages/page-router.js
```diff
@@ -97,10 +97,11 @@ class PageRouter {
 
     _hashDidChange()
     {
-        if (unescape(location.hash) == this._hash)
+        let unescapedHash = unescape(location.hash);
```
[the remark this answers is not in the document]
julienw: mmm right
If the hash doesn't have any encoded char this is OK, but I think this will be buggy as soon as you start having some. Probably we don't have any at the moment.

PR #491, opened 2025-02-26
Use WakeLock

camillobruni:
- Add `WakeLock` wrapper class around the [web-api](https://developer.mozilla.org/en-US/docs/Web/API/WakeLock)
- Use `WakeLock.aquire()` and `WakeLock.release()` in the benchmark runner

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -278,26 +278,26 @@ function seededHashRandomNumberGenerator(a) {
 }
 
 class WakeLock {
-    _wakeLockSentinel = undefined;
+    #wakeLockSentinel = undefined;
     async request() {
         if (!navigator.wakeLock)
             return;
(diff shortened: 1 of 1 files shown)


resources/benchmark-runner.mjs
```diff
@@ -277,6 +277,31 @@ function seededHashRandomNumberGenerator(a) {
     };
 }
 
+class WakeLock {
+    _wakeLockSentinel = undefined;
```
rniwa, line 281: Use a private variable?

rniwa (approved):

julienw (approved):
yay, thanks!
I tried it on my phone, this worked perfectly.

resources/benchmark-runner.mjs
camillobruni: Done.

PR #492, opened 2025-02-26
Use "next" as version ID on the main branch

camillobruni:
I accidentally ran the wrong version locally and would like to make it more obvious which branch we're on.
Happy to take suggestions here :).

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,7 +13,7 @@
             <section id="home">
                 <a href="#home" class="logo">
                     <img srcset="resources/logo@2x.png 2x" src="resources/logo.png" alt="Speedometer" />
-                    <div class="version">3.0</div>
+                    <div class="version next">Next</div>
                 </a>
                 <div class="content">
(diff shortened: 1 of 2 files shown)


camillobruni:
Screenshot 2025-02-26 at 17 26 33

rniwa (approved):

index.html
```diff
@@ -13,7 +13,7 @@
             <section id="home">
                 <a href="#home" class="logo">
                     <img srcset="resources/logo@2x.png 2x" src="resources/logo.png" alt="Speedometer" />
-                    <div class="version">3.0</div>
+                    <div class="version next">Next</div>
```
julienw, line 16: (optional) I would make it without a capital letters, all small letters, because the speedometer logo is also all small letters.
(maybe change the `alt` above to reflect that too)

julienw (approved):
thanks, makes sense!

PR #497, opened 2025-03-03
Use "next" as version on main consistently

camillobruni:
Let's use "next" for the main branch version identifier consistently in all sections.

### DIFF
--- a/about.html
+++ b/about.html
@@ -12,7 +12,7 @@
             <section id="about" class="visible" data-title="About">
                 <a href="./" class="logo">
                     <img srcset="resources/logo@2x.png 2x" src="resources/logo.png" alt="Speedometer" />
-                    <div class="version">3.0</div>
+                    <div class="version next">next</div>
                 </a>
                 <div class="section-grid">
                     <h1 class="section-header">About Speedometer 3</h1>
--- a/about.html
+++ b/about.html
@@ -17,7 +17,7 @@
                 <div class="section-grid">
                     <h1 class="section-header">About Speedometer</h1>
                     <div class="section-content">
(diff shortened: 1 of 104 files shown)


about.html
```diff
@@ -12,7 +12,7 @@
             <section id="about" class="visible" data-title="About">
                 <a href="./" class="logo">
                     <img srcset="resources/logo@2x.png 2x" src="resources/logo.png" alt="Speedometer" />
-                    <div class="version">3.0</div>
+                    <div class="version next">next</div>
                 </a>
                 <div class="section-grid">
                     <h1 class="section-header">About Speedometer 3</h1>
```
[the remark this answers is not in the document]
julienw: Also package.json should probably read 4.0.0-alpha

instructions.html
```diff
@@ -13,7 +13,7 @@
             <section id="instructions" class="visible" data-title="Test Instructions">
                 <a href="./" class="logo">
                     <img srcset="resources/logo@2x.png 2x" src="resources/logo.png" alt="Speedometer" />
-                    <div class="version">3.0</div>
+                    <div class="version next">next</div>
                 </a>
                 <div class="section-grid">
                     <h1 class="section-header">Speedometer 3 General Instructions</h1>
```
julienw, line 19: Maybe we can remove the version at all here?

about.html
julienw, line 18: This specific one is still there :-)  But maybe we can remove the version completely for this line, what do you think?

julienw (approved):
Thanks

rniwa (approved):

instructions.html
camillobruni: I've tried now both, kinda prefer having the version consistently displayed in all views on the side.
But I've removed the "3" in the instruction text now.

julienw: yeah my comment was targeting line 19

PR #506, opened 2025-04-14
Fix non-standard suites and tags with developerMode

camillobruni:
- Keep params and Suites selection in sync
- Refactor updateParamsSuitesAndTags into separate method

### DIFF
--- a/resources/developer-mode.mjs
+++ b/resources/developer-mode.mjs
@@ -127,6 +127,7 @@ function createUIForSuites() {
         checkbox.checked = !suite.disabled;
         checkbox.onchange = () => {
             suite.disabled = !checkbox.checked;
+            updateParamsSuitesAndTags();
             updateURL();
         };
         checkboxes.push(checkbox);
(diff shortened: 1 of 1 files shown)


resources/developer-mode.mjs
```diff
@@ -235,34 +236,44 @@ function createUIForRun() {
...
+    let commonTags = new Set(selectedSuites[0].tags);
+    for (const suite of Suites) {
+        if (suite.disabled)
+            suite.tags.forEach((tag) => commonTags.delete(tag));
+        else
+            commonTags = new Set(suite.tags.filter((tag) => commonTags.has(tag)));
+    }
+    if (selectedSuites.length > 1 && commonTags.size) {
+        const tags = [...commonTags];
+        if (tags[0] !== "default")
```
rniwa, line 259: Implicitly relying on "default" being the first element isn't great. Why not just do this instead?
```js
params.tags = commonTags.has("default") ? [...commonTags] : [];
```

rniwa (approved):

resources/developer-mode.mjs
camillobruni: ah yes, much nicer that way!

PR #510, opened 2025-04-30
Customize Force Layout

camillobruni:
Currently we have two different modes in which we force layout

- `getBoundingClientRect` used by default in the benchmark runner and tests
- `elementFromPoint` used in the remote workloads

This PR adds a new `layoutMode` param to switch between the two modes.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -23,6 +23,7 @@ function getParent(lookupStartNode, path) {
 class Page {
     constructor(frame) {
         this._frame = frame;
+        this._leakedLayout = {};
     }
 
     getLocalStorage() {
@@ -30,8 +31,13 @@ class Page {
     }
 
     layout() {
-        const body = this._frame.contentDocument.body.getBoundingClientRect();
-        this.layout.e = document.elementFromPoint((body.width / 2) | 0, (body.height / 2) | 0);
+        const body = this._frame ? this._frame.contentDocument.body : document.body;
+        const bodyRect = body.getBoundingClientRect();
+        if (params.layoutMode === "getBoundingClientRect")
+            this._leakedLayout.sum = bodyRect.width;
+        if (params.layoutMode === "elementFromPoint")
+            this._leakedLayout.element = document.elementFromPoint((bodyRect.width / 2) | 0, (bodyRect.height / 2) | 0);
+        return bodyRect;
     }
 
     async waitForElement(selector) {
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -1,5 +1,6 @@
 import { Metric } from "./metric.mjs";
 import { params } from "./shared/params.mjs";
+import { forceLayout } from "./shared/helpers.mjs";
 import { SUITE_RUNNER_LOOKUP } from "./suite-runner.mjs";
 
 const performance = globalThis.performance;
@@ -23,7 +24,6 @@ function getParent(lookupStartNode, path) {
 class Page {
     constructor(frame) {
         this._frame = frame;
-        this._leakedLayout = {};
     }
 
     getLocalStorage() {
@@ -32,12 +32,8 @@ class Page {
 
     layout() {
         const body = this._frame ? this._frame.contentDocument.body : document.body;
-        const bodyRect = body.getBoundingClientRect();
-        if (params.layoutMode === "getBoundingClientRect")
-            this._leakedLayout.sum = bodyRect.width;
-        if (params.layoutMode === "elementFromPoint")
-            this._leakedLayout.element = document.elementFromPoint((bodyRect.width / 2) | 0, (bodyRect.height / 2) | 0);
-        return bodyRect;
+        const value = forceLayout(body, params.layoutMode);
+        body._leakedLayoutValue = value; // Prevent dead code elimination.
     }
 
     async waitForElement(selector) {
--- a/resources/shared/helpers.mjs
+++ b/resources/shared/helpers.mjs
@@ -24,8 +24,7 @@ export function getAllElements(selector, path = [], lookupStartNode = document) 
 }
 
 export function forceLayout(body, layoutMode = "elementFromPoint") {
-    if (body !== undefined)
-        body = document.body;
+    body ??= document.body;
     const rect = body.getBoundingClientRect();
     if (layoutMode === "elementFromPoint")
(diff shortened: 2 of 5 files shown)


resources/shared/params.mjs
```diff
@@ -27,6 +29,8 @@ export class Params {
     // "generate": generate a random seed
     // <integer>: use the provided integer as a seed
     shuffleSeed = "off";
+    // Choices: ""
```
julienw, line 32: I feel like you wanted to add a better comment ;-)

resources/shared/params.mjs
```diff
@@ -1,3 +1,5 @@
+export const LAYOUT_MODE = Object.freeze(["getBoundingClientRect", "elementFromPoint"]);
```
julienw, line 1: optional nit
```suggestion
export const LAYOUT_MODES = Object.freeze(["getBoundingClientRect", "elementFromPoint"]);
```

resources/shared/params.mjs
```diff
@@ -121,14 +126,14 @@ export class Params {
...
 
-    _parseMeasurementMethod(searchParams) {
-        if (!searchParams.has("measurementMethod"))
-            return defaultParams.measurementMethod;
-        const measurementMethod = searchParams.get("measurementMethod");
-        if (measurementMethod !== "raf")
-            throw new Error(`Invalid measurement method: '${measurementMethod}', must be 'raf'.`);
-        searchParams.delete("measurementMethod");
-        return measurementMethod;
+    _parseChoiceParam(searchParams, paramKey, choices) {
```
julienw, line 129: optional nit: I'd call the function `_parseEnumParam`:
```suggestion
    _parseEnumParam(searchParams, paramKey, enumArray) {
```

resources/benchmark-runner.mjs
```diff
@@ -23,15 +23,21 @@ function getParent(lookupStartNode, path) {
...
         return this._frame.contentWindow.localStorage;
     }
 
     layout() {
-        const body = this._frame.contentDocument.body.getBoundingClientRect();
-        this.layout.e = document.elementFromPoint((body.width / 2) | 0, (body.height / 2) | 0);
+        const body = this._frame ? this._frame.contentDocument.body : document.body;
+        const bodyRect = body.getBoundingClientRect();
+        if (params.layoutMode === "getBoundingClientRect")
+            this._leakedLayout.sum = bodyRect.width;
```
julienw, line 37: I don't know if that makes a difference, but the previous code was using `.height`.

resources/shared/helpers.mjs
```diff
@@ -24,6 +24,7 @@ export function getAllElements(selector, path = [], lookupStartNode = document)
 }
 
 export function forceLayout() {
+    // FIXME: sync implementation with Page.prototype.layout().
```
julienw, line 27: Why not doing the sync now?

julienw (changes requested):
thank you, this looks good to me

Just a couple of suggestions, tell me what you think

resources/shared/helpers.mjs
julienw: ah I see, it's used in the remote page.
Do you have a plan about this?

resources/benchmark-runner.mjs
camillobruni: ups, yeah let's stick to the old code.

resources/shared/helpers.mjs
camillobruni: I would like to generally get rid of the Page abstraction and just have the helpers there (but probably it's easier to move the Page to the shared module).

camillobruni: ah let me just move all code to this helper for now... then at least it's uniform :)

julienw (commented):
I think the new code doesn't have the same perf characteristics as the old one, speedometer seems to run faster. Do you see that as well?

resources/shared/helpers.mjs
```diff
@@ -23,9 +23,11 @@ export function getAllElements(selector, path = [], lookupStartNode = document)
     return elements;
 }
 
-export function forceLayout() {
-    // FIXME: sync implementation with Page.prototype.layout().
-    const rect = document.body.getBoundingClientRect();
-    const e = document.elementFromPoint((rect.width / 2) | 0, (rect.height / 2) | 0);
-    return e;
+export function forceLayout(body, layoutMode = "elementFromPoint") {
+    if (body !== undefined)
```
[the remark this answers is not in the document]
julienw: oh maybe this is the reason of the different perf characteristics ^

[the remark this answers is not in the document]
camillobruni: ups, this was a wrong refactoring.

camillobruni:
Just double-checked and I can't really see a perf diff anymore.

resources/shared/params.mjs
```diff
@@ -1,3 +1,5 @@
+export const LAYOUT_MODES = Object.freeze(["getBoundingClientRect", "elementFromPoint"]);
```
rniwa, line 1: It's a bit confusing that "elementFromPoint" does both getBoundingClientRect and elementFromPoint.
Maybe that should be reflected in the enum name? e.g. onlyBoundingRect, boundingRectAndElementFromPoint

rniwa (approved):

resources/shared/helpers.mjs
```diff
@@ -23,8 +23,10 @@ export function getAllElements(selector, path = [], lookupStartNode = document)
...
-export function forceLayout() {
-    const rect = document.body.getBoundingClientRect();
-    const e = document.elementFromPoint((rect.width / 2) | 0, (rect.height / 2) | 0);
-    return e;
+export function forceLayout(body, layoutMode = "getBoundingRectAndElementFromPoint") {
+    body ??= document.body;
+    const rect = body.getBoundingClientRect();
+    if (layoutMode === "getBoundingRectAndElementFromPoint")
+        return document.elementFromPoint((rect.width / 2) | 0, (rect.height / 2) | 0);
+    return rect.height;
```
rniwa, line 31: Should we assert that the value is `getBoundingClientRect` here?

rniwa (approved):

resources/shared/helpers.mjs
camillobruni: DONE

julienw (approved):
Yeah I don't see a perf difference anymore

thanks!

PR #511, opened 2025-04-30
Use Params.prototype.toSearchParams() in developer-mode.mjs

camillobruni:
Remove duplicated code and directly use `Params.prototype.toSearchParams()`.
Now the Params class is the centralized place to implement the serialization logic.


julienw (approved):
looks good to me, thanks

I noticed one small bug (that was present before): after clicking "unselect all", I expect to see the parameter `suites=`, but instead the parameter disappears completely. So when reloading the page we're back to the default set of suites.

camillobruni:
- Using Params.tags with `[ "default"]` by default now
- Ignore single default tag
- Support no suite selection (will still display a warning)

PR #512, opened 2025-05-21
Improve deverloper menu and non-standard params table styling

camillobruni:
I very often click on the step instad of the start button and don't immediately see what's the non-standard param value.

- Use tags-style for step-button
- Increase start / step button icon
- Use red highlight color for non-standard params

Screenshot 2025-05-21 at 14 56 45
Screenshot 2025-05-21 at 14 56 35

### DIFF
--- a/resources/main.css
+++ b/resources/main.css
@@ -287,6 +287,7 @@ button,
 }
 
 .developer-mode-content button {
+    font-size: 16px;
     background: white;
     flex: auto;
     padding: 4px;
@@ -294,8 +295,20 @@ button,
     border: 2px solid rgba(255, 255, 255, 0.5);
     border-radius: 10px;
 }
+
+.developer-mode-content button span {
+    padding-left: 10px;
+    font-size: 25px;
+    line-height: 10px;
+    position: relative;
+    top: 2px;
+}
+
 .developer-mode-content .tag {
     border-radius: 100vh;
+}
+
+.developer-mode-content .step-button {
     color: white;
     background: rgba(255, 255, 255, 0.1);
 }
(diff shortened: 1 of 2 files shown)


resources/main.css
```diff
@@ -287,15 +287,28 @@ button,
...
     flex: auto;
     padding: 4px;
     appearance: button;
     border: 2px solid rgba(255, 255, 255, 0.5);
     border-radius: 10px;
 }
+
+.developer-mode-content button span {
+    padding-left: 10px;
+    font-size: 25px;
```
julienw, line 301: i'd use a size in em (for example `1.5em` looks fine)

resources/main.css
```diff
@@ -287,15 +287,28 @@ button,
...
+    line-height: 10px;
+    position: relative;
+    top: 2px;
```
julienw, line 304: I understand you're trying to center them but that doesn't work well:
image

The main issue is that the button isn't high enough I believe?

I think you want something like this on the button:
```css
display: inline-flex;
align-items: center;
justify-content: center;
```

But I couldn't make it _very_ centered. I always found it difficult to center font icons rather than real icons TBH...

So I'm fine if it's not perfect either.

resources/main.css
```diff
@@ -547,6 +560,9 @@ section#details .non-standard-params h2 {
 #non-standard-params-table tbody td {
     padding: 0.1em 0.3em;
 }
+#non-standard-params-table tbody td:nth-child(2) {
+    color: var(--highlight);
```
julienw, line 564: This red isn't very contrasted:
image

Wanna make it lighter?

julienw (approved):
Just a few fixes but it should be OK then

rniwa (approved):

resources/main.css
camillobruni: Here on my mac this looks reasonably centered on all browsers:

Safari: Screenshot 2025-05-27 at 16 21 12
Chrome:  Screenshot 2025-05-27 at 16 21 29
Firefox: Screenshot 2025-05-27 at 16 21 41


I guess the rest will be down to custom fonts with different base positions.

resources/main.css
camillobruni: Done.

PR #514, opened 2025-05-31
Update Lit TodoMVC to latest Lit and TypeScript

justinfagnani:
Update's Lit to a production version and TypeScript to the latest which supports standard decorators. Switches decorators from experimental decorators to standard.


camillobruni:
Let's maybe move this to the `/experimental` folder since it's unclear which lit version we ultimately want to test.
Would that be ok with you?

PR #515, opened 2025-06-02
Enable Speedometer to use an external config.json file

flashdesignory:
This pr sets up Speedometer to use an optional `config` url param, to assign different workloads / tests to use with the benchmark. 

The initial config.json enables the news site next workload, which is currently the only workload that opted into the postMessage api for it's experimental version. The test name in the config is optional, if the default tests are targeted.

The new `data-provider' checks an allowed list of external urls that the workloads can be pulled in from. Currently this just houses the netlify preview folder. 

Other changes:
the `suites` and `tags` are not global anymore and the `data-provider` is the source of truth. This is to ensure that the `data-provider` can process the suites / tests, prior to running speedometer in any mode. 

news-site dist folder changed, since I had to run the `build` again, after updating shared files from the benchmark.


resources/developer-mode.mjs
```diff
@@ -1,7 +1,10 @@
-import { Suites, Tags } from "./tests.mjs";
 import { params } from "./shared/params.mjs";
 
-export function createDeveloperModeContainer() {
+let suites, tags;
+
+export function createDeveloperModeContainer(_suites, _tags) {
```
julienw, line 5: nit: usually prefixing with a `_` means that it's unused.

You could call them `aSuites` and `aTags`...

resources/main.mjs
```diff
@@ -1,10 +1,11 @@
 import { BenchmarkRunner } from "./benchmark-runner.mjs";
 import * as Statistics from "./statistics.mjs";
-import { Suites } from "./tests.mjs";
 import { renderMetricView } from "./metric-ui.mjs";
 import { defaultParams, params } from "./shared/params.mjs";
 import { createDeveloperModeContainer } from "./developer-mode.mjs";
 
+const { dataProvider } = await import("../resources/data-provider.mjs");
```
julienw, line 7: I feel like you use `await` here because you have the top level await in data-provider, but that's not necessary, I'd say to just use a static import.

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await

Although we might want to avoid the top level await and instead have a promise we can wait for in the constructor below, or something like that.

resources/main.mjs
```diff
@@ -19,9 +20,11 @@ class MainBenchmarkClient {
     _metrics = Object.create(null);
     _steppingPromise = null;
     _steppingResolver = null;
+    _dataProvider = null;
 
-    constructor() {
-        window.addEventListener("DOMContentLoaded", () => this.prepareUI());
```
julienw, line 24: we're not waiting for DOMContentLoaded anymore?

Or is it because of the `await` above we're _usually_ past this event... but this is racy so we should probably keep it but _also_ check for `document.readyState`...

julienw (commented):
I didn't test fully locally and I think I don't get the full picture yet, but here is a first set of comments.

resources/data-provider.mjs
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            defaultSuites.flatMap((suite) => suite.tags).forEach((tag) => this._tags.add(tag));
+            defaultSuites.forEach((suite) => this._suites.push(suite));
+        }
+
+        this._freezeTags();
+        this._freezeSuites();
+    }
+
+    enableSuites(names, tags) {
```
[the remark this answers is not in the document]
julienw: okay; this can be made simpler to read for sure...

julienw (commented):
Thanks, this looks cleaner

My main issue now is that the button doesn't react until the config file is loaded, this isn't a pleasant experience.

I made a suggestion below to solve that without too many changes.
Tell me what you think

resources/data-provider.mjs
```diff
@@ -0,0 +1,166 @@
...
+                validTags: Array.from(this._tags),
+            };
+        }
+        alert(message);
+        console.error(message, debugInfo);
+    }
+}
+
+const dataProvider = new DataProvider();
+await dataProvider.init();
```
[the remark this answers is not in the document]
flashdesignory: that makes sense - thanks for the suggestion. 
I'll make the changes 👍

resources/main.mjs
```diff
@@ -339,10 +341,10 @@ class MainBenchmarkClient {
         });
 
         if (params.suites.length > 0 || params.tags.length > 0)
-            Suites.enable(params.suites, params.tags);
+            this._dataProvider.enableSuites(params.suites, params.tags);
 
         if (params.developerMode) {
-            this._developerModeContainer = createDeveloperModeContainer(Suites);
+            this._developerModeContainer = createDeveloperModeContainer(this._dataProvider.suites, this._dataProvider.tags);
```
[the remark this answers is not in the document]
flashdesignory: We should be able to do that. 
I'll make that change 😄

rniwa:
Is this something we want to merge given we've created the experimental folder??

flashdesignory:
> Is this something we want to merge given we've created the experimental folder??

Yes, since both achieve different things.

rniwa:
> > Is this something we want to merge given we've created the experimental folder??
> 
> Yes, since both achieve different things.

But why do we want this change? We definitely don't want to add the ability to load arbitrary cross-origin content in `browserbench.org`.

flashdesignory:
The reasons in the document I previously shared (Speedometer 4 External Test Config Proposal) are still valid and one advantage of this pr is that it'll allow us to test workloads that are not static. 

We have currently workloads in Speedometer, that use frameworks where we force a static output. That is not how they are used on the web.  We have to find a way to allow non-static builds, and this solution would provide a way of doing that. 

We are safe-guarding with the allow list where these workloads are coming from.

rniwa:
> We are safe-guarding with the allow list where these workloads are coming from.

I don't think that mitigates the concern we have regarding content being loaded from another domain. The whole point of using a folder instead of a separate repository was to avoid this exact problem so I don't see how we'd go back & also land this change.

resources/main.mjs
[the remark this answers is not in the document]
flashdesignory: done

resources/data-provider.mjs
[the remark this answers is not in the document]
flashdesignory: @julienw - implemented. Hope you still have time to review 😄

camillobruni:
External configurations are and always have been a separate topic from the experimental workloads.
We still want to test cross-origin performance in the long run and decouple the runner from the workloads to gain more freedom (this is an addition to experimental workloads).

We can lock this down for the official hosted browserbench.org page if that mitigates the concerns and only allow this on the development branch IIF the experimental folder is present.

@rniwa, @mdubet  would this address your concerns?

resources/data-provider.mjs
```diff
@@ -0,0 +1,165 @@
+// example url for local testing:
+// http://localhost:8080/?config=http://localhost:8080/resources/config.json
+import { defaultSuites } from "./tests.mjs";
+import { params } from "./shared/params.mjs";
+
+const DEFAULT_TAGS = ["all", "default", "experimental"];
+const ALLOWED_DOMAINS = {
+    "app.netlify.com": "/sites/webkit-speedometer-preview/",
```
bgrins: At the meeting today, we discussed changing this to restrict the hosts that are allowed to load config files at all (to netlify & localhost) **instead of** restricting the domains that tests are allowed to load from. This way it would be clear that browserbench.org never loads any external config/tests.

resources/data-provider.mjs
```diff
@@ -0,0 +1,149 @@
...
+    _freezeTags() {
+        Object.freeze(this._tags);
+    }
+
+    async init() {
+        if (params.config) {
+            try {
+                const configUrl = new URL(params.config);
+                // Don't fetch if the URL is from DISALLOWED_DOMAINS
+                if (DISALLOWED_DOMAINS.includes(configUrl.hostname)) {
```
rniwa: This is not the check we want.

We want to only allow fetching of a cross-origin configuration file if Speedometer is hosted on localhost or https://webkit-speedometer-preview.netlify.app.

resources/data-provider.mjs
```diff
@@ -0,0 +1,149 @@
...
+                    console.warn(`Configuration fetch from a disallowed domain: ${params.config}. Loading default suites.`);
+                    this._loadDefaultSuites();
+                    return;
+                }
+
+                const response = await fetch(params.config);
+                // ✅ Validate that the network request was successful
```
rniwa: Can we not use emojis?

resources/data-provider.mjs
```diff
@@ -0,0 +1,149 @@
+// example url for local testing:
+// http://localhost:8080/?config=http://localhost:8080/resources/config.json
+import { defaultSuites } from "./tests.mjs";
+import { params } from "./shared/params.mjs";
+
+const DEFAULT_TAGS = ["all", "default", "experimental"];
+const DISALLOWED_DOMAINS = ["browserbench.org"];
+export class DataProvider {
```
rniwa: `DataProvider` sounds very generic. How about `ConfigurationFetcher`?

resources/data-provider.mjs
```diff
@@ -0,0 +1,149 @@
...
+    _tags = new Set(DEFAULT_TAGS);
```
rniwa: Can we just use private variables?

resources/data-provider.mjs
```diff
@@ -0,0 +1,149 @@
...
+
+                const config = await response.json();
+                // ✅ Validate the structure of the fetched config object
+                if (!config || !Array.isArray(config.suites))
+                    throw new Error("Could not find a valid config structure!");
+
+                config.suites.flatMap((suite) => suite.tags || []).forEach((tag) => this._tags.add(tag));
+                config.suites.forEach((suite) => {
+                    // ✅ Validate each suite object before processing
+                    if (suite && suite.url && this._isValidUrl(suite.url))
```
rniwa: Why doesn't this result in an exception instead?
Silently skipping a suite when its URL is invalid sounds rather misleading.

resources/shared/params.mjs
```diff
@@ -193,6 +205,14 @@ export class Params {
     }
 }
 
+function isValidJsonUrl(url) {
+    try {
+        return new URL(url) && url.toLowerCase().endsWith(".json");
```
rniwa: This will return true for `foo.html?query=a.json` or `foo.html#bar.json`.
I don't think that's what we want? Can we just use `URL`'s `pathname` instead?

resources/data-provider.mjs
flashdesignory: It does a little more than just fetching a config, how about `BenchmarkConfigurator` or `SuitesManager` ?

resources/data-provider.mjs
flashdesignory: throwing an error now

resources/shared/params.mjs
flashdesignory: Added a better validation. Let me know if that works for you.

resources/data-provider.mjs
camillobruni: I would prefer a block list instead of an allow list, since chrome does have separate domains where we host workloads and I could see this change in the future. I'd prefer not to have to modify the benchmark sources for that.

What's the concern otherwise here?

flashdesignory: one change I made yesterday, is to check the hostname of `window.location`, instead of the `params.config`, since the main concern was to check where Speedometer is hosted and not the config file. 
@rniwa please let me know if this works for you.

resources/data-provider.mjs
rniwa: Sure, either name is better than `DataProvider`.

resources/data-provider.mjs
```diff
@@ -0,0 +1,151 @@
...
+    _freezeTags() {
+        Object.freeze(this._tags);
+    }
+
+    async init() {
+        if (params.config) {
+            try {
+                const benchmarkUrl = new URL(window.location);
+                // Don't fetch if the URL is from DISALLOWED_DOMAINS
+                if (DISALLOWED_DOMAINS.includes(benchmarkUrl.hostname)) {
```
rniwa: Please disallow all subdomains of `browserbench.org` at minimum.
Something like `benchmarkUrl.hostname.endsWith('browserbench.org')`.

resources/data-provider.mjs
```diff
@@ -0,0 +1,151 @@
...
+
+    _freezeTags() {
+        Object.freeze(this._tags);
+    }
+
+    async init() {
+        if (params.config) {
+            try {
+                const benchmarkUrl = new URL(window.location);
+                // Don't fetch if the URL is from DISALLOWED_DOMAINS
```
rniwa: This comment just repeats the code. I don't think we need it.

resources/data-provider.mjs
flashdesignory: done

resources/data-provider.mjs
flashdesignory: renamed to BenchmarkConfigurator

resources/data-provider.mjs
bgrins: Yeah, checking the window.location of the host is right (as discussed in the meeting). That way a development build can load a config from anywhere, and we are certain not to load any remote configs on browserbench.

If we decided to move to doing fully-remote-config tests even for official tests we could relax it to allow is so long as the config is also on browserbench, but we could cross that bridge in the future if needed - I like keeping it simple here as this does.

bgrins (approved):

flashdesignory:
@rniwa - following up, if we're good with the pr.

resources/benchmark-configurator.mjs
```diff
@@ -0,0 +1,149 @@
+// example url for local testing:
+// http://localhost:8080/?config=http://localhost:8080/resources/config.json
+import { defaultSuites } from "./tests.mjs";
+import { params } from "./shared/params.mjs";
+
+const DEFAULT_TAGS = ["all", "default", "experimental"];
+const DISALLOWED_DOMAINS = ["browserbench.org"];
+export class BenchmarkConfigurator {
+    _tags = new Set(DEFAULT_TAGS);
```
rniwa: Why not use private variables?

resources/tests.mjs
```diff
@@ -2,1118 +2,1022 @@ import { BenchmarkTestStep } from "./benchmark-runner.mjs";
...
-};
-
-Suites.push({
-    name: "TodoMVC-LocalStorage",
-    url: "experimental/todomvc-localstorage/dist/index.html",
-    tags: ["todomvc"],
-    async prepare(page) {
-        (await page.waitForElement(".new-todo")).focus();
-        page.getLocalStorage().getItem("javascript-es5");
+export const defaultSuites = [
```
rniwa: Can we undo this change? It's difficult to review all these line changing the format.

resources/benchmark-configurator.mjs
```diff
@@ -0,0 +1,149 @@
+// example url for local testing:
+// http://localhost:8080/?config=http://localhost:8080/resources/config.json
+import { defaultSuites } from "./tests.mjs";
```
rniwa: Maybe we wanna rename `tests.mjs` to `default-tests.mjs`?

resources/shared/params.mjs
```diff
@@ -193,6 +205,18 @@ export class Params {
     }
 }
 
+function isValidJsonUrl(url) {
+    if (typeof url !== "string" || url.length === 0)
+        return false;
+
+    try {
+        const urlObject = new URL(url, "http://www.example.com");
+        return urlObject.pathname.toLowerCase().endsWith(".json");
```
rniwa: This isn't necessarily required, right? You can have other file extensions or no file extension at all with `Content-Type: application/json`.

flashdesignory: correct, is this a blocker?

resources/tests.mjs
flashdesignory: I'd love to not revert. I'm aware that this file looks like there are many changed lines, but basically it's a simple cleanup. Happy to walk through the changes in this file, if a sync is helpful.

flashdesignory: the gist is:
default-tests file now only contains the defaultSuites array, but it doesn't extend the array anymore (Suites.enable) and it doesn't automatically freeze the Suites / Tags.

The BenchmarkConfigurator is now responsible for assigning suites and calling enableSuites / freezeSuites / freezeTags.

camillobruni: Maybe if wouldn't rename the file the diff is easier.

resources/config.json
```diff
@@ -0,0 +1,13 @@
+{
+    "suites": [
+        {
+            "name": "NewsSite-PostMessage",
+            "url": "resources/newssite/news-next/dist/index.html",
+            "tags": ["default", "newssite", "language"],
```
rniwa: Why is this tag including "default"? Isn't it supposed to be "experimental" instead?

resources/benchmark-configurator.mjs
```diff
@@ -0,0 +1,149 @@
...
+            });
+        } else {
+            console.warn("Neither names nor tags provided. Enabling all default suites.");
+            this.#suites.forEach((suite) => {
+                suite.enabled = suite.tags.includes("default");
+            });
+        }
+        if (this.#suites.some((suite) => suite.enabled))
+            return;
+        let message, debugInfo;
```
rniwa: Nit: please don't declare multiple variables in a single statement.

resources/benchmark-configurator.mjs
```diff
@@ -0,0 +1,149 @@
...
+                validNames: this.#suites.map((each) => each.name),
+            };
+        } else if (tags?.length) {
+            message = `Tags "${tags}" does not match any Suite. No tests to run.`;
+            debugInfo = {
+                providedTags: tags,
+                validTags: Array.from(this.#tags),
+            };
+        }
+        alert(message);
```
rniwa: I think we should extra a helper function like _reportError which takes message & debugInfo instead of declaring local variables then if-else branching it like this. So that we can do instead:
```js
if (names?.length)
    this._reportError(`Suites "${names}" does not match any Suite. No tests to run.`, { providedNames: names, validNames: this.#suites.map((each) => each.name) });
else
    this._reportError(`Tags "${tags}" does not match any Suite. No tests to run.`, { providedTags: tags, validTags: Array.from(this.#tags) });
```
It could be a local lambda too.

resources/shared/params.mjs
rniwa: I'd rather not have arbitrary rule enforced on a filename.

flashdesignory: removed the extension check

camillobruni (approved):

PR #517, opened 2025-07-01
Bump min nodejs version to v22

camillobruni:
We already had a min version in the top-level package.json file, but it was not the same for all workloads.
This can easily causes issues when having to regenerate individual workloads where the nodejs version is accidentally not matching the previous version used for created the `dist` output.


rniwa (approved):

camillobruni:
let's bumpt to v24 directly in #562

PR #530, opened 2026-01-06
Update TodoMVC-Jaspr-* benchmarks with newer Jaspr & Dart SDK

mkustermann:
This adopts some API changes in Jaspr framework (e.g. build methods no longer returning iterables but a single component, ...) and uses newer Dart2Wasm compiler


camillobruni (approved):
LGTM – seems to improve a bit in score for the dart2js version but other than that no big score change.

mkustermann:
Thanks, @camillobruni ! I don't have permission to merge it, could you do that?

PR #543, opened 2026-06-11
Update developer menu layout

camillobruni:
- Use gridlayout to avoid fixd settings panel height and better adjust to window height
- Indent settings without chekcbox
- Better style select elements
- Use consistent spacing around checkbox in settings and suite selection


camillobruni:
# OLD
Screenshot 2026-06-11 at 13 14 21

# NEW
Screenshot 2026-06-11 at 13 13 27

flashdesignory (approved):

PR #545, opened 2026-06-11
Add --retry support for end2end tests

camillobruni:
Run the the end2end tests with an additionaly retry run to mitigatate flaky browsers with selenium.
This can be further configured with a new --retry flag.


flashdesignory (approved):
would be nice to see a screencast of the retries working

camillobruni:
you mean generally as test artifact after a failure? 
Or rather as demo run where we get a failure and then succeed with the retry?

camillobruni:
https://drive.google.com/file/d/1L9tWJo7UQoZnS4Bnn3P0Q9ZwPCNpJvG9/view?usp=sharing

PR #547, opened 2026-06-15
Add more delay params

camillobruni:
For locall investigation we found that having these additional delay params was very helpful:
- waitAfterSetup: Post setup delay (before running the steps)
- waitAfterSuite: Post suite delay (after running all steps)


Andrew-Fryer:
LGTM

Andrew-Fryer (approved):

PR #549, opened 2026-06-15
Move suites to dedicated folder

camillobruni:
Instead of storing suites partially in resources/ and experimental, let's use a more consistent
naming scheme that makes it more clear where which code lives

- "resources/" runner and shared helpers
- "suites/" contains the default suites
- "suites-experimental/" contains the experimental suites

Based on the rename from PR #548.


flashdesignory (approved):

Andrew-Fryer (commented):
Renaming/moving is definitely good 💯 .

I hit an error that seems relevant on the latest preview.

Image


LGTM if error is fixed and merge conflicts resolved 👍

camillobruni:
investigated this a bit more, and looks like this charAt error might have existed before caused by an "empty" url param `&` appended in `_loadFrame`. I'm going to fix this first in a separate PR #552.

camillobruni:
Landing #552 fixed the charAt issue we were seeing.

Andrew-Fryer:
> investigated this a bit more, and looks like this charAt error might have existed before caused by an "empty" url param `&` appended in `_loadFrame`. I'm going to fix this first in a separate PR #552.

Ah, that makes sense that it was stuck on `React-Stockcharts-SVG` (the suite before `Perf-Dashboard`).

Andrew-Fryer (approved):
LGTM

PR #556, opened 2026-07-07
Update .gitignore file

camillobruni:
- globally exclude node_modules folder
- ignore typical agent folders
- exclude build-dev from all suites by default


Andrew-Fryer (approved):
LGTM

PR #561, opened 2026-07-13
Clean up sum

camillobruni:
Use explicit 0 accumulator value for BenchmarkRunner totals sum

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -468,7 +468,7 @@ export class BenchmarkRunner {
             }
 
             values.sort((a, b) => a - b); // Avoid the loss of significance for the sum.
-            const total = values.reduce((a, b) => a + b);
+            const total = values.reduce((a, b) => a + b, 0);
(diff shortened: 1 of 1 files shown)


resources/benchmark-runner.mjs
```diff
@@ -468,7 +468,7 @@ export class BenchmarkRunner {
             }
 
             values.sort((a, b) => a - b); // Avoid the loss of significance for the sum.
-            const total = values.reduce((a, b) => a + b);
+            const total = values.reduce((a, b) => a + b, 0);
```
Andrew-Fryer, line 471: Out of curiosity, I'm wondering if a linter was complaining or if there's a case where `values` is empty...

Andrew-Fryer (approved):
LGTM

resources/benchmark-runner.mjs
camillobruni: humm you're right.. double cehckd this and the lint error has not appeared again... not sure what I did.

PR #562, opened 2026-07-13
Non-hashed builds and prepare bump node version to 24

camillobruni:
- Disable content hashing across Next.js, Nuxt, Vite, Vue CLI, and Angular CLI  configurations to ensure deterministic build outputs with static filenames (e.g.  main.js  instead of  main.d511.js ).
 - React enforces content-hashing thus we maintain status quo there
 - Bump minimum Node.js version to  >=24.0.0  across all  package.json  files so we can build again on the github CI which requires >= 22
 - Update package-lock.json by running "npm run install" on all workloads


camillobruni:
This is a preparatory PR so we can run the build steps on the CQ in the future. github is depreaction node 20 and we better just bump the min version to npm LTS (e.g. 24) so we can have a proper verifying CQ that the workloads still build.

Avoiding hash / content-hashed file names makes local development easier.

The actual workload update (e.g. dist/ build/) will happen in a follow-up PR (so from a scoring perspective this PR is a noop)

flashdesignory (approved):

Andrew-Fryer:
What's CQ?

camillobruni:
sorry, probably chrome-lingo :) CQ => Commit Queue

PR #564, opened 2026-07-14
Bind dev server to IPv4 address

camillobruni:
Using "localhost' can cause some issues on certain OS / browser combination where we end up binding to the IPv6 address instead of IPv4 and the browser uses the other version for resolving "localhost". Forcing IPv4 for the test server seems to resolve this.


Andrew-Fryer (approved):
I haven't reproduced, but sounds good to me.

PR #569, opened 2026-07-16
Add workload link to developer menu

camillobruni:
This makes it easy to navigate to the raw workload for debugging and investigation.


camillobruni:
Screenshot 2026-07-16 at 10 23 55

Bas-moz (approved):

PR #573, opened 2026-07-21
Improve MainBenchmarkClient with state and better error handling

camillobruni:
- Add explicit BENCHMARK_STATE as preparation for ServiceWorker-based preloading
- Use new state to early-abort runs if there are toplevel errors
- Add explicit top-level error and unhandledrejection handlers to surfce errors better


Bas-moz (approved):
Looks like it's strictly an improvement in terms of information and I can't see how it would affect the test score.

PR #453, opened 2024-11-15
Experimental Workload: Responsive Design

issackjohn:
This pull request introduces a new experimental responsive design workload, featuring a single-page cooking website built with Lit and Tailwind CSS.

The hosted version can be found [here](https://neon-gumdrop-d49b03.netlify.app/).
[Quick link to dev mode with workload selected.
](https://neon-gumdrop-d49b03.netlify.app/?developerMode=&suites=Responsive-Design)
Please look at the [README](https://github.com/WebKit/Speedometer/pull/453/files#diff-83d565323d6c3520a4207b0f9b17122f2c51a3a16238c868e7e53effacfffc9f) in `experimental/responsive-design` for more information about the code and how it is organized.

Not included in this PR:
- Support for different languages

PR #579, opened 2026-07-24
Track Responsive-Design carousel resize events

issackjohn:
Follow-up to the review feedback on #453, which asked to track the number of resize events so we can verify consistency across browsers.

- Add PageElement.observeResizeEvents() to count distinct content-box width changes delivered by ResizeObserver. The first delivery seeds the baseline and is not counted, and repeated deliveries of the same width are ignored.
- Observe the recipe carousel during ReduceWidthIn5Steps and IncreaseWidthIn5Steps, and log the observed count so the numbers can be compared across browsers.
- Add unittests covering baseline seeding, distinct width changes, and duplicate-width deliveries.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -163,6 +163,46 @@ class PageElement {
         this.#node.scrollIntoView(options);
     }
 
+    observeResizeEvents() {
+        const contentWindow = this.#node.ownerDocument.defaultView;
+        const state = {
+            count: 0,
+            lastWidth: null,
+        };
+        let markReady;
+        // Resolves on the first delivery so callers can seed a baseline before resizing.
+        const ready = new Promise((resolve) => {
+            markReady = resolve;
+        });
+        const observer = new contentWindow.ResizeObserver((entries) => {
+            for (const entry of entries) {
+                const contentBoxSize = entry.contentBoxSize;
+                const inlineSize = Array.isArray(contentBoxSize) ? contentBoxSize[0]?.inlineSize : contentBoxSize?.inlineSize;
+                const width = inlineSize ?? entry.contentRect.width;
+                // The first delivery seeds the baseline width without counting it as a change.
+                if (state.lastWidth === null) {
+                    state.lastWidth = width;
+                    markReady();
+                    continue;
+                }
+                if (width === state.lastWidth)
+                    continue;
+                state.count++;
+                state.lastWidth = width;
+            }
+        });
+        observer.observe(this.#node);
+        return {
+            ready,
+            get count() {
+                return state.count;
+            },
+            disconnect() {
+                observer.disconnect();
+            },
+        };
+    }
+
     dispatchEvent(eventName, options = NATIVE_OPTIONS, eventType = Event) {
         if (eventName === "submit")
             // FIXME FireFox doesn't like `new Event('submit')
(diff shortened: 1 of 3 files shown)


issackjohn:
@camillobruni PTAL :)

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,46 @@ class PageElement {
...
+        const contentWindow = this.#node.ownerDocument.defaultView;
+        const state = {
+            count: 0,
+            lastWidth: null,
+        };
+        let markReady;
+        // Resolves on the first delivery so callers can seed a baseline before resizing.
+        const ready = new Promise((resolve) => {
+            markReady = resolve;
+        });
```
camillobruni, line 176: Is there a specific use-case you had in mind for detaching observer setup and sarting it?

I think we could just either:
- always await this promise here
- just not await it and assume we that we migth be off by 1? (e.g. just don't count it) and we have to make sure that at the time we stop, we have handlede all events.

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,46 @@ class PageElement {
...
+        const observer = new contentWindow.ResizeObserver((entries) => {
+            for (const entry of entries) {
+                const contentBoxSize = entry.contentBoxSize;
+                const inlineSize = Array.isArray(contentBoxSize) ? contentBoxSize[0]?.inlineSize : contentBoxSize?.inlineSize;
+                const width = inlineSize ?? entry.contentRect.width;
+                // The first delivery seeds the baseline width without counting it as a change.
```
camillobruni: nit: I was a bit confused about "delivery" :) I "callback" might be more common?

resources/benchmark-runner.mjs
```diff
@@ -163,6 +163,46 @@ class PageElement {
...
+        });
+        observer.observe(this.#node);
+        return {
+            ready,
+            get count() {
+                return state.count;
+            },
+            disconnect() {
+                observer.disconnect();
+            },
```
camillobruni, line 200: Would it make sense to have a single stop() api here that disconnects and returns the count? 
I don't tihnk we would want to manually interact with the underlaying observer in any other way.

suites-experimental/suites.mjs
```diff
@@ -289,6 +301,12 @@ export const ExperimentalSuites = freezeSuites([
                 }
 
                 await new Promise((resolve) => requestAnimationFrame(() => requestAnimationFrame(resolve)));
+                const { count } = carouselResizeObservations;
+                carouselResizeObservations.disconnect();
+                if (count)
+                    console.warn(`IncreaseWidthIn5Steps: recipe-carousel ResizeObserver delivered ${count} coalesced width change(s).`);
+                else
+                    console.warn("IncreaseWidthIn5Steps: recipe-carousel ResizeObserver delivered 0 width changes; expected width changes during iframe resize.");
```
camillobruni: nit: maybe add a helper to print this message?

resources/benchmark-runner.mjs
issackjohn: Done.

resources/benchmark-runner.mjs
issackjohn: Agreed. Done.

suites-experimental/suites.mjs
issackjohn: Done.

resources/benchmark-runner.mjs
issackjohn: I do not have a use case for separating setup from readiness, so awaiting internally is simpler.

camillobruni (approved):
LGTM thanks!

PR #531, opened 2026-01-30
Add TodoMVC with indexedDB storage

issackjohn:
Results for the hosted prototype as of 12/12/2025

label | macos.arm64.local.0_0 | macos.arm64.local.0_1 | macos.arm64.local.0_2 | macos.arm64.local.0_3
-- | -- | -- | -- | --
browser | Microsoft Edge | Firefox | Google Chrome | Safari
version | 143.0.3650.80 | 146.0.0 | 143.0.7499.110 | 26.2.0.0.21623.1.14.11.9
os | macos 26.2 arm64 | macos 26.2 arm64 | macos 26.2 arm64 | macos 26.2 arm64
model | MacBookPro17,1 | MacBookPro17,1 | MacBookPro17,1 | MacBookPro17,1
cpu | Apple M1 8 cores | Apple M1 8 cores | Apple M1 8 cores | Apple M1 8 cores
runs | 5 | 5 | 5 | 5
failed runs | 0 | 0 | 0 | 0
TodoMVC-WebComponents | 2.71 ± 4.7% | 3.760 ± 2.2% | 2.87 ± 6.4% | 4.129 ± 1.9%
TodoMVC-WebComponents-IndexedDB | 32.29 ± 2.7% | 32.50 ± 1.3% | 35.61 ± 1.7% | 40.07 ± 1.2%
TodoMVC-WebComponents-DexieJS | 39.23 ± 2.3% | 48.9 ± 2.5% | 41.2 ± 2.5% | 53.8 ± 2.0%
Score | 68.1 ± 2.2% | 55.53 ± 1.1% | 62.9 ± 2.5% | 48.37 ± 0.41%

PR #580, opened 2026-07-24
Avoid function state in IndexedDB page wait

issackjohn:
Follow-up to the review on #531.

`DeletingAllItems` stored its resolver as a property on the
`iterationFinishedListener` function object and registered a
`previous-page-loaded` listener that was never removed.

This replaces that with a `waitForPreviousPageLoaded()` helper which
registers a one-shot listener per page transition, so the resolver is
local to the navigation it belongs to and the listener cleans itself up.


issackjohn:
@camillobruni PTAL :)

camillobruni (approved):

PR #582, opened 2026-08-10
Add experimental Media Performance Workloads

SyedAbuTalib:
This benchmark suite measures the performance, responsiveness, and processing throughput of modern web media APIs.

---

### InitialPlayback

Tests standard video playback initialization latency via Media Source Extensions (MSE) to simulate typical streaming user journeys on platforms like YouTube or Vimeo. The time measured is from the `play` button clicked to the first video frame painted on the screen. We use `requestVideoFrameCallback` to ensure that the video is painted on the screen.

### Seek

Measures the latency of performing a quick skip-ahead action within an already loaded media stream. This simulates a common user interaction, such as scrubbing through a video timeline or double-tapping right to skip forward on a streaming platform. The time measured is from the `seek` button clicked to the first video frame painted on the screen after seeking. We use `requestVideoFrameCallback` to ensure that the video is painted on the screen after seeking.

---

### VideoChat

Tests real-time video encoding and decoding pipelines using WebCodecs. While the UI simulates a Video Chat call, WebRTC is not tested; the focus is purely on measuring VideoEncoder and VideoDecoder throughput. The time measured is from the `video-benchmark` button clicked until all video frames have been encoded, decoded, and the session is torn down.

### VoiceChat

Tests real-time audio encoding, decoding using WebCodecs and audio routing and effects processing using WebAudio. It measures the browser's efficiency in processing audio streams and applying standard audio nodes. The time measured is from the `voice-benchmark` button clicked until all audio frames have been encoded/decoded and the WebAudio offline rendering completes.


camillobruni:
The media-streaming workload does not seem to work on safari:

```
Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
```
Screenshot 2026-08-18 at 12 04 53


Could we figure out a mode that makes it possible to work out of the box there too?

suites-experimental/suites.mjs
```diff
@@ -292,4 +292,44 @@ export const ExperimentalSuites = freezeSuites([
...
+            await page.waitForElement("#video-benchmark");
+        },
+        tests: [
+            new BenchmarkTestStep("VideoChat", async (page) => {
+                page.querySelector("#video-benchmark").click();
+                await page.waitForElement("#video-benchmark.completed");
+            }),
+            new BenchmarkTestStep("VoiceChat", async (page) => {
+                page.querySelector("#voice-benchmark").click();
+                await page.waitForElement("#voice-benchmark.completed");
```
camillobruni: waitForElement does not work in the measured phase of the benchmark :/ since it's doing the work with requestAnimationFrame which means we get at most framerate timing resolution (and framerate dependendent measurement which we try to avoid).

- You have to to do something like new `await page.callAsyncBlocking(...)` method or so (sorry this is not very fleshed out yet)

suites-experimental/media-performance/streaming.js
```diff
@@ -0,0 +1,182 @@
...
+                return;
+            }
+            const checkFrame = (now, metadata) => {
+                // Ensure the frame painted matches or exceeds our target media position.
+                if (metadata.mediaTime >= minMediaTime)
+                    resolve();
+                else
+                    video.requestVideoFrameCallback(checkFrame);
+            };
+            video.requestVideoFrameCallback(checkFrame);
```
camillobruni, line 57: If I interpret this correctly, `requestVideoFrameCallback` is likey display refresh-rate bound? [1]

```
One thing to bear in mind is that requestVideoFrameCallback() does not offer any strict guarantees that the output from your callback will remain in sync with the video frame rate. It may end up being fired one vertical synchronization (v-sync) later than when the new video frame was presented. (V-sync is a graphics technology that synchronizes the frame rate of a video with the refresh rate of a monitor.)
```

If this is the case we need to find another way to measure this. Speedometer is menat to be refresh-rate agnositc for the measurement part.

[1] https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/requestVideoFrameCallback

suites-experimental/suites.mjs
SyedAbuTalib: Created new `callAsyncBlocking()` method. Let me know if you agree.

PR #583, opened 2026-08-11
Add new experimental ChatRoom-React workload

canova:
Hey everyone! This PR adds a new workload called `ChatRoom-React`. It's similar to chat room applications like Slack/Element/Discord etc. The workload exercises room switching. It's built using React + Vite. It contains 40 rooms of 1500 deterministically generated messages each.

There are 3 times steps:

- **ScrollTimeline** reads back through history a page at a time (rows are recycled and cached heights reused), then covers it in long strides the way dragging the scrollbar does (windows replaced outright, correction runs against estimates).
- **SwitchRooms** clicks through the sidebar 30 times. Each timeline is keyed by room id, so a switch fully unmounts the old one and mounts the new one.
- **LoadOlderMessages** scrolls past the start of the loaded history 10 times, which fetches the next page and inserts it above the viewport. That's the case browsers ship scroll anchoring for, and nothing else in Speedometer exercises it.

Note that I captured some profiles of this workload vs Slack vs Element(Matrix) and it looks like they are similar in terms of the things that they spend time on.


canova:
@camillobruni Hey, wanted to point out something and make sure we're on the same page, since I'm doing something a bit differently in this workload. I added a `yieldTask` helper [here](https://github.com/WebKit/Speedometer/pull/583/changes#diff-77f580c7cdce8f084f8c5a6f0f578d6b24d3b58e7793a8f91ff73c0492dfed00R6-R24) and I'm using it to yield a task between the interactions inside a measured step ([for example here](https://github.com/WebKit/Speedometer/pull/583/changes#diff-77f580c7cdce8f084f8c5a6f0f578d6b24d3b58e7793a8f91ff73c0492dfed00R364)), so React commits the pending update before the next click or scroll is driven.

The other async workload "Responsive-Design" just calls `page.layout()` after each click, but that isn't a substitute here. it's a layout barrier, not a scheduler flush. When I dispatch the scroll, React's handler runs synchronously and calls `setState`, but the render and commit get handed off to React's scheduler as a posted task. So the new rows aren't in the DOM yet, and forcing layout would just re-layout the tree that was already there. Without the yield all 20 channel switches batch into a single render at the end, and the step measures 1-2ms instead.

It's a `MessageChannel` round-trip because a posted task is the cheapest way to get past the end of the current task, which a microtask can't do at all. `setTimeout(0)` and rAF work too, but these steps are async, so the timer's clamp or a full frame lands inside the measured window and gets reported as ChatRoom's time. 30 channel switches I measured 397ms with `MessageChannel` vs 427ms with `setTimeout(0)`.

Let me know what you think! I also lowered some of the iteration counts in the steps, so they're a bit more on par with the other workloads.

camillobruni:
from afar this looks ok. I must read up a bit more on the interactions with the message channel. But generally speaking, I'm in favour of having a way to force react's task queue to drain properly without having to path the framework code.

PR #585, opened 2026-08-19
Fix the warmup suite, which has not run since the SuiteRunner refactor

### LINKED ISSUE #584: useWarmupSuite always aborts the run: "invalid 0-time total for suite Warmup"
### Summary

Enabling the warmup suite (`?useWarmupSuite`, or "Use Warmup Suite" in the developer menu) causes every run to abort before any real suite is measured, so the option is currently unusable.

### Cause

`SuiteRunner._recordTestResults()` intentionally returns early for the warmup suite, so `#suiteResults.total` is never incremented and stays `0`:

https://github.com/WebKit/Speedometer/blob/86710fec5ec537503bb252e988c6e76dc8c908df/resources/suite-runner.mjs#L127-L130

However `_validateSuiteResults()` is still called for the warmup suite, and it throws on a `0` total:

https://github.com/WebKit/Speedometer/blob/86710fec5ec537503bb252e988c6e76dc8c908df/resources/suite-runner.mjs#L99
https://github.com/WebKit/Speedometer/blob/86710fec5ec537503bb252e988c6e76dc8c908df/resources/suite-runner.mjs#L107-L110

These two behaviours contradict each other: the warmup suite is required to record nothing, but is then validated as though it must have recorded something. The result is:

```
Got invalid 0-time total for suite Warmup: 0
```

Note there are two unguarded call sites, so a guard at a single call site would leave the remote path broken:

- `SuiteRunner._runSuite()` — line 99
- `RemoteSuiteRunner` — line 209

### Steps to reproduce

Reproduced on the released Speedometer 3.1. The `main` branch was verified by code inspection (see Cause above) rather than by running it.

1. Serve the Speedometer directory over HTTP, e.g. `python3 -m http.server 8901`
2. Open `http://127.0.0.1:8901/?developerMode&useWarmupSuite&suites=TodoMVC-JavaScript-ES5&iterationCount=2`
3. Click **Start Test**.

The run aborts before any real suite is measured. On 3.1 the summary shows "Error — One or more subtests produced no duration"; on `main` the same contradiction throws `Got invalid 0-time total for suite Warmup: 0`.

### Regression

This appears to have been introduced by #406 (merged 2024-07-25), which added the zero-total validation while fixing #399. The warmup suite added in #124 deliberately records no results, so the two have never been compatible.

### Additional context

In the released Speedometer 3.1 (before the `SuiteRunner` refactor) the same contradiction exists in `resources/benchmark-runner.mjs`, where `_validateSuiteTotal()` dereferences `this._measuredValues.tests["Warmup"]`. Since that entry is never created, it fails as a `TypeError`, and the UI reports the misleading message "One or more subtests produced no duration" rather than naming the warmup suite.

Speedometer 3.1 with `useWarmupSuite` enabled:

Image

### Suggested fix

Skip validation for the warmup suite. Because there are two call sites, an early return inside `_validateSuiteResults()` covers both and mirrors the existing guard in `_recordTestResults()`:

```js
_validateSuiteResults() {
    // The warmup suite's results are intentionally not recorded, so it has no total.
    if (this.#suite === WarmupSuite)
        return;
    ...
```

`WarmupSuite` is already imported in `suite-runner.mjs`, so no new import is needed.

I'm happy to open a PR if this approach looks right.

harini9803:
Fixes #584.

Enabling `useWarmupSuite` has had no effect on `main` since the `SuiteRunner` refactor. Reviving it turned up four independent defects, each only reachable once the previous one is fixed — which is likely why they went unnoticed.

**1. The warmup suite is never run.** `runAllSuites()` skips any suite without an `enabled` flag, and `WarmupSuite` does not define one — that flag is only set by `BenchmarkConfigurator` on the regular suites. So `useWarmupSuite` is currently a silent no-op.

**2. Its `url` is stale.** `_loadFrame()` resolves suite URLs against the document base, but `WarmupSuite.url` was still the 3.1-era `warmup/index.html`, which `_prepareSuite()` used to prefix with `resources/`. The file now lives at `suites/warmup/index.html`, so the frame 404s.

**3. Its results zero the score.** The `SuiteRunner` constructor eagerly registers `measuredValues.tests[suite.name]`, including for the warmup suite. Because `_recordTestResults()` intentionally skips warmup, that entry keeps `total: 0`, and the geomean loop does `product *= suiteTotal` over every entry — so the geomean becomes `0` and the score `Infinity`. In 3.1 the entry was created lazily inside `_recordTestResults()`, so it never existed. This is the most serious of the four: it produces a wrong score rather than failing loudly.

**4. Validation throws on the zero total.** `_validateSuiteResults()` is called for every suite and throws when the total is `0`, which the warmup suite's always is. Guarding inside the method covers both call sites (`SuiteRunner._runSuite` and `RemoteSuiteRunner`).

Plus one UI fix: the warmup suite is not counted in `stepCount`, so `didFinishSuite()` would push the progress bar past its maximum.

### Verification

Run `?developerMode&useWarmupSuite&suites=TodoMVC-JavaScript-ES5&iterationCount=2`:

- before: `useWarmupSuite` is accepted but no warmup runs
- after each fix in turn: 404 on the warmup frame, then `Geomean 0.00` with an `Infinity` score, then a clean run
- with all four: the warmup suite loads and runs, and the geomean equals the single suite's total (87.65 in my run), confirming warmup stays out of the aggregate

Happy to split this into separate commits or drop the progress-bar change if you would prefer a narrower diff.

### DIFF
--- a/resources/benchmark-runner.mjs
+++ b/resources/benchmark-runner.mjs
@@ -239,7 +239,13 @@ function geomeanToScore(geomean) {
 // compilation of runner methods in the middle of the measuring cycle.
 export const WarmupSuite = {
     name: "Warmup",
-    url: "warmup/index.html",
+    // _loadFrame() resolves suite URLs against the document base, so this needs the same
+    // root-relative form as the regular suites. It was still the 3.1-era path, which
+    // _prepareSuite() used to prefix with "resources/".
+    url: "suites/warmup/index.html",
+    // runAllSuites() skips suites that are not enabled. Unlike the regular suites, the warmup
+    // suite is not passed through BenchmarkConfigurator, so nothing else sets this for it.
+    enabled: true,
     async prepare(page) {
         await page.waitForElement("#testItem");
     },
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -1,4 +1,4 @@
-import { BenchmarkRunner } from "./benchmark-runner.mjs";
+import { BenchmarkRunner, WarmupSuite } from "./benchmark-runner.mjs";
 import * as Statistics from "./statistics.mjs";
 import { renderMetricView } from "./metric-ui.mjs";
 import { defaultParams, params } from "./shared/params.mjs";
@@ -149,6 +149,9 @@ class MainBenchmarkClient {
     }
 
     async didFinishSuite(suite) {
+        // The warmup suite is not part of stepCount, so counting it would overshoot the bar.
+        if (suite === WarmupSuite)
+            return;
         this._finishedTestCount++;
         this._progressCompleted.value = this._finishedTestCount;
         if (this._steppingPromise)
--- a/resources/suite-runner.mjs
+++ b/resources/suite-runner.mjs
@@ -101,6 +101,9 @@ export class SuiteRunner {
     }
 
     _validateSuiteResults() {
+        // The warmup suite's results are intentionally not recorded, so it has no total.
+        if (this.#suite === WarmupSuite)
+            return;
         // When the test is fast and the precision is low (for example with Firefox'
         // privacy.resistFingerprinting preference), it's possible that the measured
         // total duration for an entire is 0.
--- a/resources/suite-runner.mjs
+++ b/resources/suite-runner.mjs
@@ -19,10 +19,16 @@ export class SuiteRunner {
 
     constructor(frame, page, params, suite, client, measuredValues) {
         // FIXME: Create SuiteRunner-local measuredValues.
-        this.#suiteResults = measuredValues.tests[suite.name];
-        if (!this.#suiteResults) {
+        if (suite === WarmupSuite) {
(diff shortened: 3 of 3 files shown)


harini9803:
Verified on this PR's deploy preview: https://deploy-preview-585--webkit-speedometer-preview.netlify.app/?developerMode&useWarmupSuite&suites=TodoMVC-JavaScript-ES5&iterationCount=2 -- the warmup suite loads and runs, and the geomean (122.41 ms) equals the single suite's total, confirming warmup stays out of the aggregate. The same URL on the production preview accepts useWarmupSuite but never runs a warmup suite.

camillobruni (approved):
thanks for the fix!

camillobruni:
Shall I land this for you? Please let me know so we can close the PR

harini9803:
> Shall I land this for you? Please let me know so we can close the PR

Yes Sure. Thanks.

PR #586, opened 2026-08-24
Name each remote workload's in-page suite after the workload

canova:
I noticed this while I was looking at some profiles with performance.mark/measure markers. I could see some `suite-default` which looks odd compared to what we usually have in other workloads with proper names (like `suite-Responsive-Design` etc.).

It looks like this was becuse we were passing `"default"` as a name paramater to the BenchmarkSuite constructor. This patch adds proper names for these workfloads so they can get better names instead of `suite-default` in their user timing entries.

See the performance.mark/measure calls here:
https://github.com/WebKit/Speedometer/blob/9304efe45ef1d982c5eb9ac14f6353b264d56940/resources/shared/benchmark.mjs#L56-L78


canova:
I also saw that we had 2 `suite-indexeddb` runs per iteration. It looks like this is because both TodoMVC-WebComponents-IndexedDB and TodoMVC-WebComponents-DexieJS tests were named as `indexeddb` which makes it difficult to see which one is which. I added the second commit to fix that.

camillobruni (approved):

canova:
Thanks for the review @camillobruni! I don't have the permissions to merge it myself, would you mind merging?

PR #587, opened 2026-08-31
Fix benchmark startup sequence for testing

camillobruni:
There could be a race condition between sending the SpeedometerReady and and the testing running installing the event handler. This PR delays sending the SpeedometerReady event to a later async point in the `evaluateParams` method..

### DIFF
--- a/resources/main.mjs
+++ b/resources/main.mjs
@@ -26,16 +26,17 @@ class MainBenchmarkClient {
     _steppingPromise = null;
     _steppingResolver = null;
     _benchmarkConfiguratorPromise = null;
+    _isInitialized = false;
+
+    get isInitialized() {
+        return this._isInitialized;
+    }
 
     constructor() {
         this._benchmarkConfiguratorPromise = import("./benchmark-configurator.mjs");
         this.prepareUI();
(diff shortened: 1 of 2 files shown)


resources/main.mjs
```diff
@@ -26,16 +26,17 @@ class MainBenchmarkClient {
     _steppingPromise = null;
     _steppingResolver = null;
     _benchmarkConfiguratorPromise = null;
+    _isInitialized = false;
+
+    get isInitialized() {
```
Andrew-Fryer, line 31: Nit:
Does this need to be a getter instead of just a method like `isRunning` and `hasFinished`?

Andrew-Fryer (approved):
This looks good to me. 👍 

I like that we won't dispatch "SpeedometerReady" if something in the constructor throws, so there's a bit less risk of silently reporting a wonky benchmark score.

I think it's a bit weird that we dispatch the "SpeedometerReady" from `evaluateParams`, but we're already calling `this.start()` from there.

Perhaps we should refactor in a future PR to move these to a callback passed to `evaluateParams().then(callback)` in the constructor...?

